PR #54, opened 2025-07-30
Add SSO Google Meet Support + Abstract Bot logic

jspada200:
This PR adds support for Google Meet when hosted with a Google Workspace. It handles the sign-in flow through okta including 2FA when required as well as the slightly different join flow from a public google meet. 

Since much of the logic between the google and google-okta is shared, a new abstract class was created and the google bot updated to use it. This allows the specific logic for joining to be easily overridden. 

This MR also includes small quality of life improvements like adding a docker-compose local and the ability to keep bot containers around to inspect the logs. 

- Adds Okta SSO support with google workspaces. 
- Adds abstract class for handling platform specific logic.
- Updates run_vexa_interaction.sh to allow for selection of a specific platform
- Adds docker-compose-local.yml to mount bot-manager dir and provide env var to keep bot containers around.


jspada200:
Example of joining a google workspace meet

PR #195, opened 2026-03-25
fix: bot leaves active meeting when video track arrives

### LINKED ISSUE #189: [Bug] Bot leaves meeting when participants join Google Meet
## What happened?

The bot leaves the google meet immediately when a participant joins. Pre-built images from the Docker registry do not have this issue. This also happens when anybody refereshes the meet page.

## What did you expect?

Bot is expected to be present in the Google Meet.

## How to reproduce?

1. Build the Lite image locally: docker build -f docker/lite/Dockerfile.lite .
2. Start the bot and have it join a Google Meet.
3. Have another participant join the meeting
4. Bot leaves shortly after the participant joins

## Version / env?

Vexa Lite Locally Built

## Root Cause

The getVideoBlockInitScript() in services/vexa-bot/core/src/services/screen-content.ts (introduced in commits 1b9709d, a71b2af) patches RTCPeerConnection to block incoming video for transcription-only bots. 
When a participant joins and sends a video track, the script calls:

1. event.track.stop() — permanently destroys the MediaStreamTrack
2. event.transceiver.direction = 'inactive' — triggers WebRTC SDP renegotiation

These destabilize the WebRTC connection, causing Google Meet to detect a broken state (e.g. "You left the meeting" / "Connection lost"). The bot's removal monitor then picks up these error messages and initiates graceful shutdown.

Registry images were built before these commits, which is why they work.

mvletter:
Closes #189

I ran into this in production: our bot was leaving meetings that were still
active. After digging into the logs, the participant count was dropping to
zero right around the time a new video track arrived, not when people
actually left.

Traced it to the video-blocking code in getVideoBlockInitScript. Two lines:

  event.track.stop()
  event.transceiver.direction = 'inactive'

track.stop() permanently destroys the WebRTC track. Google Meet detects the
broken state and rebuilds the participant DOM, momentarily clearing all
participant elements. The left_alone_timeout fires on that zero count and
the bot walks out of a perfectly active meeting.

The transceiver line makes it worse: setting direction to 'inactive' triggers
SDP renegotiation, which causes Google Meet to do the same DOM rebuild.

The fix is simple: just set track.enabled = false. That stops the bot from
processing incoming video frames without touching WebRTC state. No
renegotiation, no DOM rebuild, participant count stays accurate.

Tested in production. Browser refresh mid-meeting no longer causes the bot
to leave.

### DIFF
--- a/services/vexa-bot/core/src/services/screen-content.ts
+++ b/services/vexa-bot/core/src/services/screen-content.ts
@@ -1274,23 +1274,16 @@ export function getVideoBlockInitScript(): string {
         window.RTCPeerConnection = function(...args) {
           const pc = new OrigRTC(...args);
 
-          // Block incoming video: disable tracks AND stop transceivers
+          // Block incoming video: disable tracks only.
+          // Do NOT call track.stop() — it permanently destroys the track and causes
+          // Google Meet to detect a broken WebRTC state, clearing the participant DOM.
+          // Do NOT set transceiver.direction = 'inactive' — it triggers SDP renegotiation,
+          // which also causes Google Meet to rebuild the DOM and wipe participant elements.
+          // Both of these side effects cause left_alone_timeout false positives.
           pc.addEventListener('track', (event) => {
             if (event.track && event.track.kind === 'video') {
               event.track.enabled = false;
-              event.track.stop();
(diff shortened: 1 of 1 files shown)


DmitriyG228:
See my note on #196 — both superseded by main. Closing this one too. Thanks for the clean diff + explanatory comment.

PR #196, opened 2026-03-25
fix: bot stays in empty meeting / leaves when participant refreshes

### LINKED ISSUE #190: [Bug] Bot does not auto-leave Google Meet when all participants leave
## What happened?

The bot stays in the meeting indefinitely after all participants leave, only exiting ~15 minutes later via reconciliation scheduler instead of the configured everyoneLeftTimeout (default 60s).

## What did you expect?

1 Minute After Everybody Leaves the Meeting the Google Meet Bot should have exited the meeting. But it didn't for Around 15 Minutes.

## How to reproduce?

1. Build a Vexa Lite Image Locally Using Dockerfile.lite
2. Join a Google Meet
3. Make Everybody Leave the Meeting. Then wait and watch the logs

## Version / env?

Vexa Lite

Root Cause:

In services/vexa-bot/core/src/platforms/googlemeet/recording.ts, the extractParticipantsFromMain()  function counts participants by scanning all leaf text nodes in <main> with text length 2-50 characters. This picks up UI elements (button labels like "Mute", "Camera", "Present now", timestamps, tooltips, etc.) as"participants", 
so the count is always well above 1 even when the bot is alone in the meeting.

The monitoring check at line 806:
if (currentParticipantCount <= 1) {
 never triggers because the noisy text scan returns 10+ "participants" from UI elements alone.

mvletter:
Closes #190

Two opposite bugs, same root cause.

We had the bot staying in meetings long after everyone had left. We also had
it leaving while people were still there, triggered by someone refreshing
their browser.

Both trace back to how participants are counted. The current approach scans
all text nodes inside <main> and counts anything that looks like a name. The
problems with that:

It matches UI labels, tooltips, and other random text as participants. So
after everyone leaves, the bot still sees "participant count > 0" and never
triggers left_alone_timeout.

It does a point-in-time DOM query. During a browser refresh, Google Meet
briefly removes and redraws its participant elements. The scan hits that
window, gets zero, and the bot thinks it is alone.

The fix replaces text scanning with a MutationObserver keyed on
[data-participant-id], which is Google Meet's stable per-device identity
attribute. The observer fires on every DOM change so the registry is always
current. A 5-second grace timer absorbs transient removals like refreshes
and brief reconnects, so a participant only counts as gone after their
element has been absent for 5 seconds.

Tested in production. Bot stays when someone refreshes, leaves correctly
after 60 seconds when the meeting is genuinely empty.

### DIFF
--- a/services/vexa-bot/core/src/platforms/googlemeet/recording.ts
+++ b/services/vexa-bot/core/src/platforms/googlemeet/recording.ts
@@ -706,44 +706,51 @@ export async function startGoogleRecording(page: Page, botConfig: BotConfig): Pr
               initializeGoogleSpeakerDetection(whisperLiveService, audioService, botConfigData);
             }
 
-            // Simple single-strategy participant extraction from main video area
-            (window as any).logBot("Initializing simplified participant counting (main frame text scan)...");
+            // Event-driven participant registry using MutationObserver on data-participant-id.
+            // Replaces the text-scan approach which was unreliable: it matched UI labels,
+            // tooltips, and other text nodes as "participants", causing both false positives
+            // (bot stays when everyone left) and false negatives (bot leaves when someone refreshes).
+            // data-participant-id is Google Meet's stable, per-device identity attribute.
+            // A grace timer absorbs transient DOM removals (e.g. browser refresh, tab switch)
+            // so the bot doesn't leave during the reconnect window.
+            (window as any).logBot("Initializing participant registry (MutationObserver)...");
 
-            const extractParticipantsFromMain = (botName: string | undefined): string[] => {
-              const participants: string[] = [];
-              const mainElement = document.querySelector('main');
-              if (mainElement) {
-                const nameElements = mainElement.querySelectorAll('*');
-                nameElements.forEach((el: Element) => {
-                  const element = el as HTMLElement;
-                  const text = (element.textContent || '').trim();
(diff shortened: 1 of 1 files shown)


DmitriyG228:
Thanks for these two PRs — both are now superseded by separate commits that landed on `main` since you filed them.

- The text-scan → `data-participant-id` replacement you built in [#196](https://github.com/Vexa-ai/vexa/pull/196) was done separately (see [`services/vexa-bot/core/src/platforms/googlemeet/recording.ts:737`](https://github.com/Vexa-ai/vexa/blob/main/services/vexa-bot/core/src/platforms/googlemeet/recording.ts#L737)).
- The `track.stop()` / `transceiver.inactive` removal your [#195](https://github.com/Vexa-ai/vexa/pull/195) targets is already gone — the file now carries an explicit comment at [`services/vexa-bot/core/src/services/screen-content.ts:1304`](https://github.com/Vexa-ai/vexa/blob/main/services/vexa-bot/core/src/services/screen-content.ts#L1304) documenting exactly why those calls must not be added back.

Closing both as superseded-by-main. Your diagnosis in the PR bodies is credited — especially the `LEAVE_GRACE_MS` grace-timer idea in #196. If the current `data-participant-id` counting still has the "browser refresh wipes DOM" issue you identified, a fresh small PR adding a grace timer on top of it would land quickly. Would you like to explore that?

PR #260, opened 2026-04-25
fix(vexa-bot): browser-session entrypoint exits when node exits (#258)

### LINKED ISSUE #258: browser-session pod stays Running after node process exits — entrypoint keeps container alive forever for VNC
## Summary

`services/vexa-bot/core/entrypoint.sh` keeps a browser-session container Running indefinitely after the bot's `node dist/docker.js` process has exited cleanly. K8s never sees the pod terminate, so a pod that the user has stopped continues to consume a slot until something else kills it.

## Where

`services/vexa-bot/core/entrypoint.sh` lines 96–102 (browser-session-mode branch):

```bash
# Run node and keep container alive even if node crashes
echo "[entrypoint] Starting browser session node process..."
node dist/docker.js
EXIT_CODE=$?
echo "[entrypoint] node dist/docker.js exited with code $EXIT_CODE — keeping container alive for VNC access"
# Keep alive so VNC + websockify remain accessible
wait
```

`wait` (no args) waits for ALL background children. The browser-session entrypoint earlier started:

- `Xvfb :99` (line 9, background)
- `fluxbox` (line 70, background)
- `x11vnc` (line 71, background, `-forever`)
- `autocutsel` x2 (lines 72–73, background)
- `websockify` (lines 78/80, background)
- The CDP-readiness + socat subshell (line 88, background)
- `sshd` (line 94, daemon — but several of the above never exit)

`x11vnc -forever` and `websockify` never return. `wait` therefore blocks forever. The container only exits when something external kills the pod.

## Real-world impact (release-005 dogfood, 2026-04-25)

Two browser-session pods on staging were `Running` for 6+ hours after their `node` process exited cleanly. The user explicitly tried to **Stop** them via the dashboard:

- The Stop request reached the bot.
- The bot saved state and `node` exited with code 0.
- The pod stayed Running because `wait` blocks on `x11vnc` / `websockify`.

From the user's perspective, **Stop didn't work.** They had to wait for `idle_timeout` (or manually `kubectl delete pod`) to free the slot. This also defeats any concurrency cap a deployment relies on (orphaned pods count against the limit).

## Root cause

The `wait` call was added so VNC + websockify remain reachable for post-mortem debugging after `node` exits. But:

1. There is no time bound — debug access lasts forever, even after the user has stopped the session.
2. K8s readiness/liveness probes (if configured) would still see the pod as healthy.
3. There is no signal back to the runtime that the bot is done.

Cleanup should happen as soon as `node` exits (the bot has saved its state by then). Long-term debug-after-exit access can be opt-in.

## Proposed fix

Make the post-exit keep-alive duration configurable, default `0` (exit immediately).

**`services/vexa-bot/core/entrypoint.sh`** (browser-session branch):

```bash
node dist/docker.js
EXIT_CODE=$?
KEEPALIVE=\${VEXA_BROWSER_SESSION_VNC_KEEPALIVE_SECONDS:-0}
if [ \"\$KEEPALIVE\" -gt 0 ]; then
  echo \"[entrypoint] node exited with \$EXIT_CODE — keeping container alive \${KEEPALIVE}s for VNC access\"
  timeout \"\$KEEPALIVE\" wait || true
else
  echo \"[entrypoint] node exited with \$EXIT_CODE — exiting (set VEXA_BROWSER_SESSION_VNC_KEEPALIVE_SECONDS>0 to debug)\"
fi
exit \$EXIT_CODE
```

Properties:
- Default `0` → container exits as soon as `node` returns. Stop via dashboard works correctly.
- Operators who want the debug window set `VEXA_BROWSER_SESSION_VNC_KEEPALIVE_SECONDS=3600` (or any value) in the runtime profile.
- Exit code is preserved so K8s sees a clean exit (0) vs. a crash.

The env var can be plumbed through `services/runtime-api/runtime_api/profiles.py` profile `env:` map per the existing convention (no code change needed there — `profiles.py` already passes the `env` block straight into the container).

## Alternatives considered

1. **Time-bound the wait, hardcoded** (`timeout 3600 wait || true`). Simpler but inflexible.
2. **Drop the keep-alive entirely.** Simplest. Loses the post-exit VNC debug feature. Acceptable if no one uses it; we can bring it back via env var if someone asks.

We chose option 2 from the platform-side discussion (env-configurable, default 0) so the safe lifecycle is the default and the debug feature is preserved for those who need it.

## PR

PR with the option-2 fix incoming.

## Reporter

Filed from vexa-platform release-005 dogfood (2026-04-25). See vexa-platform `release/releases/release-005-rolling-refresh-and-retire/prod-feedback.md` entry F12 for the platform-side context.

DmitriyG228:
## Summary

- Fixes #258: `wait` in the browser-session entrypoint blocks forever on `x11vnc -forever` and `websockify`, so the container never exits after `node dist/docker.js` returns. K8s sees the pod as Running indefinitely and "stop" from the dashboard appears to do nothing.
- New env var `VEXA_BROWSER_SESSION_VNC_KEEPALIVE_SECONDS` (default `0`) controls a bounded post-exit debug window. Default behavior: container exits as soon as `node` does.
- Preserves the original exit code so K8s sees `0` for clean exits and `≠0` for crashes.

## Diff

`services/vexa-bot/core/entrypoint.sh` (browser-session branch only — meeting-mode branch unchanged):

```bash
node dist/docker.js
EXIT_CODE=$?
KEEPALIVE="${VEXA_BROWSER_SESSION_VNC_KEEPALIVE_SECONDS:-0}"
if [[ "$KEEPALIVE" =~ ^[0-9]+$ ]] && [ "$KEEPALIVE" -gt 0 ]; then
  echo "[entrypoint] node dist/docker.js exited with code $EXIT_CODE — keeping container alive ${KEEPALIVE}s for VNC access"
  sleep "$KEEPALIVE"
else
  echo "[entrypoint] node dist/docker.js exited with code $EXIT_CODE — exiting (set VEXA_BROWSER_SESSION_VNC_KEEPALIVE_SECONDS>0 to debug)"
fi
exit "$EXIT_CODE"
```

## Implementation note

The issue body sketched `timeout $N wait` — that doesn't actually work because `wait` is a shell builtin and `timeout` only runs binaries. `sleep $N` achieves the equivalent bounded keep-alive: the parent shell sleeps, then exits, and the kernel cleans up the still-running `x11vnc` / `websockify` / `sshd` children.

Non-numeric values fall through to the default branch (`[[ ... =~ ^[0-9]+$ ]]` guard) so a typo can't crash the script.

## How to opt back into the post-exit debug window

In `services/runtime-api/runtime_api/profiles.py` profile YAML:

```yaml
browser-session:
  image: ...
  env:
    VEXA_BROWSER_SESSION_VNC_KEEPALIVE_SECONDS: "3600"
```

`profiles.py` already passes the `env` block straight into the container — no code change needed there.

## Test plan

- [x] `bash -n services/vexa-bot/core/entrypoint.sh` (syntax)
- [x] Smoke-test the three branches (default-0, valid N, non-numeric) — verified the if/else picks the right branch and `EXIT_CODE` is preserved
- [ ] Build the runtime image and exercise a real browser session: confirm pod terminates within seconds of bot stop on staging
- [ ] Set `VEXA_BROWSER_SESSION_VNC_KEEPALIVE_SECONDS=60` in a profile, exercise stop, confirm container stays for ~60s then exits with the original exit code

### DIFF
--- a/services/vexa-bot/core/entrypoint.sh
+++ b/services/vexa-bot/core/entrypoint.sh
@@ -93,13 +93,22 @@ FBAPPS
   echo "[entrypoint] Starting SSH server on port 22"
   /usr/sbin/sshd
 
-  # Run node and keep container alive even if node crashes
+  # Run node, then exit so K8s sees the pod terminate.
+  # Optional post-exit VNC debug window: set
+  # VEXA_BROWSER_SESSION_VNC_KEEPALIVE_SECONDS to a positive integer to
+  # keep the container alive that long after node exits. Default 0 =
+  # exit immediately so that "stop" from the dashboard frees the slot.
   echo "[entrypoint] Starting browser session node process..."
   node dist/docker.js
   EXIT_CODE=$?
-  echo "[entrypoint] node dist/docker.js exited with code $EXIT_CODE — keeping container alive for VNC access"
-  # Keep alive so VNC + websockify remain accessible
-  wait
+  KEEPALIVE="${VEXA_BROWSER_SESSION_VNC_KEEPALIVE_SECONDS:-0}"
+  if [[ "$KEEPALIVE" =~ ^[0-9]+$ ]] && [ "$KEEPALIVE" -gt 0 ]; then
+    echo "[entrypoint] node dist/docker.js exited with code $EXIT_CODE — keeping container alive ${KEEPALIVE}s for VNC access"
+    sleep "$KEEPALIVE"
+  else
+    echo "[entrypoint] node dist/docker.js exited with code $EXIT_CODE — exiting (set VEXA_BROWSER_SESSION_VNC_KEEPALIVE_SECONDS>0 to debug)"
+  fi
+  exit "$EXIT_CODE"
 else
   # Meeting mode — start VNC for browser view on dashboard
   echo "[entrypoint] Meeting mode — starting VNC for browser view"


DmitriyG228:
Planned for merge in the current hardening release (cycle `260424`); closes #258.

DmitriyG228:

Issue #362, opened 2026-05-23
[Pack] PACK 7 - Release Identity And Packaging Hardening


DmitriyG228:
# Pack Epic: PACK 7 - Release Identity And Packaging Hardening

Status: proposed pack epic for `0.10.6.x` replay from `v0.10.6`.

## CEO outcome

Operators can tell exactly what candidate is running and ship or roll back with confidence.

## CTO outcome

Source, Docker image, Helm chart, dashboard version, CI, and hardening metadata identify the same candidate and avoid release-only drift.

## User outcome

Self-hosted and production operators see coherent version/package behavior instead of mixed source/runtime identity.

## Included raw issues / PRs

- #349
- release/provenance/hardening scope from #348/#331

## Reusable committed code references

Base: `v0.10.6^{}` = `6da3abb` / `6da3abb92fd50c1939f1c0ecfb85246db364e427`

Lineage head used for reusable committed work: `codex/release-0.10.6.2.1` = `e8cd9e7` / `e8cd9e73aa9967cd0ebb9f2ff3a0d701b9684106`

Reusable commits:
- `1707445`
- `4bff1e4`
- `43ed25f`
- `edefff3`
- `dbb42f7`

Main file surfaces:
- `VERSION`
- `RELEASE_NOTES.md`
- `deploy/compose/docker-compose.yml`
- `deploy/helm/charts/vexa/Chart.yaml`
- `deploy/helm/charts/vexa/templates/_helpers.tpl`
- `services/dashboard/scripts/assert-release-version.js`
- `services/dashboard/scripts/generate-release-version.js`
- `services/dashboard/src/lib/release-version.ts`
- `services/meeting-api/meeting_api/security_headers.py`
- `service Dockerfiles and dependency manifests touched by hardening commits`

## Explicitly out of scope

- Current uncommitted product changes in the local worktree are excluded from this pack.
- Dead release evidence folders are excluded as source truth.
- Deprecated harness evidence is excluded; reusable checks must live in product tests or dedicated skills.
- The removed realtime transcript regression pack is not replay scope here.

## Blast radius

Release provenance, Docker images, Helm chart SemVer/appVersion, dashboard version display, dependency floors, security headers, CI workflows.

## Data / schema / API / public-contract decisions

No product behavior pack is accepted by this pack alone; this pack verifies candidate identity, packaging, and hardening once product packs exist.

## Isolation requirements

- Develop from base `v0.10.6^{}` (`6da3abb`), not from current main.
[...]

PR #363, opened 2026-05-23
[Pack 7] Release identity and packaging hardening

DmitriyG228:
# Pack PR: [Pack] PACK 7 - Release Identity And Packaging Hardening

Pack epic: https://github.com/Vexa-ai/vexa/issues/362
Pack id: `0.10.6x-pack-7-release-identity-packaging-hardening`
Release: `0.10.6.x replay`
Base branch: `v0.10.6^{}`
Integration branch: `codex/release-0.10.6x-pack-integration`
Evidence: `.agents/packs/0.10.6x-pack-7-release-identity-packaging-hardening/`

## Outcomes

CEO: Operators can tell exactly what candidate is running and ship or roll back with confidence.

CTO: Source, Docker image, Helm chart, dashboard version, CI, and hardening metadata identify the same candidate and avoid release-only drift.

User: Self-hosted and production operators see coherent version/package behavior instead of mixed source/runtime identity.

## Scope

- #349
- release/provenance/hardening scope from #348/#331

## Out of scope

- Current uncommitted product changes in the local worktree are excluded from this pack.
- Dead release evidence folders are excluded as source truth.
- Deprecated harness evidence is excluded; reusable checks must live in product tests or dedicated skills.
- The removed realtime transcript regression pack is not replay scope here.

## Blast radius

Release provenance, Docker images, Helm chart SemVer/appVersion, dashboard version display, dependency floors, security headers, CI workflows.

## Validation

Synthetic: pass. Covered source identity proof, dashboard release generator/assertion proof, meeting-api security header unit tests, Helm lint/template, and npm lock consistency dry-runs.

Compose: machine config proof pass. Human Compose blast-radius eyeball validation and Compose live meeting deployment validation are pending and block PR-ready status.

Lite: source-level identity proof pass. Human Lite blast-radius eyeball validation and Lite live meeting deployment validation are pending and block PR-ready status.

Live/human: live meeting deployment validation is now mandatory for every pack against both Compose and Lite. Google Meet `hsg-wddw-jhz` was approved for this validation. Overall human eyeball validation is mandatory and pending.

Hardenloop: ran with `--fix none`; release blockers `0`; decision `incomplete_coverage` because several optional local scanners are not installed. Private advisory/raw finding payloads were omitted from committed evidence; see `hardenloop/state.md`.

## Evidence checklist

- [x] `pack.json`
- [x] `runtime.json`
- [x] `ops/ops.jsonl`
- [x] `tests`
- [x] `compose`
- [ ] `compose/meeting-deployment-test.md` has `Status: pass`
- [ ] `compose/human-eyeball.md` has `Status: pass`
- [x] `lite`
- [ ] `lite/meeting-deployment-test.md` has `Status: pass`
- [ ] `lite/human-eyeball.md` has `Status: pass`
- [ ] `human/overall-functionality.md` has `Status: pass`
- [x] `hardenloop`
- [x] `review.md`
- [x] `pr.md`

## Stitching notes

- Shared files must be merged by hunk and reviewed against adjacent pack edits.
- If stitching exposes a behavior bug, route it back to the responsible pack instead of hiding a stitch-time code change.
- This epic is not a release sign-off; it is the delivery contract for one independently reviewable pack.

## Review checklist

- [x] Pack branch starts from `v0.10.6^{}`.
- [x] Only this pack's committed reuse hunks are replayed.
- [x] Synthetic checks pass before live/human checks.
- [ ] Compose gate includes passed human eyeball blast-radius validation.
- [ ] Lite gate includes passed human eyeball blast-radius validation.
- [ ] Compose live meeting deployment validation passed.
- [ ] Lite live meeting deployment validation passed.
- [ ] Overall functionality human eyeball validation passed.
- [x] Hardenloop is run for the pack.
- [x] PR body links this epic and evidence root.
- [x] Reviewer can map each reused hunk back to the commit list above.


DmitriyG228:

Scope: `git diff v0.10.6..HEAD -- services/ libs/ packages/ deploy/` on the pack-7 (release identity + packaging hardening) branch. Findings ranked most-severe first.

| # | Severity | File | Line | Summary | Failure scenario |
|---|---|---|---|---|---|
| 1 | HIGH | `deploy/helm/charts/vexa/templates/_helpers.tpl` | ~97-106 (`vexa.botImage`) | New helper inverts override precedence: `global.imageTag` short-circuits before `runtimeApi.browserImage` / `meetingApi.botImageName`. | Operator pinning all services with `global.imageTag=v0.10.6.2.1` plus a private-mirror `runtimeApi.browserImage=registry.internal/vexa-bot:pinned-sha` will have the helper render `vexaai/vexa-bot:v0.10.6.2.1` (public registry) and silently ignore their browserImage override for both `BOT_IMAGE_NAME` (meeting-api) and `BROWSER_IMAGE` (runtime-api). Previously these two overrides were honored directly. Behavior change is not documented in values.yaml. |
| 2 | HIGH | `deploy/helm/charts/vexa/Chart.yaml` | 5 | Chart `version: 0.10.6+2.1` uses SemVer build-metadata syntax (`+`). | Helm CLI accepts `+` but several common downstream tools mangle it: `helm push` to OCI registries URL-encodes/strips build metadata (the pushed artifact lands at tag `0.10.6_2.1` or `0.10.6` depending on registry), ArgoCD's chart resolver historically chokes on `+`, and `helm dependency update` can't pin a child chart by a `+` version on some registries. Operators upgrading from `0.10.5` may pull a chart that they cannot reference back. The pack does not touch any release tooling to guard this. |
| 3 | HIGH | `services/meeting-api/meeting_api/security_headers.py` | 23 | `referer_url.hostname != request.url.hostname` blocks the same-host CSP allowance whenever meeting-api sits behind a reverse proxy that does not rewrite `request.url`. | In any K8s ingress or compose-with-frontend-proxy deployment, `request.url.hostname` is the internal service hostname (e.g. `meeting-api`, pod IP, or `0.0.0.0` from uvicorn) while the browser's `Referer` is the public dashboard URL. The check therefore rejects the legitimate referer, leaves CSP at `'self' http://localhost:* https://localhost:*`, and the dashboard iframe of `/b/.../vnc/...` is blocked in production — exactly the bug this code claims to fix. Tests pass because `TestClient(base_url=...)` makes the two match. Needs `X-Forwarded-Host` handling or a values-driven allowlist. |
| 4 | MEDIUM | `services/dashboard/scripts/generate-release-version.js` | ~118-122 | `releaseDate` fallback writes today's build date as the user-visible "release date" whenever no env var is supplied and `.git/` is absent (which it always is — see `.dockerignore` line 2). | Operators building the dashboard image from a release tarball with `VERSION=0.10.6.2.1` but no `NEXT_PUBLIC_VEXA_OSS_RELEASE_DATE` will get a sidebar that says "released on <today's build date>". That misrepresents release identity — the exact thing pack 7 is supposed to guarantee. The companion `assert-release-version.js` only validates the version string, never the date. |
| 5 | MEDIUM | `services/dashboard/scripts/assert-release-version.js` | 60-64 | "Bundle contains the version string" check is a substring scan with no boundaries, easily satisfied by an unrelated literal. | The needle is `"0.10.6.2.1"` (quoted). Any unrelated Next.js bundle string that contains this substring — e.g. a sourcemap path, a Sentry release tag, a copy-pasted CHANGELOG fragment — passes the assertion even if the actual release-version export is wrong (stale `release-version.generated.json` not re-imported, tree-shaken away, etc.). Recommend asserting on the generated JSON's exact key/value pair (`"version":"0.10.6.2.1"`) or grepping the server-side bundle that imports `release-version.ts`. |
| 6 | MEDIUM | `deploy/compose/docker-compose.yml` | 348 | `NEXT_PUBLIC_VEXA_OSS_VERSION: ${VEXA_VERSION:-}` passes the empty string as a build-arg. | The generator's `normalizeVersion('')` returns `null`, so empty falls through to `VERSION` file — OK. BUT `process.env.NEXT_PUBLIC_VEXA_OSS_VERSION` is now the empty string, and the guard `if (process.env.NEXT_PUBLIC_VEXA_OSS_VERSION && !envVersion)` evaluates the empty string as falsy, so the "invalid override" error never fires. That's correct, but operators who *intend* to override with a malformed value (e.g. `VEXA_VERSION=0.10.6.2.1-rc1`) will hit the "invalid" error and may set `VEXA_VERSION=""` to escape, silently shipping VERSION-file content under a misleading-named env. Inversion of cause and effect; the empty-string sentinel deserves at least a `.env` comment. |
| 7 | LOW | `deploy/lite/Dockerfile.lite` | 160 | `npm install --legacy-peer-deps` was added unconditionally. | `--legacy-peer-deps` masks real peer-dep conflicts. Combined with the upgraded `next 16.2.6` / `eslint-config-next 16.2.6` / `nodemailer 7.0.13`, any future minor bump may silently regress the dashboard build. The main dashboard `Dockerfile` did not add this flag — Lite and Compose dashboards now use *different* resolution strategies, so what builds in Compose may not build in Lite and vice versa. |
| 8 | LOW | `services/dashboard/src/lib/release-version.ts` | 33 | `releaseUrl()` now prepends `v` unless the version starts with `v`, but Chart appVersion is `0.10.6.2.1` (4 dotted components). | GitHub tag `v0.10.6.2.1` may not exist for some packs (release tags historically `v0.10.x` 3-component). If the OSS release tag is `v0.10.6` while appVersion is `0.10.6.2.1`, the sidebar "release notes" link 404s. Worth verifying the release process tags `v0.10.6.2.1` exactly, or having the link target the major release. |

### Cross-cutting observations
- Pack scope is mostly correct: VERSION → Chart.yaml → generator → assert pipeline is a real improvement over the prior "Chart.yaml only" path.
- The api-gateway Dockerfile `chmod -R a+rX /app` after switching to `appuser` is fine but only fixes read; if any service writes under `/app` (e.g. tempfile, log file), they will get EACCES with no diagnostic.
- runtime-api CI workflow correctly pins action SHAs and drops `persist-credentials` — good hardening, no findings.
- The user's prompt mentioned `test_header_injection.py` was touched; the actual diff shows `test_security_headers.py` (new file) rather than a header-injection test. The CSP-builder string interpolation of `referer_url.hostname`/`port` was reviewed for injection risk and is bounded by `urlparse` + scheme/port validation, but a follow-up explicit test for malicious referer headers (CR/LF, exotic IDN, `userinfo@`) would close the loop.

---
Draft findings produced by automated finder sweep. **Human verdict required** per develop skill code-review.md gate.

DmitriyG228:

Pack 7 product diff reviewed: `git diff v0.10.6..HEAD -- services/ libs/ packages/ deploy/`. Findings ordered most-severe first.

---

### 1. Helm chart `version: 0.10.6+2.1` breaks OCI push and CLI version pinning
`deploy/helm/charts/vexa/Chart.yaml` now sets `version: 0.10.6+2.1`. Helm chart `version` MUST be SemVer 2, and while SemVer allows `+` for build metadata, OCI tag charset does **not** allow `+` — `helm push <chart>.tgz oci://<registry>` will be rejected (the chart filename itself becomes `vexa-0.10.6+2.1.tgz` which many tarball workflows mishandle). The template already has the tell: `helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}` (line 22 of `_helpers.tpl`) — meaning whoever wrote this knew `+` is unsafe in label values, but did not extend that knowledge to the registry-tag and `--version` matching layers. **Failure scenario**: `make release-helm-push` (or any `helm push` to GHCR/ECR/etc.) fails with "invalid tag" once the registry layer is hit, and `helm install vexa --version 0.10.6+2.1` will fail or silently match a different version because clients percent-encode or strip `+`.

### 2. `vexa.botImage` precedence regresses runtime-api `BROWSER_IMAGE` default
The new helper in `_helpers.tpl`:
```
{{- if .Values.global.imageTag -}}vexaai/vexa-bot:{{tag}}
{{- else if .Values.runtimeApi.browserImage -}}…
{{- else if .Values.meetingApi.botImageName -}}…
{{- else -}}meetingApi.image.repository:tag
```
…is wired into `deployment-runtime-api.yaml` for `BROWSER_IMAGE`. But `values.yaml` ships **`meetingApi.botImageName: "vexaai/vexa-bot:latest"` as the hard default**. With a vanilla `helm install` (no `global.imageTag`, no `runtimeApi.browserImage`), the previous code resolved `BROWSER_IMAGE` to `vexaai/meeting-api:<tag>` (`runtimeApi.browserImage | default printf meetingApi.image`). The new code resolves it to `vexaai/vexa-bot:latest`. That is (a) a different image entirely and (b) silently floating on `:latest` regardless of the chart appVersion. **Failure scenario**: a clean helm install for 0.10.6.2.1 launches browser/bot containers pinned to `:latest`, defeating the whole point of release-identity hardening on the runtime path. Tag drift between launcher and bot image becomes invisible.

### 3. `global.imageTag` silently overrides explicit `runtimeApi.browserImage`
Same helper: `global.imageTag` is FIRST in precedence. If an operator deliberately sets `runtimeApi.browserImage: my-registry/custom-bot:abc123` to test a fork, but `global.imageTag` is also set (which is common — release scripts set it from `Chart.appVersion`), the explicit `browserImage` is silently ignored and replaced with `vexaai/vexa-bot:<global.imageTag>`. The old code respected `runtimeApi.browserImage` unconditionally. **Failure scenario**: a release-engineer pins a custom bot image for canary; helm renders `vexaai/vexa-bot:0.10.6.2.1` instead; the canary tests the upstream bot, not the fork.

### 4. `_same_host_frame_ancestor` breaks behind a reverse proxy
`services/meeting-api/meeting_api/security_headers.py` compares `referer_url.hostname != request.url.hostname`. In production deployments meeting-api sits behind nginx/ingress that terminates TLS; `request.url.hostname` then reflects the internal ASGI hostname (e.g. `meeting-api` service name, or `0.0.0.0`), not the public host the browser used. The referer Origin will be the public host. **Failure scenario**: the same-host VNC iframe embedding the new tests claim to enable will work in the unit test (`base_url="http://172.238.172.98:8056"`) but fail in any real Helm/Compose deployment fronted by a reverse proxy — dashboard at `https://app.example.com:3000` referrer vs. `request.url.hostname=meeting-api` never matches, CSP omits the dashboard origin, browser blocks the iframe. The middleware needs to honor `X-Forwarded-Host` or trust a configured allowed-origin list rather than `request.url.hostname`.

### 5. `releaseUrl()` will 404 for the new 5-component version on GitHub
`services/dashboard/src/lib/release-version.ts` builds `https://github.com/Vexa-ai/vexa/releases/tag/v0.10.6.2.1`. The repo's tag list only contains 3-component tags (`v0.10.6.2` is the deepest existing tag) — there is no `v0.10.6.2.1` release tag and the GitHub UX is to click the sidebar version chip → 404. This is the exact visible "release identity" surface that the pack is claiming to harden. **Failure scenario**: every user clicking the sidebar version pill on a 0.10.6.2.x dashboard lands on a GitHub 404.

### 6. `assert-release-version.js` needle search is fragile and weakly bound
The bundle check does `bundleFiles.some(file => content.includes('"0.10.6.2.1"'))`. (a) `walk()` only scans `.next/static/**/*.{js,json}`; if a future Next.js change moves the literal into `.next/server/**` (App Router server components routinely embed `process.env.NEXT_PUBLIC_*` server-side), the assertion silently flips from "verified in bundle" to "no files contain the needle" → build fails even when the version is correct (or, more dangerously, a regex/minifier escape produces `\"0.10.6.2.1\"` and the substring check passes anyway because JSON-encoded JS source still contains the unescaped form somewhere). (b) The needle is a plain version string — any unrelated literal (e.g. a Helm chart filename in a doc snippet, a CHANGELOG entry inlined as a string) will satisfy it. **Failure scenario**: the assertion gives a false sense of "the bundle ships X" while actually only proving "some file in `.next/static` mentions X anywhere".

### 7. `tagCommitDate(\`v${version}\`)` always falls through for 5-component versions
`generate-release-version.js` now passes `\`v${version}\`` (e.g. `v0.10.6.2.1`) to `git log -1 --format=%cs v0.10.6.2.1`. The `--match 'v[0-9]*.[0-9]*.[0-9]*'` glob in `latestGitTag` is 3-component only, and no 5-component tag exists in the repo. Git silently returns nothing → catch returns null → release date falls through to `new Date().toISOString().slice(0,10)` (today). **Failure scenario**: every rebuild of the same release shows a different "released YYYY-MM-DD" in the sidebar chip because the date is "today" rather than the tag's commit date. Dashboards built on 2026-05-24 will display `released 2026-05-24`; the same source rebuilt on 2026-06-01 will display `released 2026-06-01`. The version-chip tooltip claim "released …" becomes a lie.

### 8. Compose build-arg now uses empty-string default, masking misconfigurations
`deploy/compose/docker-compose.yml` changed `NEXT_PUBLIC_VEXA_OSS_VERSION: ${VEXA_VERSION:-dev}` → `${VEXA_VERSION:-}`. The comment says "Empty VEXA_VERSION is allowed; the generator then uses VERSION." That works only because the Dockerfile now `COPY VERSION /repo/VERSION`. But the generator's strictness check is `if (process.env.NEXT_PUBLIC_VEXA_OSS_VERSION && !envVersion)` — an empty string is falsy, so it is treated as "unset" and the override-mismatch guard is skipped. Effect: if someone exports `VEXA_VERSION=` (empty) in CI to "disable" the override, the build silently uses the VERSION file instead of failing loud. More concretely: a CI job that exports a malformed `VEXA_VERSION=0.10.6-rc1` (which `normalizeVersion` rejects → returns null → falsy) will trigger the "invalid NEXT_PUBLIC_VEXA_OSS_VERSION" throw — good. But `VEXA_VERSION=""` from a missed `.env` file load passes silently. **Failure scenario**: dashboard image is built in a CI environment where the `.env` step was skipped; `VEXA_VERSION` is empty; build succeeds using `/repo/VERSION`, but the operator believed they were pinning via env. Detection is harder than the pack intends.

---

DmitriyG228:
Closing as part of 2026-06-11 WIP triage: the 0.10.6x PACK 1–7 stitching series sat 14+ days without movement, and most of it has merge conflicts against current main. Only PACK 5 (#369) survives — adopted into the active billing lane (pack #401). The branch for this PR is preserved; if this pack's scope comes back, re-open a fresh PR from a rebased branch when it claims a WIP slot.

Issue #359, opened 2026-05-23
[Pack] PACK 4 - Stop/Delete Lifecycle Convergence


DmitriyG228:
# Pack Epic: PACK 4 - Stop/Delete Lifecycle Convergence

Status: proposed pack epic for `0.10.6.x` replay from `v0.10.6`.

## CEO outcome

Stopping a meeting should reliably end the session instead of leaving stuck runtime or bot resources.

## CTO outcome

Runtime callbacks, internal secret preservation, and lifecycle metadata converge stop/delete into terminal meeting state without stale sessions.

## User outcome

When a bot is stopped or deleted, the meeting reaches a final state and resources are cleaned up.

## Included raw issues / PRs

- #313
- lifecycle/callback portions of #331/#347

## Reusable committed code references

Base: `v0.10.6^{}` = `6da3abb` / `6da3abb92fd50c1939f1c0ecfb85246db364e427`

Lineage head used for reusable committed work: `codex/release-0.10.6.2.1` = `e8cd9e7` / `e8cd9e73aa9967cd0ebb9f2ff3a0d701b9684106`

Reusable commits:
- `a11fa17`
- `1ca1ecb`

Main file surfaces:
- `services/runtime-api/runtime_api/api.py`
- `services/runtime-api/runtime_api/lifecycle.py`
- `services/runtime-api/tests/test_lifecycle.py`
- `services/vexa-bot/core/src/docker.ts`
- `services/vexa-bot/core/src/index.ts`
- `services/vexa-bot/core/src/services/unified-callback.ts`
- `services/vexa-bot/core/src/types.ts`

## Explicitly out of scope

- Current uncommitted product changes in the local worktree are excluded from this pack.
- Dead release evidence folders are excluded as source truth.
- Deprecated harness evidence is excluded; reusable checks must live in product tests or dedicated skills.
- The removed realtime transcript regression pack is not replay scope here.

## Blast radius

Runtime API lifecycle, bot callback payloads, meeting status convergence, cleanup sweeps, stale browser/bot sessions.

## Data / schema / API / public-contract decisions

Internal callback secret behavior is preserved; billing/webhook event counting is explicitly out of this pack.

## Isolation requirements

- Develop from base `v0.10.6^{}` (`6da3abb`), not from current main.
- One branch/worktree for this pack only.
- Evidence root: `.agents/packs/0.10.6x-pack-4-stop-delete-lifecycle-convergence/`.
[...]

Issue #313, opened 2026-05-04
browser_session DELETE leaves meeting stuck in 'stopping': runtime-api skips exit callback (no connection_id) AND Pack E.3.2 sweep defeated by webhook-retry updated_at bumps


DmitriyG228:
## Symptom

`browser_session` meeting 11536 (mid=11536, user=23, May 4 09:36:37 UTC) stuck in `status='stopping'` for 38+ minutes after the user issued `DELETE /bots/browser_session/bs-9041d935`. Container was deleted by runtime-api at 09:37:40 UTC (HTTP 200 ack); meeting status never advanced from `stopping → completed`. Pack E.3.2 stale-stopping sweep (5-min threshold) did NOT trigger.

### Bug 1 — runtime-api skips exit callback for browser_session

`meeting-api/meetings.py:818-823` creates browser_session containers with metadata containing **only `meeting_id`**:

```python
result = await _spawn_via_runtime_api(
    profile="browser-session",
    config={"image": BOT_IMAGE_NAME, "env": {...}},
    user_id=current_user.id,
    callback_url=f"{MEETING_API_URL}/bots/internal/callback/exited",
    metadata={"meeting_id": new_meeting.id},   # ← no connection_id
)
```

Compare with the regular bot path (`meetings.py:~1183`):

```python
metadata={"meeting_id": meeting_id, "connection_id": connection_id},
```

`runtime-api/lifecycle.py:200-203` then unconditionally drops the exit callback whenever `connection_id` is missing from container metadata:

```python
metadata = container_data.get("metadata", {})
if not metadata.get("connection_id"):
    logger.warning(f"No connection_id in metadata for {name} — skipping exit callback")
    return
```

Production log at 09:38:41 confirms: `WARNING runtime_api.lifecycle: No connection_id in metadata for browser-session-11536-341aad61 — skipping exit callback`.

Without the exit callback, the canonical `stopping → completed` transition (handled by `callbacks.bot_exit_callback`) never fires.

### Bug 2 — Pack E.3.2 stale-stopping sweep defeated by webhook-retry updated_at bumps

`meeting-api/sweeps.py:80-84` filter:

```python
.where(Meeting.status == MeetingStatus.STOPPING.value)
.where(Meeting.updated_at < threshold)  # threshold = now - 300s
```

`meetings.updated_at` is auto-bumped by SQLAlchemy on **any** UPDATE to the row, including:
- Webhook retry workers updating `data->'webhook_deliveries'`
[...]

PR #365, opened 2026-05-23
[Pack] PACK 4 - Stop/Delete Lifecycle Convergence

DmitriyG228:
# Pack PR: [Pack] PACK 4 - Stop/Delete Lifecycle Convergence

Pack epic: https://github.com/Vexa-ai/vexa/issues/359
Pack id: `0.10.6x-pack-4-stop-delete-lifecycle-convergence`
Release: `0.10.6.x replay`
Base branch: `v0.10.6^{}`
Integration branch: `codex/release-0.10.6x-pack-integration`
Evidence: `.agents/packs/0.10.6x-pack-4-stop-delete-lifecycle-convergence/`

## Outcomes

CEO: Stopping a meeting should reliably end the session instead of leaving stuck runtime or bot resources.

CTO: Runtime callbacks, internal secret preservation, and lifecycle metadata converge stop/delete into terminal meeting state without stale sessions.

User: When a bot is stopped or deleted, the meeting reaches a final state and resources are cleaned up.

## Scope

- #313
- lifecycle/callback portions of #331/#347

## Out of scope

- Current uncommitted product changes in the local worktree are excluded from this pack.
- Dead release evidence folders are excluded as source truth.
- Deprecated harness evidence is excluded; reusable checks must live in product tests or dedicated skills.
- The removed realtime transcript regression pack is not replay scope here.

## Blast radius

Runtime API lifecycle, bot callback payloads, meeting status convergence, cleanup sweeps, stale browser/bot sessions.

## Validation

Synthetic: pass
- `runtime-api`: 32 passed in `tests/test_api.py tests/test_lifecycle.py tests/test_state.py`
- `meeting-api`: 83 passed in `tests/test_callbacks.py tests/test_meetings.py tests/test_sweeps_stopping.py`
- `vexa-bot` callback/header focused checks passed.

Compose: pass
- Isolated project `vexa_0-10-6x-pack-4-stop-delete-lifecycle-convergence_compose`
- Rebuilt pack-branch `meeting-api` and `runtime-api` images under `0.10.6-pack4-260523-1521`.
- Browser session delete proof reached `completed`; Runtime API returned 404 for the deleted container.
- `make all` service health passed for API Gateway/Admin/Dashboard, then failed transcription smoke on local placeholder token HTTP 401.

Lite: blocked by isolation
- Official Lite `make lite` path uses fixed container names `vexa-lite`/`vexa-postgres`, host networking, and fixed ports 3000/8056/8057.
- Those names/ports are already occupied by unrelated local Lite lanes. Running it would violate pack isolation.

Live/human: not required; this pack was validated with synthetic and local Compose lifecycle checks.

Hardenloop: no release blockers, incomplete local scanner coverage
- Final run output: `.agents/packs/0.10.6x-pack-4-stop-delete-lifecycle-convergence/hardenloop/run-final`
- `bandit` ran; several external scanners were missing/unavailable locally.

Known caveat:
- Full `services/vexa-bot/core` build is blocked by pre-existing TypeScript issues in `src/index.ts` and local Playwright type resolution. Pack callback/header files pass focused checks.

## Evidence checklist

- [x] `pack.json`
- [x] `runtime.json`
- [x] `ops/ops.jsonl`
- [x] `tests`
- [x] `compose`
- [x] `lite`
- [x] `hardenloop`
- [x] `review.md`
- [x] `pr.md`

## Stitching notes

- Shared files must be merged by hunk and reviewed against adjacent pack edits.
- If stitching exposes a behavior bug, route it back to the responsible pack instead of hiding a stitch-time code change.
- This epic is not a release sign-off; it is the delivery contract for one independently reviewable pack.

## Review checklist

- [ ] Pack branch starts from `v0.10.6^{}`.
- [ ] Only this pack's committed reuse hunks are replayed.
- [ ] Synthetic checks pass before live/human checks.
- [ ] Compose gate is passed or explicitly marked not required in PR evidence.
- [ ] Lite gate is passed or explicitly marked not required in PR evidence.
- [ ] Hardenloop is run for the pack.
- [ ] PR body links this epic and evidence root.
- [ ] Reviewer can map each reused hunk back to the commit list above.


DmitriyG228:

Reviewing pack 4 (Stop/Delete Lifecycle Convergence): commits fa86a09 (callback auth), aa844ab (container-state removal timing), 7b2293d (browser session stop convergence). Scope: `services/`, `libs/`, `packages/`, `deploy/`.

Findings ranked most-severe first. Each is a candidate to be confirmed by human review.

### 1. HIGH — DELETE `/containers/{name}` races the natural-exit handler and can deliver the exit callback twice
- **file:** `services/runtime-api/runtime_api/api.py`
- **line:** ~336–353
- **failure_scenario:** A user-issued DELETE arrives just as the bot is already exiting (graceful leave, OOM, evict). The order is: (1) DELETE reads `data` (no `delete_requested` yet), (2) `backend.stop()` is a no-op because the container has already begun exiting, (3) the docker/k8s event listener fires `handle_container_exit` concurrently — it re-reads `container_data` and sees `delete_requested=false`, so it falls through to `_fire_exit_callback`. (4) DELETE proceeds and *also* fires `_fire_exit_callback`. Both writes target the same `pending_callback` key; `_deliver_callback` is invoked twice, and meeting-api receives two `/bots/internal/callback/exited` POSTs (the second one with possibly a different `exit_code` field). The `delete_requested` flag only protects against the case where the natural exit handler runs *after* DELETE has written the flag — there is no mutex against the inverse ordering.

### 2. HIGH — `INTERNAL_API_SECRET` snapshot at module-import time can fail-closed on every callback in Compose if env ordering is wrong
- **file:** `services/meeting-api/meeting_api/collector/auth.py`
- **line:** 11
- **failure_scenario:** `INTERNAL_API_SECRET = os.environ.get("INTERNAL_API_SECRET", "")` is captured at import. If `collector/auth.py` is imported before the env is populated (test harness, lazy import path, custom entrypoint), it caches `""`. Then in non-DEV_MODE every internal callback route — now including the ENTIRE `callbacks.py` router (Pack 4 made it `dependencies=[Depends(require_internal_secret)]`) — returns 503 "INTERNAL_API_SECRET not configured" regardless of the actual env. Result: all bot lifecycle callbacks fail, meetings stuck in `joining`/`active`/`stopping`. Recommend reading from `os.environ` inside the dep, not at module scope.

### 3. HIGH — `_fire_exit_callback` blocks the DELETE HTTP response for the full retry burst
- **file:** `services/runtime-api/runtime_api/api.py`
- **line:** 350
- **failure_scenario:** `await _fire_exit_callback(redis, name, exit_code=0)` synchronously calls `_deliver_callback`, which can sleep through `CALLBACK_RETRIES * CALLBACK_BACKOFF` (default 1+5+30 = 36 s plus three 10 s HTTP timeouts → up to ~60 s). The DELETE caller's `httpx` client in meeting-api uses `timeout=30.0` in `_stop_via_runtime_api`, so it times out, marks the stop as failed, and the outbox retries. Meanwhile the original DELETE's `state.delete_container` and `backend.remove` may or may not have run depending on where the retry burst eventually returns. Either fire-and-forget the callback or run delivery in a background task; do not block the DELETE path on consumer reachability.

### 4. HIGH — Bot's exit-callback secret falls back to `process.env.INTERNAL_API_SECRET`, but runtime-api never injects it into the container env
- **file:** `services/vexa-bot/core/src/services/unified-callback.ts`
- **line:** 157
- **failure_scenario:** `meetings.py:1133` adds `INTERNAL_API_SECRET` to `env_vars` ONLY for the recording-bot branch around line 1130. The browser-session spawn (line 823) does not inject it into `env`, and the agent-mode spawn (line 744) also doesn't. So `botConfig.internalSecret` is the *only* source for those bots. If a future refactor drops the `internalSecret` from the JSON payload or the bot loads a re-init config (`performGracefulLeave` rebuilds botConfig from `currentBotConfig?.internalSecret` — fine, but any code path that constructs a botConfig from scratch without the field will silently send no header and meeting-api will 403 the callback, leaving the meeting stuck.

### 5. MEDIUM — Stale-stopping sweep can starve tail rows; no ORDER BY plus early break at 50 means the same 50 candidates are processed forever
- **file:** `services/meeting-api/meeting_api/sweeps.py`
- **line:** 83–113
- **failure_scenario:** Query `select(Meeting).where(status='stopping').where(created_at < threshold).limit(200)` has no `ORDER BY`. After widening the per-iter cap from 50 → 200 candidates with an inner break at 50 swept rows, if Postgres consistently returns the same first 200 rows (typical with a `seq_scan` or an unordered btree scan) and 50+ of them are still in `stopping` for any reason (e.g. a downstream RPC bug that keeps re-entering `stopping`), the loop never reaches older stuck rows past those 200. Add `ORDER BY created_at ASC` for determinism and worst-case fairness.

### 6. MEDIUM — `_find_meeting_by_session("bs:<id>")` returns the meeting with no tenant/user scoping
- **file:** `services/meeting-api/meeting_api/callbacks.py`
- **line:** 272–279
- **failure_scenario:** Pre-Pack-4, callbacks resolved via `MeetingSession.session_uid`, which is a long opaque UUID. The new `bs:<int>` shortcut performs a bare `db.get(Meeting, meeting_id)` with no check that the caller's `connection_id` matches the meeting's mode/user. Inside the trust boundary (internal-secret gate) this is acceptable in steady state, but if the secret ever leaks (e.g. via a logged header, the `lite-internal-secret` hardcode in `deploy/lite/supervisord.conf`, or test fixtures), any caller with the secret can mutate ANY browser-session meeting's status. Worse, the integer space is enumerable — vs the UUID path which is computationally unguessable. Consider a HMAC of the meeting_id + a per-meeting nonce.

### 7. MEDIUM — `delete_pending_callback` is never explicitly called from the DELETE path; reliance on `_deliver_callback`'s success branch + idle_loop sweep means a partial failure can replay the exit forever after state is gone
- **file:** `services/runtime-api/runtime_api/api.py`
- **line:** 348–351
- **failure_scenario:** Sequence: DELETE fires `_fire_exit_callback`, the burst exhausts (consumer down), the pending-callback record is left for the idle_loop sweep. DELETE then calls `state.delete_container(redis, name)`. Hours later the consumer recovers; idle_loop sweep calls `_deliver_callback(name)` which posts `/bots/internal/callback/exited` with the original payload. But meeting-api has since reused the same container name (`f"bot-{user_id}-..."` patterns are deterministic), and the new meeting maps to the same name → the exit callback for the OLD container terminates a fresh meeting. Mitigation: include a monotonic spawn epoch or `container_id` in the callback payload and have meeting-api drop the callback if it does not match the current bot for that meeting.

### 8. LOW — `handle_container_exit` early-return on missing state means the listener silently drops exit data on a Redis flush/eviction
- **file:** `services/runtime-api/runtime_api/lifecycle.py`
- **line:** 184–187
- **failure_scenario:** New guard: `if not container_data: logger.info(...); return`. If Redis evicts the key (LRU under memory pressure, manual `FLUSHDB`, or the new `delete_container` having already run from the DELETE path), the exit event is dropped *without* firing the callback. For the DELETE-path case this is intentional (delete path owns the callback). For the eviction case, the meeting is silently lost: no callback, no `set_stopped`, the meeting row stays `active` until the stale-stopping sweep eventually (after 5 min in `stopping`, but never if it's still `active`) reconciles it. Recommend distinguishing "delete-owned" from "state-missing"; e.g. always emit a structured ERROR log with the missing-name, and have meeting-api's reconciler look for orphaned bot_container_ids.

---

DmitriyG228:
Closing as part of 2026-06-11 WIP triage: the 0.10.6x PACK 1–7 stitching series sat 14+ days without movement, and most of it has merge conflicts against current main. Only PACK 5 (#369) survives — adopted into the active billing lane (pack #401). The branch for this PR is preserved; if this pack's scope comes back, re-open a fresh PR from a rebased branch when it claims a WIP slot.

Issue #360, opened 2026-05-23
[Pack] PACK 5 - Billing And Webhook Idempotency


DmitriyG228:
# Pack Epic: PACK 5 - Billing And Webhook Idempotency

Status: proposed pack epic for `0.10.6.x` replay from `v0.10.6`.

## CEO outcome

Completion events must not over-bill or double-count usage when terminal paths retry or race.

## CTO outcome

Meeting API claims billing-sensitive completion events producer-side while keeping public webhook payloads and headers stable.

## User outcome

Customers see trustworthy usage/billing events and webhook delivery history without duplicate completion emissions.

## Included raw issues / PRs

- #330
- webhook/post-meeting portions of #331/#347

## Reusable committed code references

Base: `v0.10.6^{}` = `6da3abb` / `6da3abb92fd50c1939f1c0ecfb85246db364e427`

Lineage head used for reusable committed work: `codex/release-0.10.6.2.1` = `e8cd9e7` / `e8cd9e73aa9967cd0ebb9f2ff3a0d701b9684106`

Reusable commits:
- `a11fa17`
- `dbb42f7`

Main file surfaces:
- `services/meeting-api/meeting_api/outbound_events.py`
- `services/meeting-api/meeting_api/post_meeting.py`
- `services/meeting-api/meeting_api/dispatch_check.py`
- `services/meeting-api/meeting_api/webhook_delivery.py`
- `services/meeting-api/meeting_api/webhook_retry_worker.py`
- `services/meeting-api/meeting_api/webhooks.py`
- `services/meeting-api/tests/test_post_meeting_idempotency.py`

## Explicitly out of scope

- Current uncommitted product changes in the local worktree are excluded from this pack.
- Dead release evidence folders are excluded as source truth.
- Deprecated harness evidence is excluded; reusable checks must live in product tests or dedicated skills.
- The removed realtime transcript regression pack is not replay scope here.

## Blast radius

Billing/usage correctness, post-meeting hooks, webhook retries, dashboard/operator delivery history, replay/suppression operations.

## Data / schema / API / public-contract decisions

Use existing meeting data for event claims; do not change public webhook event names, payloads, or signing headers in this pack.

## Isolation requirements

- Develop from base `v0.10.6^{}` (`6da3abb`), not from current main.
- One branch/worktree for this pack only.
- Evidence root: `.agents/packs/0.10.6x-pack-5-billing-webhook-idempotency/`.
[...]

Issue #330, opened 2026-05-11
[Bug]  Cloud API: 4 billing events per meter type per session (instead of 1)


DavideMichelon10:
## What happened?
                                                                                                                  
  When using the hosted Cloud API (api.cloud.vexa.ai), every bot session generates 4 billing events per meter type
   instead of 1, resulting in being charged 4x the expected amount.

  Example for a 3-minute session (fdc-asqn-hnk):
  - 4× vexa_bot_minutes @ $0.0150 = $0.06
  - 4× vexa_tx_addon_minutes @ $0.0100 = $0.04
  - Total charged: $0.10 — expected: $0.025

  All 4 events have identical minute counts and arrive within milliseconds of each other in two bursts (~3 seconds
   apart).

  ## What did you expect?

  1 billing event per meter type per session:
  - 1× vexa_bot_minutes
  - 1× vexa_tx_addon_minutes

  ## How to reproduce?

  1. Call POST https://api.cloud.vexa.ai/bots with any Google Meet ID
  2. Wait 1–2 minutes
  3. Call DELETE https://api.cloud.vexa.ai/bots/google_meet/{meeting_id}
  4. Download the metering CSV from the dashboard

  The 4x pattern appears consistently on every session, including sessions created directly via curl with no
  application code involved.

  ## Logs / screenshots?

  Date,Meter,Minutes,Cost
  2026-05-11T08:24:22.620Z,vexa_tx_addon_minutes,3.00,$0.0100
  2026-05-11T08:24:22.592Z,vexa_tx_addon_minutes,3.00,$0.0100
  2026-05-11T08:24:22.464Z,vexa_bot_minutes,3.00,$0.0150
  2026-05-11T08:24:22.412Z,vexa_bot_minutes,3.00,$0.0150
  2026-05-11T08:24:22.294Z,vexa_tx_addon_minutes,3.00,$0.0100
  2026-05-11T08:24:22.137Z,vexa_bot_minutes,3.00,$0.0150
  2026-05-11T08:24:19.828Z,vexa_tx_addon_minutes,3.00,$0.0100
  2026-05-11T08:24:19.674Z,vexa_bot_minutes,3.00,$0.0150

  ## Version / env?

  - Hosted Cloud API: api.cloud.vexa.ai
  - No self-hosted components
  - Tested via direct curl (no application code)
  - Plan: pay-as-you-go

DmitriyG228:
[...]

Issue #327, opened 2026-05-09
meeting-api sweeps: webhook retries bump meeting.updated_at via SQLAlchemy autotouch — corrupts every consumer that filters on updated_at


DmitriyG228:
## Symptom

`meetings.updated_at` is bumped on every UPDATE to the row, including writes that record webhook delivery state (`data->'webhook_deliveries'` append, retry counters, etc). A row with active webhook retries can have its \`updated_at\` advanced indefinitely, defeating any consumer that uses \`updated_at < threshold\` as a "stale" / "no-progress" predicate.

## Concrete past failure

#313 — the Pack E.3.2 stale-stopping sweep used \`Meeting.updated_at < threshold\`. Meetings genuinely stuck in 'stopping' kept looking fresh because webhook retry workers updated \`data->'webhook_deliveries'\`, bumping \`updated_at\`, and the sweep never fired. v0.10.6.1 worked around this by switching the sweep to a derived \`last_progress_at\` from \`data->'status_transition'\`.

## Why this is bigger than the sweep

Any other code that uses \`meetings.updated_at\` as a "domain progress" signal is silently wrong. Candidates:
- analytics queries filtering meetings updated since N
- third-party integrations that poll on \`updated_at\`
- any future "stale" sweeps for other status states (joining / awaiting_admission / etc.)
- change-data-capture downstream consumers

The semantics drift from "row last changed" (current behavior) to what most consumers actually want: "domain progress last changed".

## Proposed fix options

A. **Move webhook delivery state to a separate table.** \`webhook_deliveries\` already exists in some shape — verify and migrate.
B. **Mark webhook-retry-touched columns as not bumping \`updated_at\`** at the model/SQL level (PG generated columns or trigger-based).
C. **Add an explicit \`last_domain_progress_at\` column** that is only bumped on status transitions / data shape changes, leave \`updated_at\` as the row-level last-write timestamp.

Option C is least invasive and most explicit; option A is cleanest long-term.

## Urgency

Medium. v0.10.6.1 sweep predicate routes around the symptom for the stale-stopping case, but every other \`updated_at\`-based query in the codebase is a latent bug. Should land in v0.10.7 alongside any analytics work that touches meetings.

## Acceptance

- One of the three options chosen + implemented.
- Pack E.3.2 sweep predicate reverts back to a single-column filter (no JSONB post-filter needed).
- Registry check that catches future regressions: webhook write must not bump the chosen "domain progress" column.

## Source

[...]

PR #369, opened 2026-05-24
[Pack] PACK 5 - Billing And Webhook Idempotency

DmitriyG228:
# Pack PR: PACK 5 - Billing And Webhook Idempotency

Pack epic: https://github.com/Vexa-ai/vexa/issues/360
Pack id: 0.10.6x-pack-5-billing-webhook-idempotency
Release: 0.10.6.x replay
Base branch: v0.10.6^{}
Integration branch: codex/release-0.10.6x-pack-integration

## Outcomes

CEO: Completion events do not over-bill or double-count usage when terminal paths retry or race.

CTO: Meeting API claims billing-sensitive completion events producer-side while keeping public webhook payloads and headers stable.

User: Trustworthy usage/billing events and webhook delivery history without duplicate completion emissions.

## Scope

- #330
- webhook/post-meeting portions of #331/#347

## Validation

- Synthetic: covered by tests/test_post_meeting_idempotency.py
- Compose live meeting: PASS (r3 case compose-qoh-pack5-r3, 22 segs, 14/19 anchors, WS validated, recording completed)
- Lite live meeting: pending (Lite image building)
- Hot human-in-the-loop: pending
- Hardenloop: pending
- Code review: pending
- Docs: no API changes; outbound idempotency is internal

## Recorded caveats

- Pack 5 worktree had host dir mode 700 on services/runtime-api + services/meeting-api/meeting_api. Fixed at bring-up; should be addressed at worktree-create time.
- During pack 5 compose bring-up, runtime-api re-tagged from a stale :dev image caused cross-container HTTP hang; swapped to pack 1's known-good image. Pack 5 doesn't change runtime-api, so this is environmental.
- Pack 5 dashboard re-tagged :dev requires VEXA_ADMIN_API_KEY env; defaults differ across base builds.

### DIFF
--- a/.agents/packs/0.10.6x-pack-5-billing-webhook-idempotency/README.md
+++ b/.agents/packs/0.10.6x-pack-5-billing-webhook-idempotency/README.md
@@ -1,0 +1,99 @@
+# Pack 5 — Billing & Webhook Idempotency (scope clarification)
+
+Pack id: `0.10.6x-pack-5-billing-webhook-idempotency`
+Epic: https://github.com/Vexa-ai/vexa/issues/360
+Pack PR: https://github.com/Vexa-ai/vexa/pull/369
+Release: `0.10.6.x` (stitched into `0.10.6.3`)
+
+This README documents what Pack 5 actually guarantees end-to-end, since the
+pack name ("Billing And Webhook Idempotency") is misleading about scope.
+The shipped code is generic; the word "billing" is a deployment label, not
+a property of the mechanism.
+
+## What Pack 5 ships (sender-side, exactly-once dispatch)
+
+Pack 5 adds a producer-side idempotency ledger for outbound post-meeting
+events fired by `services/meeting-api`. The contract is:
+
+- For every `(meeting_id, channel, event_type, destination)` tuple,
+  meeting-api emits **at most one** delivery attempt per claim.
+- The claim is taken under a row lock (`SELECT … FOR UPDATE`) against the
+  `meeting.data.outbound_events` ledger. A duplicate claim returns
+  `should_deliver=False` and the `attempts` counter is **not** incremented.
+- `event_id` is **stable across all retries** of the same logical event
+  — it is derived deterministically from
+  `channel:event_type:meeting_id:sha256(destination)[:16]`
+  (see `outbound_events.event_key`). Receivers can therefore dedupe on
+  `event_id` across the network retry surface.
+- Stuck `pending` claims older than the configured max-age are swept and
+  re-queued; this preserves the "at-most-one delivery per claim" property
+  without permanently stranding events on a crash between claim and
+  HTTP send.
+
+Verified during stitch validation of `v0.10.6.3`: `claim_outbound_event`
+row-lock idempotency works — duplicate claim returns
+`should_deliver=False`, `attempts` unchanged.
+
+## What Pack 5 does NOT ship (receiver-side dedupe)
+
+End-to-end exactly-once **charging** requires the receiver to dedupe on
+`event_id` as well. Pack 5 only owns the sender-side half of the
+contract:
+
+| Layer                                | Owned by Pack 5? |
+| ------------------------------------ | ---------------- |
+| Claim under row lock                 | yes              |
+| Stable `event_id` across retries     | yes              |
+| Pending-sweep recovery               | yes              |
+| HTTP transport retry / queue         | yes              |
+| Receiver-side dedupe on `event_id`   | **no** (receiver contract) |
+| Receiver's billing ledger commit     | **no** (receiver contract) |
+
+If a downstream service mounted at a `POST_MEETING_HOOKS` URL processes
+the same `event_id` twice (e.g. because it ack'd after a crash without
+persisting the dedupe row), end-to-end exactly-once is broken at the
+receiver, not at Pack 5.
+
+## "Billing" is the deployment label, not the mechanism
+
+The mechanism is a **generic** exactly-once delivery primitive for any
+URL listed in the `POST_MEETING_HOOKS` env var (see
+`services/meeting-api/meeting_api/config.py` and
+`docker-compose.yml`).
+
+- **OSS default** (this repo, `deploy/compose/docker-compose.yml:265`):
+  ```
+  POST_MEETING_HOOKS=${POST_MEETING_HOOKS:-http://agent-api:8100/internal/webhooks/meeting-completed}
+  ```
+  The only configured destination is `agent-api`, used for chat-state
+  propagation. There is **no billing service in OSS**. The exactly-once
+  guarantee still applies, but it benefits chat-state propagation, not
+  billing.
+- **SaaS / production deployments**: operators add their billing-service
+  URL(s) to `POST_MEETING_HOOKS`. They benefit from the same sender-side
+  exactly-once guarantee, provided the receiver implements `event_id`
+  dedupe.
+
+The pack name "Billing And Webhook Idempotency" was chosen for the
+CEO/CTO/user outcome framing on the epic. The shipped code carries no
+payment-processor, plan-tier, or account-credit semantics — it is a
+generic at-most-once-claim primitive plus durable retry plumbing.
+
+## Cross-references
+
+- Files shipped in PR #369 (see `gh pr diff 369`):
+  - `services/meeting-api/meeting_api/outbound_events.py` (ledger + claim)
+  - `services/meeting-api/meeting_api/post_meeting.py` (call site)
+  - `services/meeting-api/meeting_api/webhook_delivery.py` (transport)
+  - `services/meeting-api/meeting_api/webhook_retry_worker.py` (sweep)
+  - `services/meeting-api/meeting_api/webhooks.py` (public webhook
+    payloads unchanged in this pack)
+  - `services/meeting-api/meeting_api/dispatch_check.py` (separate
+    dispatch-time gate; not part of the outbound-event ledger contract,
+    but ships in the same PR)
+  - `services/meeting-api/tests/test_post_meeting_idempotency.py`
+
+- Code-review squash commit (product-only, 6 files, 881 lines):
+  https://github.com/Vexa-ai/vexa/commit/5cb881d3b6fcacbb0a277ce74a7bf4616ddfd04d
+
+- Stitched release candidate: `0.10.6.3`.
--- a/services/meeting-api/meeting_api/dispatch_check.py
+++ b/services/meeting-api/meeting_api/dispatch_check.py
@@ -1,0 +1,207 @@
+"""Generic env-gated dispatch-check call-out.
+
+When ``DISPATCH_CHECK_URL`` is set, bot-create / dispatch entry points call
+out to an external authority to ask "should this action proceed?" The
+authority returns ``allow`` / ``deny`` with an optional reason. When the
+env var is unset (default for OSS self-hosters), the helper is a no-op
+and dispatch proceeds unchanged.
+
+This module is intentionally generic: it carries no payment-processor,
+account-credit, usage-cap, plan-tier, account-identifier, or recurring-charge
+semantics. OSS just asks an opaque authority an opaque question
+and gates on the response. The authority (typically a vexa-platform
+endpoint when env var is set; nothing when unset) decides why.
+
+Failure policy: fail-OPEN on network error, timeout, parse error, or
+5xx response. Rationale: we'd rather risk an over-charge than block
+paying customers on transient gate failure. The authority side is
+expected to log gate-call-misses for reconciliation.
+
+Env vars:
+    DISPATCH_CHECK_URL    Optional. Base URL of the authority. When
+                          unset, all checks succeed unconditionally
+                          (the OSS-self-hoster default).
+    DISPATCH_CHECK_SECRET Optional. Shared secret. When set, the outbound
+                          request carries the HMAC-signed timestamped header
+                          set produced by ``webhook_delivery.build_headers``:
+                          ``Authorization: Bearer <secret>`` +
+                          ``X-Webhook-Timestamp`` +
+                          ``X-Webhook-Signature: sha256=<hmac>`` over
+                          ``<ts>.<body>``. Matches the inbound webhook
+                          verifier on the webapp side and closes the
+                          bearer-replay window.
+    DISPATCH_CHECK_TIMEOUT_S
+                          Optional. Per-call timeout in seconds (float).
+                          Default: ``2.0``.
+
+Behavior matrix:
+
+    | DISPATCH_CHECK_URL | Authority response   | Result            |
+    | ------------------ | -------------------- | ----------------- |
+    | unset              | (not called)         | allow (no-op)     |
+    | set                | 200 ``{"allow":true}``     | allow       |
+    | set                | 200 ``{"allow":false,"reason":"x"}`` | deny |
+    | set                | 5xx                  | allow (fail-open) |
+    | set                | network error/timeout | allow (fail-open)|
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+from dataclasses import dataclass
+from typing import Any, Mapping, Optional
+
+import httpx
+
+from .webhook_delivery import build_headers
+
+logger = logging.getLogger("meeting_api.dispatch_check")
+
+
+def _env_url() -> Optional[str]:
+    """Read the URL each call so tests / hot-reload pick up changes."""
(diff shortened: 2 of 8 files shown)


DmitriyG228:

Single-pass review of commit `237e9cd fix(meeting-api): add outbound event idempotency ledger`. Ranked most-severe first.

| # | Severity | File | Line | Summary |
|---|----------|------|------|---------|
| 1 | HIGH | `services/meeting-api/meeting_api/outbound_events.py` | 196-230 | `find_stale_pending_events` is defined but **never called** — no actual sweep recovery exists for stuck `pending` events. |
| 2 | HIGH | `services/meeting-api/meeting_api/post_meeting.py` | 256-271 | Crash window after successful HTTP delivery but before `mark_outbound_event` — ledger stays `pending`, and if any future sweep is wired up (per finding #1) or any re-trigger of `run_all_tasks` re-claims the `failed`/missing entry, it produces **double delivery** to the billing endpoint. |
| 3 | HIGH | `services/meeting-api/meeting_api/webhooks.py` | 74-84 | `_persist_data_preserving_updated_at` passes `meeting.updated_at` from in-memory state. If a concurrent transaction has already advanced `updated_at` in the DB, this raw UPDATE silently **rewinds** `updated_at` to the older Python-side value, breaking the Pack E.3.2 stale-stopping sweep semantics the comment claims to protect. |
| 4 | MEDIUM | `services/meeting-api/meeting_api/webhook_retry_worker.py` | 92-104 | `_update_meeting_delivery_status` updates `outbound_events[key]` **without a row lock** (no `with_for_update`). Concurrent retry-worker pass + post-meeting reclaim of a `failed` event can race; one writer's `event.update(status)` blindly overwrites the other's claim metadata (e.g. resets `attempts`, blows away a freshly-bumped `claimed_at`). |
| 5 | MEDIUM | `services/meeting-api/meeting_api/outbound_events.py` | 50-56 | `_flag_data_modified` swallows **all** exceptions from `flag_modified` (not just the MagicMock case). In production, a legitimate ORM error (detached instance, expired session) becomes a silent JSONB no-op — claim "commits" without the new event landing on disk, so a duplicate caller sees no ledger entry and re-delivers. |
| 6 | MEDIUM | `services/meeting-api/meeting_api/webhook_retry_worker.py` | 222-227 | Retry worker writes `"attempts": attempt + 1` where `attempt` is the queue-local counter starting at 0. This **clobbers** the ledger's existing `attempts` (already set to 1 by `mark_outbound_event` for the initial direct attempt). Audit count of total delivery tries becomes wrong; an entry that took 1 direct + 3 retry attempts shows `attempts=3` not `4`. |
| 7 | LOW | `services/meeting-api/meeting_api/outbound_events.py` | 33-38 | `destination_hash` truncates SHA256 to 16 hex chars (64 bits). For the small internal-hook URL set this is fine today, but a future operator adding many auto-generated hook URLs could (theoretically, birthday-bound ~2^32) collide and have one event mask another. |
| 8 | LOW | `services/meeting-api/meeting_api/outbound_events.py` | 156-178 | `mark_outbound_event` constructs a stub `{"key": key}` when the ledger entry is missing, then writes `status` (e.g. `"delivered"`) **without** `channel`/`event_type`/`destination`/`payload`. If the underlying claim row never landed (see #5) the ledger ends up with a content-free `delivered` marker that is invisible to recovery and to operators reading `meeting.data.outbound_events`. |

### Failure scenarios (concrete)

1. **No stale recovery**: producer crashes between `claim_outbound_event` commit and `deliver_with_result` return. Ledger entry sits at `status=pending` forever. `find_stale_pending_events` was written for exactly this case but no caller invokes it. Billing event is silently lost.
2. **Double billing**: caller A successfully POSTs to billing, OS-kills the worker before `mark_outbound_event` runs → ledger stuck `pending`. A second `run_all_tasks` schedule (e.g. from sweep re-trigger) sees `pending` and short-circuits — OK today. **But** if a stale-pending recovery is later added (per finding #1) and naively reclaims by flipping to `failed`/missing, the next caller re-delivers — billing event posted twice for the same meeting.
3. **`updated_at` rewind**: status webhook arrives concurrently with a domain mutation. Domain commits `updated_at = T2`. Status-webhook helper still has `meeting.updated_at = T1` in memory; raw UPDATE writes `updated_at = T1`. Stale-stopping sweep now considers the meeting older than it is and may stop a still-active bot.
4. **Retry-worker race**: failed delivery → ledger `queued`. Retry worker picks it up. Concurrently, post-meeting hooks fires for a re-scheduled run and (after retry worker marks `failed` because of exhausted attempts) reclaims and starts a new attempt. Retry worker's late writer wins → just-claimed `pending` row gets stomped back to `failed`. New attempt's bookkeeping is lost.
5. **Silent flag_modified swallow**: SQLAlchemy raises `InvalidRequestError` on a detached/expired meeting instance; `_flag_data_modified` eats it. `await db.commit()` then commits a row where JSONB wasn't marked dirty. Ledger never persisted; next call re-claims and **re-delivers**.
6. **Wrong attempts audit**: ledger reports `attempts=3` for an event that actually called the billing endpoint 4 times (1 direct + 3 retries). Operator reading the ledger thinks one retry slot is still available; reconciliation count is wrong.
7. **Hash collision** (theoretical): two POST_MEETING_HOOKS URLs share a 64-bit prefix → one ledger key — events for hook B mask hook A; A is never delivered.
8. **Content-free marker**: per #5, the claim write never persists; retry worker eventually fires the deliver and writes `mark_outbound_event(status="delivered")` against the missing key → ledger now contains `{"key": "...", "status": "delivered", "updated_at": "..."}` with no `payload`/`destination`/`channel`. Audit and stale-sweep code that filters on `status == "pending"` won't notice, but anyone trying to replay or reconcile cannot.

---

DmitriyG228:

Pack 5 review: outbound event idempotency ledger (`outbound_events.py`, `dispatch_check.py`, `post_meeting.py`, `webhook_delivery.py`, `webhook_retry_worker.py`, `webhooks.py`). Findings ranked most-severe first.

---

### 1. (HIGH) Stale-pending sweep is dead code — crash-after-claim events stuck in `pending` forever

`outbound_events.find_stale_pending_events` is defined but **no caller exists anywhere in the tree** (`grep -rn find_stale_pending services/` returns only the definition). The docstring on `fire_post_meeting_hooks` explicitly relies on it ("the retry sweep can recover the narrow crash-after-claim window").

Failure scenario: meeting-api crashes (OOM kill, K8s pod evict, SIGTERM during graceful shutdown) in the ~10-30s window between the `claim_outbound_event` commit and the `mark_outbound_event(status=queued|delivered)` commit. The ledger row is now permanently `status=pending`. On restart, `fire_post_meeting_hooks` re-fires for the same meeting and `claim_outbound_event` returns `should_deliver=False` because `pending` is in the skip set. The billing/usage hook for that meeting **silently never fires**.

Also: even if the sweep were wired, `find_stale_pending_events` scans only the **100 most recent COMPLETED/FAILED meetings** (`order_by(Meeting.id.desc()).limit(limit)`). On a busy deployment that completes >100 meetings during the outage window, older stuck rows fall off the recovery horizon forever.

Fix: wire it into the periodic sweep that already runs alongside the webhook retry worker, OR widen the predicate to `status='pending' AND claimed_at < now - interval`. The current "bounded scan" comment is misleading — it's not a sweep, it's an unreachable function.

---

### 2. (HIGH) POST_MEETING_HOOKS deliveries are **unsigned** — billing payloads ship in cleartext without HMAC

`post_meeting.py` line ~245 calls:

```python
result = await deliver_with_result(
    url=hook_url,
    payload=payload,
    timeout=10.0,
    label=f"post-meeting-hook meeting={meeting.id}",
    metadata={"meeting_id": meeting.id, "outbound_event_key": key},
)
```

`webhook_secret` is omitted (defaults to `None`), so `build_headers(None, ...)` produces no `Authorization: Bearer` and no `X-Webhook-Signature`. `dispatch_check.py` carefully signs its outbound POSTs ("closes the bearer-replay window") — but the actual billing/usage hooks fired on meeting completion do not.

Failure scenario: an attacker on the path between meeting-api and the billing endpoint (or a misconfigured `POST_MEETING_HOOKS` env pointing to a host hijacked via DNS/typo) can both **observe** meeting metadata (user_id, duration, transcription flag, name) and **forge** completion events to inflate or zero a victim's usage ledger. There is no signature verification possible on the receiving side because nothing is signed.

There is no `POST_MEETING_HOOKS_SECRET` env var anywhere in the codebase. Either add one and pipe it through, or document explicitly that POST_MEETING_HOOKS is in-cluster-only and never crosses a trust boundary.

---

### 3. (HIGH) Race: retry-worker `_update_meeting_delivery_status` does read-modify-write on `meeting.data` without `FOR UPDATE` — concurrent ledger entries clobber each other

`webhook_retry_worker.py:_update_meeting_delivery_status` uses `await session.get(Meeting, meeting_id)` — a plain MVCC snapshot read, **no row lock**. It then mutates `events[key]` and writes the entire `data` JSONB back.

Meanwhile `claim_outbound_event` and `mark_outbound_event` in `outbound_events.py` correctly use `with_for_update()`.

Failure scenario: a meeting has two hook URLs in POST_MEETING_HOOKS (e.g., billing + analytics). Both fail and get enqueued. The retry-worker processes them in quick succession. While worker is updating `outbound_events[key_A]` (read snapshot S, mutate, write), `claim_outbound_event` for a *different* event_type on the same meeting holds FOR UPDATE on the same row. Whichever commits last wins; the entire JSONB blob gets overwritten with the loser's view. The other entry's update is lost.

Also affects the `events[key_A]` vs `events[key_B]` retry-worker double-update if the worker processes both queued entries inside one poll cycle in parallel (they each open their own session via `_session_factory`).

Fix: in `_update_meeting_delivery_status`, use `select(Meeting).where(Meeting.id == ...).with_for_update()` instead of `session.get`. Pattern is already in `outbound_events.py`; copy it.

---

### 4. (HIGH) Ordering race: `mark_outbound_event(status="queued")` can overwrite the retry-worker's `status="delivered"` and revert the ledger to a lie

In `fire_post_meeting_hooks`:

```python
result = await deliver_with_result(...)        # ← can take 30s; on failure, enqueues to Redis
await mark_outbound_event(db, meeting_id=..., status=result.status, ...)  # ← writes "queued"
```

The retry-worker polls every 30s. The Redis entry's `next_retry_at = now + 60`, so normally worker won't fire until 60s later — safe.

But under load (worker behind, GC pause in caller, scheduler hiccup), the sequence can interleave:

1. `deliver_with_result` enqueues to Redis, returns `status="queued"` at T+0.
2. **The caller's coroutine is preempted** before `mark_outbound_event` runs (or `db.commit()` stalls on a busy connection pool).
3. The retry-worker wakes 60s later, pops the entry, delivers successfully, calls `_update_meeting_delivery_status` with `{"status": "delivered", "delivered_at": ...}` (this assumes finding #3 is fixed). Ledger now says `delivered`.
4. The caller's preempted coroutine resumes and calls `mark_outbound_event(status="queued")`, which **unconditionally overwrites** `status` and clears `delivered_at` (since `status="queued"` is in the no-error-pop branch).

The ledger now claims the event is `queued` even though downstream got and acked it. No one re-delivers (status is not in stale-pending sweep set), so it's just a permanent lie — but if humans use the ledger to reconcile billing, they will double-charge or under-charge.

Fix: `mark_outbound_event` should compare-and-set, not blindly update. E.g., refuse to downgrade `delivered` to `queued`. Or: only mark `queued` from inside `deliver_with_result` before returning, so the lock-holding caller and the redis push are within one critical section.

---

### 5. (MEDIUM) `event_key` collides across distinct hook URLs that hash-prefix-collide (16 hex chars = 64 bits truncated SHA-256)

```python
def destination_hash(destination: str) -> str:
    return hashlib.sha256(destination.encode("utf-8")).hexdigest()[:16]
```

64 bits is fine for accidental collision (~birthday at 2^32 ≈ 4B distinct URLs per meeting — not realistic for one meeting), but it's chosen-prefix vulnerable. If an attacker controls one POST_MEETING_HOOKS URL (e.g., a malicious internal subdomain in a multi-tenant deployment) and can grind 2^64 candidates against another URL on the same meeting, they could cause `event_key` collision and trick `claim_outbound_event` into thinking the second hook is already delivered → second hook never fires.

This is unlikely in OSS self-host, but the truncation is gratuitous. Use full hexdigest or just `urlsafe_b64encode(sha256(url).digest())`. Same write cost on JSONB keys.

More importantly: `event_key` does NOT include `event_id` (the envelope's `event_id` is set to the key itself), so if the same channel/event_type/meeting/dest is ever fired with a different payload (e.g., schema change ships, replay of completion), it will be silently skipped as "already delivered" even though the payload changed. The ledger has no payload-hash check.

---

### 6. (MEDIUM) `is_stale_pending_event` returns `True` when `claimed_at` is missing/malformed — sweep would reclaim a freshly-claimed event if the timestamp didn't serialize correctly

```python
claimed_at = _parse_iso(event.get("claimed_at") or event.get("updated_at"))
if claimed_at is None:
    return True
```

Belt-and-suspenders for missing data, but if a future refactor renames the field or a partial commit writes `status=pending` without `claimed_at`, every sweep pass will reclaim the event, restarting delivery. Combined with finding #4's "claim resets attempts to existing+1 only if `should_deliver=True`" — but `should_deliver` returns False for `pending`, so this only fires if status accidentally drops out of `{delivered, queued, pending}`.

Lower severity, but the fail-open default is the wrong direction for an idempotency ledger. Default should be: missing timestamp → log error, leave alone (assume something racy is in flight), let an operator look at it.

---

### 7. (MEDIUM) `dispatch_check` failure-open + 4xx is treated as "deny" (allow=false), but spec says fail-OPEN — 4xx from authority blocks paying customers

`dispatch_check.py` fails open on 5xx, network error, and JSON parse error. But a 4xx response (e.g., 401 if the authority is misconfigured with a wrong secret, 403 if rate-limited, 404 if the `/check` route was renamed) falls through to:

```python
allow = bool(body.get("allow", True))
```

A 4xx with empty body → `body.get("allow", True)` → `allow=True`. Good. But a 4xx with `{"detail": "..."}` (FastAPI default error shape) → `allow = bool(body.get("allow", True)) = True`. Still fine. So actually this is OK by accident — but it's not what a reader of the docstring would expect. The docstring says "fail-open on 5xx" implying 4xx is a deny path; the implementation actually fails open on 4xx too via the default. Either fix the docstring or explicitly check `if 400 <= status < 500: fail-open` to make the intent legible.

A real bug here: a *malicious or buggy* authority that returns 200 with `{"allow": false, "reason": "..."}` will deny — including for transient auth errors at the authority side (since the authority is supposed to be the source of truth). There is no circuit-breaker. Sustained authority badness will block all paying customers indefinitely. Add a deny-rate sanity check or "if last N decisions were all-deny, fail-open."

---

### 8. (LOW) `webhooks._persist_data_preserving_updated_at` issues `update(Meeting)` while the session likely already has a dirty `meeting` instance in identity map — the in-memory `meeting.data` mutation can be flushed by a later `await db.commit()` and re-bump `updated_at`

```python
data = dict(meeting.data) if meeting.data else {}
data["webhook_delivery"] = status
meeting.data = data                                      # ← marks instance dirty
await _persist_data_preserving_updated_at(db, meeting, data)  # ← raw UPDATE preserves updated_at
```

The comment says "we still mutate the in-memory `meeting.data` so the caller's subsequent reads see the new state, but we do NOT call flag_modified". For top-level JSONB assignment SQLAlchemy *does* detect the change without flag_modified (it's an attribute set, not a mutation-in-place), so the instance is dirty. If the calling code later does `await db.commit()` (e.g., post_meeting flow does many commits), SQLAlchemy will flush the dirty instance and re-issue an UPDATE that *does* hit the `onupdate=func.now()` default — defeating #327's whole point.

Fix: after the raw UPDATE, call `db.expire(meeting, ["data"])` or `db.refresh(meeting)` to clear the dirty flag, OR mutate `data` in place and never reassign `meeting.data`. The current shape is fragile — works only as long as no caller commits the session afterward, which is not a property the code enforces.

---

### Summary of severities

- HIGH: 4 (stale-sweep unwired, unsigned hooks, worker missing FOR UPDATE, mark/worker ordering race)
- MEDIUM: 3 (key collision + payload-blind dedup, sweep fail-open default, dispatch-check no circuit breaker)
- LOW: 1 (updated_at preservation undermined by dirty instance)

---

DmitriyG228:
Adopted into the **billing lane** (pack #401 — crash-proof completion hooks). This PR is the only survivor of the 0.10.6x PACK 1–7 stitching series: it's mergeable, non-draft, and directly overlaps the lane's scope (#397 duplicate hook delivery, platform-side meter idempotency). Next step: re-review against current main and fold into the #401 delivery.

DmitriyG228:

PR #381, opened 2026-05-27
fix(googlemeet): support 'Join now' button for OPEN meetings (accessType=OPEN)

### LINKED ISSUE #380: [Bug] Google Meet bot cannot join specific meeting links (self-hosted Vexa)
I’m currently self-hosting Vexa and testing the Google Meet transcription/join flow.

The bot works correctly for:

Google Meet rooms that I create myself
Normal test meetings in my environment

However, there is an issue with a specific Google Meet link sent by a customer:

The bot fails to join the meeting
Transcription immediately fails
The UI shows:

Transcription failed
Bot exited with code 1; reason: self_initiated_leave
Error: requested

I attached a screenshot of the error UI.

Image

helbertparanhos:
## Problem

When joining a Google Meet created via the **Google Meet REST API** with `accessType: OPEN`, the anonymous-flow bot fails with `self_initiated_leave` (exit code 1) and shows **Transcription failed** in the UI.

**Root cause:** The anonymous flow used only `googleJoinButtonSelectors[0]` — an XPath selector for `'Ask to join'`. OPEN meetings render a **'Join now'** button instead (also labeled _'Participar agora'_ in pt-BR). The bot waited 60 s for a button that never appears, then timed out and left.

The name input field (`'Your name'`) **is** present on OPEN meetings, so the bot fills it correctly — the only failure is in the subsequent join-button click.

## Reproduction

1. Create a meeting space via the Google Meet API with `{ "config": { "accessType": "OPEN" } }`
2. Send the bot to the resulting `meetingUri`
3. Bot reaches **Joining** state then immediately transitions to **Failed** with reason `self_initiated_leave`

Bot screenshot at the moment of failure shows the correct join screen with a **'Join now'** button (greyed out until name is filled), not an **'Ask to join'** button.

## Fix

Replace the single-selector wait in the anonymous flow with a sequential loop over all `googleJoinButtonSelectors`. The existing selector order is preserved — `'Ask to join'` is still tried first (8 s timeout), so regular and lobby meetings are unaffected. `'Join now'` (index 2) is reached only when the earlier selectors time out.

```typescript
// Before
const joinSelector = googleJoinButtonSelectors[0];
await page.waitForSelector(joinSelector, { timeout: 60000 });
await page.click(joinSelector);

// After
let joinClicked = false;
for (const sel of googleJoinButtonSelectors) {
  try {
    await page.waitForSelector(sel, { timeout: 8000 });
    await page.click(sel);
    joinClicked = true;
    break;
  } catch (e) {
    log(`Join button not found with selector: ${sel}, trying next...`);
  }
}
if (!joinClicked) throw new Error('No join button found after trying all selectors');
```

## Testing

Verified on self-hosted `vexa-lite` (patch applied to compiled JS in container):
- OPEN meeting (`accessType: OPEN`) → bot joins successfully, transcription starts ✅
- Regular meeting (waiting room / `'Ask to join'`) → unaffected, still works ✅

Fixes #380
Related #343

### DIFF
--- a/services/vexa-bot/core/src/platforms/googlemeet/join.ts
+++ b/services/vexa-bot/core/src/platforms/googlemeet/join.ts
@@ -110,7 +110,7 @@ export async function joinGoogleMeeting(
     await page.screenshot({ path: '/app/storage/screenshots/bot-checkpoint-0-after-join-now.png', fullPage: true });
     log("📸 Screenshot taken: After join click (authenticated)");
   } else {
-    // Anonymous flow: enter bot name and ask to join
+    // Anonymous flow: enter bot name and join
     log("Attempting to find name input field...");
 
     const nameFieldSelector = googleNameInputSelectors[0];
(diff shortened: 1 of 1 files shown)


DmitriyG228:
Closing with credit — you found a real hard-fail: OPEN-access meetings show 'Join now' instead of 'Ask to join' and the bot timed out. The join-selector rework that later landed on release/0.10.7 made the button matching locale-agnostic (matching 'Ask to join'/'Join now'/localized equivalents alike), which covers this case with the same intent as your loop, so merging the diff now would regress that. Your diagnosis stands credited here, and 0.12 handles OPEN meetings too (test-pinned). Thanks @helbertparanhos.

PR #452, opened 2026-06-20
feat: add `discord` as a non-bot ingest platform

rennf93:
Adds Discord as a first-class platform whose transcripts are written directly by an external adapter (no Vexa bot), the same non-bot treatment as browser_session.

meeting-api:
- Platform enum: add DISCORD = "discord" (unblocks the api-gateway path-param validation and the transcript collector, both of which type on Platform).
- construct_meeting_url: explicit DISCORD branch (channel-ID native_id, no URL).
- concurrency limit: exclude discord from the active-bot count (like browser_session) so external ingest doesn't consume max_concurrent_bots.

dashboard:
- Platform union + PLATFORM_CONFIG: add a discord entry (name/colors/pattern).
- meeting list + detail: render a Discord icon instead of falling back to Zoom.
- add public/icons/discord.svg (brand glyph, Discord blurple).


DmitriyG228:
Thanks for this @rennf93 — the platform plumbing (enum, concurrency exclusion, dashboard config) is clean and mirrors the `browser_session` non-bot pattern well.

Before we merge, we want to understand the **end-to-end** flow, because the piece that actually makes Discord work isn't in this PR and we can't find it in the repo. Our reading is that this follows the **"transcripts written directly by an external adapter"** model — i.e. finished segments are pushed into the collector's Redis stream, and Vexa itself does no capture/transcription for Discord. A few questions to confirm:

1. **What is the component that connects to Vexa?** Is there an external Discord adapter/bot (in a separate repo, or planned)? If it exists, a link would help us validate the whole path.
2. **Who runs STT?** Discord's voice gateway only exposes raw Opus audio — it has no native captions/transcription API. So something has to transcribe. Does the adapter run its **own** STT and deliver text segments, or does it feed Discord audio into Vexa's transcription pipeline (whisperlive)?
3. **How does a Discord meeting get created, and how are segments published?** Which entrypoint creates the `platform="discord"` meeting, and what writes segments — directly onto `REDIS_STREAM_NAME` in the same shape the transcription service emits, or some other path? (Related: `construct_meeting_url` returns `None` for Discord, unlike `browser_session://{id}` — was that deliberate?)
4. **Can you point us at a working end-to-end run** (adapter → segments visible via `GET /transcripts/discord/{id}`)? We'd like to ship 0.10 with Discord actually functional, not just the platform label.

Happy to help wire it up if the adapter is open — just want the full picture before adding a platform to the UI.

rennf93:
Hi @DmitriyG228 , thanks for approving the PR!

I'm not at my computer at the moment, but I opened issue [463](https://github.com/Vexa-ai/vexa/issues/463) that explains a bit how I'm doing it.

This might help as a direct link too: https://github.com/rennf93/discord-vexa-bridge

As soon as I get to my PC, I'll reply properly to every one of your questions.

Thanks again!

rennf93:
Hey @DmitriyG228 ,

Thanks for the merge, and for the thorough questions. Replying properly now as promised.

The short version: the piece that makes Discord actually work is an external adapter I've built and open-sourced at **https://github.com/rennf93/discord-vexa-bridge**. It's a self-hosted Discord bot (bring-your-own `DISCORD_TOKEN`) that joins a voice channel, decrypts the audio, transcribes it, and writes the rows into Vexa's Postgres — so a Discord call shows up on the dashboard / API / MCP like any other meeting. It needs no Vexa bot and no Vexa API key.

### 1. What is the component that connects to Vexa?

A standalone bot, separate repo, exists today (MIT): rennf93/discord-vexa-bridge. It runs as its own service on the same Docker network as Vexa's `postgres` and `transcription-worker`, and talks to exactly two Vexa components:

  - Vexa's **Postgres** (read/write — creates the meeting, writes transcripts)
  - Vexa's **transcription-worker** (POSTs WAV for STT)

It does **not** call `meeting-api`, `api-gateway`, `admin-api`, or Redis. That's a deliberate choice (single-tenant, self-hosted, minimal surface) but it has one architectural implication I want to be upfront about — see the note under Q3.

### 2. Who runs STT?

**Vexa does.** The adapter does not run its own STT. Per utterance it downmixes the decrypted Opus to 16 kHz mono WAV and POSTs it to Vexa's existing transcription-worker at the same endpoint Vexa's own runtime uses (`http://transcription-worker:8000/v1/audio/transcriptions`, OpenAI-compatible, faster-whisper). It reads `{"text": ...}` back and writes that to Postgres. So Vexa's STT pipeline stays the single source of transcription — Discord just feeds it audio instead of a bot capturing a Meet/Zoom tab.

### 3. How is a Discord meeting created, and how are segments published?

Today the adapter writes Postgres directly (no Redis, no collector):

  - On `/join`: `INSERT INTO meetings (user_id, platform, platform_specific_id, status, start_time, data, created_at, updated_at) VALUES (..., 'discord', <channel_id>, 'active', now(), '{}'::jsonb, now(), now()) RETURNING id`, then `INSERT INTO meeting_sessions (meeting_id, session_uid, session_start_time) VALUES (..., <uuid>, now())`.
  - Per utterance: `INSERT INTO transcriptions (meeting_id, start_time, end_time, text, speaker, language, session_uid, segment_id, created_at) VALUES (...)`. `start_time`/`end_time` are seconds-relative-to-session-start (same convention as Meet/Zoom), `speaker` is the Discord display name, `segment_id` a fresh UUID.
  - On `/leave`: `UPDATE meetings SET status='completed', end_time=now(), ...`.

`native_id` is the Discord **channel ID** (string). `construct_meeting_url` returning `None` for Discord is deliberate and correct — there is no joinable Vexa URL to construct; the meeting is ingested externally, same shape as `browser_session://{id}` but with no URL at all.

**The honest caveat, since you raised the collector:** this means the adapter currently **bypasses Vexa's `transcription_segments` Redis stream and the collector**. That works today because the gateway reads transcripts from Postgres and the stream is an ingestion buffer, not the canonical store. But I understand your intended external-adapter model is "publish onto the stream, let the collector write Postgres" — which would also give Discord the collector's `session_start`/`session_end` handling, speaker-event caching, segment dedup, webhook delivery, and the `MeetingToken` auth model for free.

I'm happy to **move the adapter onto the collector path** instead of direct Postgres — i.e. emit `session_start` / `transcript` / `session_end` messages onto `transcription_segments` in the shape your `processors.py` expects (`type`, `uid`, `segments[]` with `start/end/text/segment_id/completed`, `platform`, `native_meeting_id`, and a `MeetingToken` JWT signed with `ADMIN_TOKEN`, `aud=transcription-collector`, `scope=transcribe:write`). To do that cleanly I'd want to mirror exactly how `browser_session` obtains its `meeting_id` + token today — could you point me at that flow so Discord copies it rather than inventing one? If you'd rather keep the direct-Postgres path, I'm also glad to harden that — just say which you prefer for 0.10.

### 4. Can I point you at a working end-to-end run?

Yes. Runbook (full version in the repo's README):

  1. Create a Discord bot (Developer Portal, Server Members Intent, invite wit `bot` + `applications.commands`, View Channels + Connect).
  2. Deploy the adapter as a compose service on Vexa's network, with `DISCORD_TOKEN`, `DATABASE_URL`, `TRANSCRIBE_URL`, `VEXA_USER_ID` (the id from your `create-user` call). The repo ships a `compose-snippet.yml` plus a full `docker-compose.yaml` that brings up the whole Vexa stack + the adapter together.
  3. In Discord, join a voice channel and run `/join`. Speak. Run `/leave`.
  4. `GET /transcripts/discord/{meeting_id}` returns the rows; the dashboard renders the meeting with the Discord icon from #452.

I'm glad to run a live demo call (or send a recording + the resulting transcript rows / API response) so you can see it end-to-end before 0.10.

### Why the adapter is non-trivial (the bit that isn't in #452)

Since March 2026 Discord enforces E2EE on all voice via the **DAVE protocol** (MLS + per-sender frame crypto). Off-the-shelf Discord libraries can't decrypt it, so the adapter implements the DAVE **receive** path itself: voice-gateway v8 handshake, MLS group join, rtpsize transport decrypt (AES-256-GCM / XChaCha20), and per-SSRC DAVE frame decrypt + Opus decode. The MLS/frame-crypto core is delegated to Discord's own `libdave` via the `dave.py` PyPI binding; the rest is protocol glue. That's the part that makes a from-scratch Vexa reimplementation expensive — and the reason an external adapter is the natural shape here.

Happy to wire it up however you prefer for 0.10. If the adapter is useful as the canonical Discord ingest path, I'd also be glad if Vexa's docs/README linked to it so people can find it — and I'll keep maintaining it against DAVE bumps.

Issue #439, opened 2026-06-11
[Pack] Extract joining layer into a zero-infra, community-hackable package (packages/meet-join)


DmitriyG228:
## Lane: bot (QUEUED — `status:available`, not in-progress)

WIP gate: the bot lane is currently #407 (Restore GMeet Join). This pack is the **next** bot-lane pack; it bakes in the lessons from #407 rather than competing with it.

## Goal

Carve the **joining layer** out of `vexa-bot/core` into a self-contained, community-hackable package (`packages/meet-join/`) with a zero-infra standalone runner, so selector-rot and admission/lifecycle bugs become reproducible by a contributor with **no Vexa backend** — just Node + a free meeting link.

## Empirically validated (2026-06-11 spike)

A ~120-line `standalone-join.ts` imported `join.ts` + `admission.ts` directly, launched Chromium, and walked the full GMeet join flow against a live meeting on a MacBook — **no Docker, Redis, meeting-api, or transcription**. Confirmed:
- ✅ join/admission modules import cleanly without `index.ts`
- ✅ status callbacks are already null-tolerant (`No callback URL configured` → proceeds)
- ✅ the layer reaches "Ask to join" → lobby → admission-poll with only the two source files
- ✅ reproduced #432 (page goes black + reCAPTCHA ~0.5s after click) on residential IP — the standalone runner IS a #432 repro harness

## The seam

`platforms/shared/meetingFlow.ts` already defines `PlatformStrategies` (`join`, `waitForAdmission`, `checkAdmissionSilent`, `prepare`, `startRemovalMonitor`, `leave`). Extraction = inverting the back-imports from `meetingFlow.ts` → `index.ts` (`hasStopSignalReceived`, `callStartupCallback`, camera/chat triggers) into an event/callback contract. The package emits state events + hands back the `Page`; the host does media. **Not a rewrite** — a dependency inversion.

## Packaging fixes discovered in the spike (each a `good-first-issue`)

1. **Screenshot paths hardcoded to `/app/storage/screenshots`** → `ENOENT` off-Docker. Thread `screenshotDir` through config, default cwd.
2. **`x11vnc` spawns unconditionally** (escalation path) → crashes on non-Linux. Guard VNC behind a platform/env check.
3. **`joinGoogleMeeting(page, url, name, config)` is 4 positional args** while the sibling strategy interface is `(page, config)`. Normalize to a config object.
[...]

PR #454, opened 2026-06-21
fix(gmeet): treat Gemini "take notes" consent prompt as a pre-admission gate

### LINKED ISSUE #429: GMeet: auto-dismiss Google's Gemini 'take notes' in-call banner (suspected transcription interference)
## Context
Observed on prod meeting **14560** (Google Meet, status `active`, 0 transcription rows). The Google Meet UI showed Google's **Gemini "take notes for me" banner** (the in-call prompt asking whether to let Gemini take notes/transcribe).

## Symptom
- Bot was `active` (in the call) but produced **0 transcription segments**.
- A Gemini notes **banner** was present in the meeting UI.

## Hypothesis (unconfirmed)
A banner (vs a modal) normally shouldn't block audio capture, so the empty transcript *may* be incidental (meeting just started / no speech yet). But Google's Gemini/notes prompts are a growing class of in-call UI that can:
- overlap controls or capture focus, and/or
- signal Google-side note-taking that competes with the bot.

Needs confirmation whether the banner actually correlates with missing audio capture.

## Proposed
1. **Verify**: reproduce with the Gemini-notes banner present and check whether the bot finds audio tracks / sends to whisper (rule in/out interference).
2. **Harden**: add the Gemini/notes banner's dismiss/close control to the bot's **post-admission dialog-dismissal** step (`services/vexa-bot/core/src/platforms/googlemeet/`), alongside existing pop-up dismissals, so such prompts are cleared on join and can't interfere.

## Evidence
- Prod meeting 14560 (active, 0 transcriptions) — Gemini notes banner observed by operator.

Labels: bot, google-meet, robustness

thatditsyboy:
Closes #429.

## Problem
Per the issue's **corrected** problem statement: Google's Gemini "take notes for me" prompt is a **consent gate**, not cosmetic chrome. The bot false-detects "admitted" (it sees meeting-like controls) and reports **ACTIVE** while still behind the consent prompt → *status active, 0 transcriptions* (prod meeting 14560).

## Fix
Detect "consent pending" as a distinct **pre-admission** state and route to `needs_human_help` instead of ACTIVE — **without auto-consenting** (consent is a human decision; the issue explicitly says do not auto-click it).

- **`selectors.ts`** — `googleConsentPromptIndicators`, targeted at the prompt's copy (`take notes for me`) so it does **not** match the always-present Gemini toolbar button.
- **`admission.ts`** — `hasConsentPrompt()` detector + a **negative guard** in `checkForGoogleAdmissionIndicators` (mirrors the existing waiting-room guard) so a pending consent prompt suppresses the "admitted" signal. When detected, escalate to `needs_human_help` (`consent_required`) and keep polling so admission proceeds once a human grants consent — mirroring the existing reCAPTCHA "stay for human solve" handling.
- **`admission.test.ts`** — extends the existing source-shape suite to pin the consent-gate wiring.

This is the same false-positive class as the admission settle-fix, specialized to consent-pending — and it reuses the existing escalation + reCAPTCHA patterns rather than introducing new machinery.

## Verification
`npx tsx src/platforms/googlemeet/admission.test.ts` → **11 passed, 0 failed** (5 existing + 6 new), following the repo's source-shape test convention for admission ("a full Playwright DOM test would require brittle UI fixtures").

## Caveats (please confirm on your repro)
- **Selectors need live-DOM confirmation.** I can't reproduce the Gemini banner here, so the `googleConsentPromptIndicators` are best-effort, based on the prompt copy + the reCAPTCHA precedent. Please verify against the live prompt via the #439 standalone runner (`npx tsx standalone-join.ts <meet-url>`) and tweak the copy/selectors if Google's wording differs. They're deliberately specific (the prompt's "take notes for me" text) to avoid matching the Gemini toolbar button, which would otherwise break *all* joins.
- I couldn't run the full `tsc` build locally (heavy native deps — `onnxruntime-node`, playwright browsers). The changes are type-trivial mirrors of existing typed functions in this file.

### DIFF
--- a/services/vexa-bot/core/src/platforms/googlemeet/admission.test.ts
+++ b/services/vexa-bot/core/src/platforms/googlemeet/admission.test.ts
@@ -113,5 +113,44 @@ expectFileContains(
   'hasExplicitDenyText',
 );
 
+console.log('\n=== Google Meet Gemini consent-gate handling (#429) ===');
+
+expectFileContains(
+  'selectors define a consent-prompt indicator list',
+  SELECTORS_TS,
+  'googleConsentPromptIndicators',
+);
+
+expectFileContains(
+  'consent selectors target the take-notes prompt copy',
+  SELECTORS_TS,
+  'take notes for me',
+);
+
+expectFileContains(
+  'admission.ts has a consent-prompt detector',
+  ADMISSION_TS,
+  'export async function hasConsentPrompt',
+);
+
+expectOrder(
+  'admission suppresses ACTIVE when a consent prompt is present',
+  admissionBody,
+  'const consentPending = await hasConsentPrompt(page)',
+  'return false',
+);
+
+expectFileContains(
+  'consent gate escalates to needs_human_help (consent_required)',
+  ADMISSION_TS,
+  'triggerEscalation(botConfig, "consent_required")',
+);
+
+expectFileContains(
+  'consent is escalated, not auto-clicked (human decision)',
+  ADMISSION_TS,
+  'not auto-consenting',
+);
+
 console.log(`\n=== googlemeet admission summary: ${passed} passed, ${failed} failed ===`);
 process.exit(failed > 0 ? 1 : 0);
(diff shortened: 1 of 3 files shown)


thatditsyboy:
Rebased onto the latest `release/0.10.7` to clear the merge conflict. The conflict was in `admission.test.ts`, where the base picked up the ESM-safe `__dirname` fix (#455) right where this PR appends its consent-gate checks. Kept both blocks, so the rejection-outcome assertions and the consent-gate assertions now sit side by side. Reran the structural suite on the rebased head: 14 passed, 0 failed. PR is green again whenever you want to take a look.

PR #464, opened 2026-06-27
fix(meeting-api): stop unfinalized-recordings sweep from wedging the event loop

DmitriyG228:
## Problem (production CrashLoopBackOff → dashboard /meetings outage)

`_sweep_unfinalized_recordings` and `recover_recordings_jsonb_from_storage`
(`meeting_api/sweeps.py`) listed the **whole-user** S3 prefix
`recordings/{user_id}/` (10k+ objects for heavy users) via a **synchronous**
boto3 call **directly on the async event loop**, every 60s.

On a heavy user (>10k objects) the list froze the loop for ~1.5s, so `/health`
and `/readyz` timed out → kubelet liveness probe killed the pod →
CrashLoopBackOff → the meeting-api Service had zero endpoints → the dashboard
`/meetings` page intermittently failed to load. `main.py` even documents the
trap: `/health` is *"always 200 unless event loop is wedged"*.

Three compounding bugs, all fixed here:

1. **Blocking I/O on the loop.** Both list calls (and the inline finalizer
   recovery path) now run via `asyncio.to_thread`, so the event loop — and its
   probes — never stalls.

2. **User-scoped prefix.** The real key layout (write side `recordings.py`,
   parser `_parse_recording_chunk_key`) is
   `recordings/<user>/<rec>/<session>/<media>/<seq>`. The `rec` id sits
   *between* user and session, so a session can't be prefixed directly. New
   `StorageClient.list_common_prefixes()` (delimiter listing) enumerates
   per-recording prefixes cheaply, then `_list_session_chunks()` lists only the
   `recordings/<user>/<rec>/<session>/` slice. This also fixes a **correctness**
   bug: the old `max_keys=10000` truncation could drop a late-sorting session's
   chunks entirely, so it never reconciled and was re-swept forever.

3. **No terminal cap.** A terminal meeting whose session never reconciles now
   counts attempts in `meeting.data`; after `UNFINALIZED_RECORDINGS_MAX_ATTEMPTS`
   it's marked `recording_finalize_abandoned` and excluded from the selection
   query, so the loop converges instead of re-listing storage forever.

## Tests

New/updated tests in `services/meeting-api/tests/`:
- session-scoped listing + `asyncio.to_thread` offload (no sync boto3 on loop)
- pathological prefix-count cap
- attempt-increment when a session has no reconcilable chunks
- abandon-at-cap exclusion
- `list_common_prefixes` for both Local and MinIO backends

Full meeting-api suite: **288 passed, 10 skipped** (Python 3.10).

## Notes for the hosted deploy

The hosted platform additionally re-vendors its Helm subchart to expose
meeting-api probe `timeoutSeconds`/`failureThreshold` knobs (mirroring the
`ttsService` fix from v0.10.6.1) so a single replica rides out slow spells —
tracked separately in the platform repo. This PR is the durable root-cause fix
for the OSS service.

### DIFF
--- a/services/meeting-api/meeting_api/storage.py
+++ b/services/meeting-api/meeting_api/storage.py
@@ -76,6 +76,19 @@ class StorageClient(ABC):
         memory growth from a user with very many chunks."""
         ...
 
+    @abstractmethod
+    def list_common_prefixes(self, prefix: str) -> list:
+        """List the immediate child "directories" under prefix (delimiter '/').
+
+        Returns a sorted list of common-prefix strings, each ending in '/'.
+        Lets callers enumerate per-recording prefixes
+        (`recordings/<user>/<rec>/`) cheaply — one entry per recording
+        rather than one object per chunk — so a session-scoped chunk search
+        doesn't have to page a heavy user's entire object list (which both
+        blocks the caller and can truncate before reaching a late-sorting
+        session's chunks)."""
+        ...
+
 
 class MinIOStorageClient(StorageClient):
     """MinIO/S3-compatible storage client using boto3."""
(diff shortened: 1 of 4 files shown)


DmitriyG228:

PR #476, opened 2026-07-06
fix(api-gateway): COPY guard_config.py into the image

rennf93:
## Problem

`services/api-gateway/main.py` does `from guard_config import apply_guard` (added in #465), but the api-gateway `Dockerfile` only `COPY`s `main.py` and `requirements.txt` by name — `guard_config.py` is never copied into the image. The image builds, but the container crashes on startup:

```
ModuleNotFoundError: No module named 'guard_config'
```

## Fix

One line — copy the guard config module alongside `main.py`:

```dockerfile
COPY ./services/api-gateway/guard_config.py /app/
```

## Scope

`guard_config.py` was introduced in #465 (merged into `release/0.10.7` on 2026-07-04); the Dockerfile wasn't updated alongside it. The same gap is present on `main`, `release/0.10.7`, and `0.11-integration` — happy to cherry-pick onto the release branch if that's useful.

### DIFF
--- a/services/api-gateway/Dockerfile
+++ b/services/api-gateway/Dockerfile
@@ -17,6 +17,7 @@ RUN pip install --no-cache-dir /app/services/meeting-api
 # Copy application code and requirements
 COPY ./services/api-gateway/requirements.txt /app/
 COPY ./services/api-gateway/main.py /app/
+COPY ./services/api-gateway/guard_config.py /app/
 
 # Install dependencies
 RUN pip install --no-cache-dir -r requirements.txt


DmitriyG228:
Checked in the tree, and you're right — this is a real boot-killer, verified exactly as described:

- `services/api-gateway/main.py:188` imports `guard_config` at module import time (`apply_guard(app)` at line 190) — on `release/0.10.7` and on the `v0.10.7` tag.
- `services/api-gateway/Dockerfile:18-19` copies exactly `requirements.txt` and `main.py` by name; `guard_config.py` never enters the image.
- `CMD ["uvicorn", "main:app", ...]` (Dockerfile:31) therefore dies on first import with `ModuleNotFoundError: No module named 'guard_config'`. The image builds green; the container is dead on arrival.

Your scope note about `main` was also correct when you filed: pre-0.12 `main` (90e5c726, which had merged `release/0.10.7` via #473) carried the same import and the same COPY gap. One small correction for the record: `0.11-integration` doesn't have the gap — `guard_config.py` was never merged there and nothing on that branch imports it.

**Why this one-liner can't land anymore, through no fault of yours:** on 2026-07-09, `main` was replaced by the 0.12 line (1e4bd329, released as v0.12.0 the next day). `services/api-gateway` no longer exists — the gateway was rewritten at `core/gateway/services/gateway`. Your diff was correct for the tree it targeted for the entire time that tree was `main`; the tree moved out from under it. It also sat here quiet for six days, which is longer than our alpha response promise — that part is on us.

**What survives this PR — two things, both yours as signal:**

1. **It exposes a real regression.** #465's per-IP layer did not survive the line adoption. The 0.12 gateway has a per-user limiter (`core/gateway/services/gateway/src/gateway/ratelimit.py`, enforced post-auth in `app.py:145-150`, keyed on `user_id`) — but nothing pre-auth: invalid-key floods and many-keys-from-one-IP are unthrottled at the edge, with no auto-ban. That is exactly the gap #465 was accepted to close. That regression is now a prepared issue — #544 — in the current shape: a declared acceptance floor, where presenting the table merges the PR — and it references this PR. **You're the natural owner**: fastapi-guard's maintainer, and author of the 0.10 integration. If you want it, claim it on #544. From a light read of #465, the pieces that carry over are the env-driven config surface, the exclude-paths sync with the per-key limiter, and the CORS/security-headers dedup rulings in your docstring; what changes is the target service — and this time, "the built image boots with the module present" is an acceptance row, not an accident.

2. **It names a failure class we're closing.** An image that cannot start passed every gate on the 0.10 line, because nothing ever ran the built image. The prepared issue's floor carries an image-boots leg (its row A7) so this class dies with the fix instead of recurring.

**Reopen condition for the one-liner itself:** the only live tree it applies to is `release/0.10.7`. The current ruling is to supersede the 0.10 line with 0.12 rather than patch it in place (the 0.10.x cut running in our deployments predates #465, so nothing live is affected). If a 0.10.7.x maintenance release is ever cut, your cherry-pick offer is the first thing it takes, and this PR is the reference diff.

Closing as superseded by the 0.12 adoption, with credit — the v0.12.0 release notes already thank this PR by number. Three PRs in, you keep finding real things, Renzo; hope to see you on the prepared issue.

PR #504, opened 2026-07-11
fix: fresh install setup wizard must end with a working bot (fixes #502)

### LINKED ISSUE #502: Fresh install, first run: the setup wizard must end with a working bot
### Why this matters

A fresh install should end with a working bot. On a clean machine today it doesn't: we ran the
guided path start to finish on a fresh Mac right after the release and hit three blockers inside
the wizard flow itself:

1. The wizard's "Claude subscription detected" check **cannot succeed on a fresh install** — it
   probes a container path that is empty unless a variable was set before boot, and the
   remediation it suggests can never work.
2. Transcription configured in the wizard tests **green**, but **every launched bot fails
   transcription anyway** — the launch path reads a different config source than the one the
   wizard wrote.
3. The transcript-by-meeting-link API endpoint returns `[]` on every fresh install — a key that
   provisioning mints is never wired into the environment.

None of these is deep; all three sit in the first hour of every new user and every evaluation.

### Value this issue delivers

> On a clean machine: `git clone && make lite` (or `make all`) → the wizard's every green means
> what it says → a bot joins a real meeting → the transcript comes back through the API — with
> **zero hand-editing of `.env`**.

### Where we are (honest — each claim from the clean-machine run, file:line in hand)

- **F10 (wizard credential detect):** the "Claude subscription on this machine" probe checks a
  container path that mounts empty unless `HOST_CLAUDE_CREDENTIALS` was set pre-boot; the
  wizard's "run `claude /login` and Re-check" remediation therefore can never turn it green on a
  fresh install. Correct setup docs exist (`configuration.mdx`) but no first-touch surface links
  them.
- **F11 (the lying green):** the bot-spawn transcription gate reads **process env only**, while
  the spawn itself resolves Settings-over-env a few lines later
  (`bot_spawn/service.py::_resolve_transcription_backend`). So wizard-configured (Settings-stored)
  transcription tests green but gates every spawn with a 503.
- **Bot API key:** provisioning mints a bot key and drops it — `deploy/compose/.env:122` ships
  empty, `deploy/lite/entrypoint.sh:59` defaults empty → the native-id transcript endpoint
  (`GET /transcripts/{platform}/{native_id}`) silently returns `[]` on every fresh install.
- **Quickstart prereqs:** `make` is not listed and build sizing (8 vCPU/16 GB) is undocumented —
  the first command fails on a minimal box before any of the above is even reachable.

---

### The components (validation waypoints of the ONE PR that delivers this issue)

- [ ] **C1** — The launch path trusts the same config the wizard wrote (kill the lying green)
- [ ] **C2** — The wizard's credential check can actually succeed (and its remediation is real)
- [ ] **C3** — The bot API key is wired at provision (native-id transcripts work out of the box)
- [ ] **C4** — The quickstart tells the truth about prerequisites

---

## C1 · The launch path trusts the same config the wizard wrote

**Target: ONE seam — the transcription gate in
`core/meetings/services/meeting-api/src/meeting_api/bot_spawn/service.py` (the env-only check vs
`_resolve_transcription_backend`, a few lines apart in the same flow).**

**Value:** if the wizard's transcription test is green, a launched bot transcribes. One source of
truth; no gate consults a config source the spawn itself doesn't.

**Prepared solution:**
1. Make the gate consult the SAME resolver the spawn uses (`_resolve_transcription_backend`:
   Settings > platform setting > process env) instead of raw env.
2. When the resolver yields nothing, the spawn fails with a TYPED reason ("no transcription
   backend configured — set it in Settings or env") — never a bare 503.
3. Unit test both paths: Settings-only config → spawn proceeds; nothing configured → the typed
   error.

**Along the way:** the gate exists to keep no-STT bots cheap → keep that behavior, just feed it
the resolved truth · resolver does I/O and the gate is hot → the spawn already calls it; reorder,
don't duplicate.

**Early validation:** unit red→green: a Settings-only backend passes the gate at head, fails at
base.

## C2 · The wizard's credential check can actually succeed

**Target: ONE seam — the credential-detect path the wizard probes (the container mount fed by
`HOST_CLAUDE_CREDENTIALS` + the wizard's Re-check remediation in the terminal's SetupGate).**

**Value:** on a fresh machine the wizard's check either turns green after the user does what the
wizard says, or tells them the one true thing to do — no circles.

**Prepared solution:**
1. At `make all`/`make lite` time, autodetect the host's Claude credentials (the standard host
   path) and wire the mount automatically when present (F10a — the small patch the finding
   proposed).
2. When absent, the wizard's failure text says exactly what works: the documented setup from
   `configuration.mdx`, linked — not a remediation that can't succeed.
3. Test: with a fake host credential file → detect green; without → the message contains the
   working instruction (assert the text, it's the product here).

**Along the way:** platform differences (macOS vs Linux paths) → detect both, document which was
found · security concern about auto-mounting credentials → mount read-only, and print WHAT was
mounted at startup so nothing is silent (that line is part of the fix).

**Early validation:** the two detect tests red→green; the misleading remediation text is gone
from the tree (grep shows zero hits).

## C3 · The bot API key is wired at provision

**Target: ONE seam — provision→environment (`deploy/lite/entrypoint.sh` + the compose provision
step that generates `.env`).**

**Value:** `GET /transcripts/{platform}/{native_id}` returns your words on a fresh install —
integrators' first API call works.

**Prepared solution:**
1. Where provisioning mints the bot key, write it into the runtime env exactly the way
   `ADMIN_API_TOKEN` already travels (lite: entrypoint export; compose: the generated `.env`).
2. README note for existing installs (one line: where to read the key, where to put it).
3. If the key is empty at bot start, log a typed warning naming the endpoint that will degrade —
   never silent `[]`.

**Along the way:** implemented as written but the endpoint still returns `[]` → the resolver has
a second break — that's an INVALIDATED observation we explicitly want; attach it.

**Early validation:** fresh-install integration check red→green: native-id query empty at base,
returns segments at head.

## C4 · The quickstart tells the truth about prerequisites

**Target: ONE surface — the quickstart (README + docs quickstart page): `make` in prereqs, build
sizing (8 vCPU/16 GB) stated.**

**Value:** the first command a stranger runs does not fail for a reason we knew about.

**Prepared solution:** add `make` to prerequisites; state build sizing honestly; one line on
what to expect on an undersized box (fails how, at what step).

**Early validation:** docs build; a reviewer on a clean box follows the quickstart top-to-bottom
without hitting an undocumented prerequisite.

---

### The acceptance table — present these observations and your PR merges. That's a promise.

If your bundle satisfies the table and something is still wrong, that's OUR bug in the table,
not yours. "Red→green" = same check failing at base commit, passing at head, both shown.

| # | Observation | Negative control (shown RED) | Anchor |
|---|---|---|---|
| A1 | C1 gate test red→green (Settings-only backend passes gate) | revert the gate change → red | test output, base+head shas |
| A2 | Typed no-backend error test green (bare 503 gone) | remove the typed error → red | test output |
| A3 | C2 detect tests red→green (fake host credentials → green; absent → true instruction in message) | point detect at wrong path → red | test output |
| A4 | C3: native-id query on a fresh install — `[]` at base, segments at head | unset the key at head → the typed warning appears (not silent `[]`) | curl output + bot log line, head sha |
| A5 | LIVE, clean machine: clone → `make lite` → wizard end-to-end (every green earned) → bot joins a real meeting YOU host → words back via native-id API | — | terminal transcript of the full session; meeting native id + timestamps; machine specs stated |
| A6 | C4: quickstart diff (make + sizing) present | — | the diff |
| A7 | No-regression: touched modules' lanes + repo gates green at head | — | CI at head sha |

**One person is enough** for A5 — you host the meeting, you admit the bot; this validates the
install path, not conversation. State your machine (OS, arch, CPU/RAM) in the bundle — a green
on arm64 AND amd64 from two signers is the ideal close, but one honest run merges the PR.

### Authorship

Any tools, agents included — this issue is written to be handed to one. No agent co-author
trailers: what you ship is yours — full responsibility, full authorship, full credit.

---

## Protocol retrofit — D5b · D6c · D12b · D7 (added at preparation-standard update)

**D5b verdict — bundle (one diff: the wizard flow; one value per first-run truth).** The wizard change carries several separately witnessable values — bot key wired without .env surgery; STT default honest (fully-local English or remote multilingual, chosen knowingly); the run ends with a bot that actually joins — each with its own acceptance row. The fresh-clone findings (F1–F11) fold into these rows as the red states.

**Deployments to validate (D12b).** Here the deployment IS the value: **Lite compose, fresh clone**, on **macOS and Linux x86 both** — two separate table rows, neither vouches for the other. Provenance is the whole proof: `git clone` timestamp, sha, OS + Docker versions, and ZERO manual env edits (any hand-fix voids the row — that is the discriminating condition). Long-lived checkouts and k8s are explicitly out of scope. arm64 is its own declared roadmap lane, not smuggled in here.

**Docs surface (D6c).** The quickstart is this issue's mirror — every wizard step must match the published quickstart page step-for-step (drift in either direction is a red row); the install how-to per OS; troubleshooting for the named failure forks (model-detect dead end, STT gate). The wizard's own on-screen text is user-facing documentation and is reviewed as such.

**Principle check (D7).** The defect class: a first-run path that ships assuming config it never wires (the bot key; the STT gate) — the config-contract-drift class (the 04-23 incident's kin, at install time instead of deploy time). Indicts the config-contract principle: a consumer of config that isn't produced by the path that ships it. The gate that should have caught it — a fresh-clone CI job that runs the wizard headlessly to a joined bot — is named TO BUILD in this issue.

**Validation request (D9, named).** Preferred signer: a genuine first-timer on a machine we've never touched — the value is defined from their chair; the owner's own Mac run (source of F1–F11) is the control, not the proof.

ShashwatSinha03:
### Summary

Fixes #502 — ensures a fresh `git clone && make lite` (or `make all`) results in a working bot end-to-end: wizard greens are truthful, a launched bot transcribes, and the transcript API returns words — with zero hand-editing of `.env`.

#### C1 — Transcription gate trusts the same config the wizard wrote
- **Problem:** `router.py` env-only `capability_state("stt")` gate rejected every spawn even after the wizard successfully configured transcription via Settings — the gate saw no env vars, while the spawn's `_resolve_transcription_backend()` found the Settings-stored config. The wizard's green was a lie.
- **Fix:** Moved the gate into `service.py` right after the resolver resolves, so it consults **Settings > env** — the same resolution the spawn trusts. Raised `TranscriptionNotConfigured` (typed, not bare 503). Router catches it → 503. No duplicate I/O.

#### C2 — Wizard credential check works
- **Problem:** The "Claude subscription detected" probe checked a container path that mounts empty unless `HOST_CLAUDE_CREDENTIALS` was set pre-boot; the "run `claude /login` and Re-check" remediation can never succeed on a fresh install.
- **Fix:** Updated `SetupGate.tsx` remediation to link to `configuration.mdx`. Updated server-side probe error to point to docs. Lite `Makefile` now autodetects `~/.claude/.credentials.json` and wires the mount automatically.

#### C3 — Bot API key wired at provision
- **Problem:** The native-id transcript endpoint (`GET /transcripts/{platform}/{native_id}`) silently returns `[]` on fresh installs — `VEXA_BOT_API_KEY` ships empty in `entrypoint.sh` and `.env.example`.
- **Fix:** `provision-key.sh` (lite) now mints **two** tokens — `VEXA_API_KEY` and `VEXA_BOT_API_KEY` — both with `bot,tx` scopes. `.env.example` (compose) updated with guidance to provision the bot key.

#### C4 — Quickstart tells the truth about prerequisites
- **Problem:** `make` not listed as prerequisite; build sizing (8 vCPU / 16 GB) undocumented.
- **Fix:** Added `make` to prerequisites. Documented build sizing and noted that `make lite` is recommended for smaller hosts.

### Acceptance verification

| # | Observation | Status |
|---|---|---|
| A1 | C1 gate: Settings-only backend passes gate (unit test) | ✅ `test_post_bots_transcribe_with_settings_stt_passes` |
| A2 | Typed no-backend error (not bare 503) | ✅ `test_post_bots_transcribe_without_stt_fails_loud` |
| A3 | C2: Wizard remediation links to docs, not impossible command | ✅ Text updated |
| A4 | C3: Bot API key minted at provision | ✅ Both tokens written to `key.env` |
| A5 | C4: Quickstart diff (make + sizing) | ✅ Both README.md and quickstart.mdx updated |
| A7 | All existing tests pass | ✅ 15 tests in test_bot_spawn.py pass |

### Related
- Fixes #502


DmitriyG228 (dismissed):
First contribution through the delivery loop — thank you for taking #502, and for taking all four components in one bundle as the issue asked. C1's mechanism is exactly the right shape, and your test updates are the honest kind of maintenance. Requesting changes on two correctness points plus the remaining acceptance rows; both fixes are small and this is close.

## 🔴 Two correctness fixes needed

**1. The moved gate fires *after* the meeting row is written — orphaned `requested` meetings.**

In `request_bot`, `repo.create_meeting_guarded(...)` runs before the transcription-resolution block where your new raise sits (`service.py`, step 3 vs step 4). The router gate you removed ran — per its own comment — *before any DB write*. Since `transcribe_enabled` defaults **true** (`router.py::_resolve_transcribe_enabled`, env default `"true"`), a fresh install with no STT configured now creates a `requested` meeting row on every `POST /bots`, then 503s. The retry after fixing config hits the dedup guard (`DuplicateMeeting` → 409) or the reopen path — plus DB litter of exactly the stuck-`requested` orphan shape documented at `bot_spawn/service.py:304`.

**Fix:** hoist the resolver block (env fallback + `_resolve_transcription_backend`) above the meeting-row write and gate there — "reorder, don't duplicate," as the issue's C1 fork says. The resolved values then flow down to the invocation build unchanged. (The issue's C1 text should have said "before the meeting row is written" in those words — that imprecision is on our side of the plan-bug rule; a sharpening comment is going on #502.)

**2. Lite automount exports `HOST_CLAUDE_CREDENTIALS` as the *host* path — the worker can't read it, but the wizard goes green.**

`deploy/lite/entrypoint.sh:62-65` is explicit: in lite, `HOST_CLAUDE_CREDENTIALS` is the path **inside** the container ("mount it in … and set `HOST_CLAUDE_CREDENTIALS=/claude-creds.json`. Lite's runtime uses the process backend, so the worker reads the file directly"). Your mount target `/var/lib/vexa/host-claude-credentials` correctly satisfies the wizard probe's fixed `CREDS_PATH` (`core/agent/control_plane/config_test.py:28`) — but the env you export is the host path, which doesn't exist in-container. Result: wizard Test button green, worker + config.v1 `/health` probe red. That's a new lying-green — the exact class this issue exists to retire.

**Fix:** `-e HOST_CLAUDE_CREDENTIALS=/var/lib/vexa/host-claude-credentials`. Compose's semantics genuinely differ (there it *is* a docker-host path, see `deploy/compose/docker-compose.yml:159`) — that's the trap; worth a one-line comment in the Makefile so the next person doesn't fall in it.

## 🟡 Acceptance floor — #502's table as declared (the PR body renumbered it)

| Row | Status |
|---|---|
| A1/A2 — C1 gate tests, typed error | ✅ present and discriminating (`test_post_bots_transcribe_with_settings_stt_passes` asserting `BOT_CONFIG.transcriptionServiceUrl` is exactly right) — attach the red→green base/head run to close the row |
| A3 — C2 detect tests (fake creds file → green; absent → true instruction in message) | ❌ missing — C2 currently ships text changes with no test |
| A4 — native-id `[]`→segments, negative control: **typed warning when the key is empty** (never silent `[]`) | ❌ the typed warning (issue C3 step 3) isn't implemented, so the control can't be shown |
| A5 — live leg: clean machine, clone → `make lite` → wizard end-to-end → bot in a real meeting → words via native-id API | ❌ not presented — and finding 2 above is precisely the bug only this row catches |
| A6 — quickstart diff | ✅ |
| A7 — touched modules' lanes + repo gates green at head | ⚠️ 15 unit tests claimed; fork PRs don't run the gate suite in CI — run `node scripts/gates.mjs all` locally and attach the tail |

Per the issue's own promise: present the table and this merges. The floor is the floor — nothing beyond it will be asked.

## 🟢 Verified good

- C3's lite wiring works end-to-end: `key.env` → the terminal's `sh -c 'set -a; . /run/vexa/key.env; …'` (supervisord.conf) overrides the empty `environment=` injection, and `clients/terminal/server.mjs:43` is the real consumer of the fallback chain. Checked, correct.
- The `.env.example` comment documents the auth chain accurately. C4 is truthful.

## Minor (non-blocking, batch with the fixes if convenient)

- `SetupGate.tsx` points end users at `docs/docs/configuration.mdx` — a repo source path. Link the published docs page instead; same in the server-side probe message, which also lost the mechanism detail (`HOST_CLAUDE_CREDENTIALS` unset / file missing) — keep the mechanism *and* add the pointer. The macOS Keychain remedy (`config_test.py:31-33`) shouldn't disappear from that path either.
- #502's C2.1 asks for autodetect at `make all` *and* `make lite`; compose still requires hand-setting the variable. Fine to scope compose out explicitly — just say so in the PR body rather than silently.
- `$$CLAUDE_MOUNT` is unquoted and breaks on a `$HOME` containing spaces — rare, cheap to guard.
- The removed router comment anchored the old gate to ADR-0026 (canonical config.v1 capability gate). The issue sanctions the divergence, but land a sentence noting the capability tri-state still drives boot preflight + `/health` while spawn now trusts the resolver — otherwise the next reader "fixes" it back.

**Merge path:** the two 🔴s, the A3 tests + A4 typed warning, gates output, then the A5 live run (a throwaway VM per the deployment docs works fine). Ping here when the table's presented — validation gets arranged fast.

DmitriyG228:
**The full path to merge, stated up front** — so nothing surprises you after the fixes land. Two human layers sit beyond the code review, per `docs/DELIVERY.md` (D8–D12b). This is the first PR through this loop, so spelling it out here also documents it for everyone after you.

## Layer 1 — your own observation bundle (D8/D10)

The PR carries two artifacts judged on different axes: the **diff** (is it correct and safe? — that's the review above) and the **observation bundle** (is the value real?). The repo's PR template has the exact sections — worth restructuring the PR body into it. Concretely:

- Map each row of #502's acceptance table **as declared** (A1–A7) to its evidence: red→green outputs with base+head shas, negative controls shown red, anchors.
- **A5 is your own live run**, not someone else's: clean machine (a fresh VM is fine), clone → `make lite` → wizard end-to-end with every green earned → bot joins a real meeting you host → words back via the native-id API. Attach the terminal transcript, the meeting native id + timestamps, machine specs.
- Present the table and this merges — that's #502's promise, and we hold it. Rows you exceed with new witnessed value are welcome and credited, never scope creep.

## Layer 2 — the non-author value attestation (D9)

After your bundle is up, a **non-author** witnesses the value live and signs. The validator can be any competent non-author; the preferred signer for #502 is the originating reporter of the fresh-install findings (the maintainer's fresh-Mac run that produced F10/F11 — so validation will be arranged fast, likely same-day).

What the attestation is: the validator runs the flow themselves and states, in their own words, *"I witness the value: on a fresh install, every wizard green meant what it said, the bot transcribed, the transcript came back through the API."* It is **multichannel** — what the bot did (logs), what the human saw (wizard, transcript), what the instruments recorded (test output, API responses) — and the channels must **corroborate**. A human/instrument divergence is a first-class finding that blocks merge until reconciled; the human's eyeball carries no special authority over the instruments, nor vice versa.

Per **D12b**, the attestation names its setup and build provenance: which deployment was validated (Lite here — it takes the most divergent path, which is why #502 runs on it), repo sha, fresh-clone vs long-lived, and any env deltas from stock. "Works on my install" without how the install was made is an unanchored claim. Deployments the issue names but nobody ran stay honestly unclaimed.

Also per **D6c**: the attestation covers the docs story — your C4 changes land on the right pages at the right altitude, consistent with how the docs already teach.

## Sequence

1. The two 🔴 fixes + A3 tests + A4 typed warning (review above)
2. Gates locally: `node scripts/gates.mjs all` at head, attach the tail (fork PRs don't get the suite in CI)
3. Your A5 live run → bundle complete, table presented
4. Non-author attestation posted here (we'll arrange the signer the moment you ping)
5. Merge. Your name goes in the release notes as author; the validator's as signer (D15/D16).

The `value-signed` status check is being built; until it lands this protocol is enforced by hand, which is what this comment is. Questions about any step — ask here, answers become part of the protocol's documentation.

DmitriyG228:
**pr-value — autonomous value leg: ✅ GREEN** ([run 29261417440](https://github.com/Vexa-ai/vexa/actions/runs/29261417440), merge ref of this PR on current main).

The full stack built from this PR's tree passed the product FSM end-to-end — including the STT-config gate paths your change touches (spawn with transcribe_enabled=false, canaccess-deny). This machine-presents the backend acceptance rows; what remains from the earlier CHANGES_REQUESTED review: the 2 flagged correctness fixes, and the user-visible wizard→working-bot story, which will be witnessed in the release's combined witness pass rather than by a separate hand-made demo. The path to merge just got shorter.

DmitriyG228:
**Maintainer fix-forward** (owner ruling: fix the two flagged bugs in place and prove the flow autonomously). Three commits pushed: 51a47da, 44aa6c7, 788a96b.

## The two review findings — fixed

**Finding 1** — *"the moved gate fires after the meeting row is written — orphaned `requested` meetings"* → fixed in `core/meetings/services/meeting-api/src/meeting_api/bot_spawn/service.py` (51a47da). Reorder, not duplicate: the resolver block (env fallback + `_resolve_transcription_backend`) is hoisted above `create_meeting_guarded` as step 1b; the resolved url/token flow down unchanged to the invocation build. The PR had also left `tests/test_config_contract.py` red (the old capability-gate assertions); those tests now encode the resolver contract, including the discriminating assertion — **pre-fix** the new test fails exactly here:

```
>       assert repo._meetings == {}, f"refused spawn wrote a meeting row: {repo._meetings}"
E       AssertionError: refused spawn wrote a meeting row: {1: {...status: 'requested'...}}
```
**post-fix**: `607 passed, 2 skipped` (full meeting-api suite).

**Finding 2** — *"lite automount exports HOST_CLAUDE_CREDENTIALS as the host path — worker can't read it, wizard goes green"* → fixed in `deploy/lite/Makefile` (44aa6c7): `-e HOST_CLAUDE_CREDENTIALS=/var/lib/vexa/host-claude-credentials` (the mount target), with the requested comment on why lite (in-container path) and compose (docker-host path) genuinely differ.

## Autonomous flow proof — red→green on the real stack

New `deploy/compose/tests/wizard_stt_flow_test.py` (788a96b), one stack / one user / one `native_meeting_id`:

1. **RED (negative control):** stack boots with no `TRANSCRIPTION_SERVICE_URL` env → default (transcription-on) `POST /bots` → **503** with the typed reason `no transcription backend configured — set it in Settings or environment variables …`, and the test asserts **no meeting row exists** (finding 1's regression guard at L3).
2. **GREEN:** STT configured via `PUT /internal/settings/transcription` on admin-api — the exact endpoint the wizard's Settings surface calls (`clients/terminal/src/app/api/admin/settings/[key]/route.ts` is a verbatim proxy: same method, body, `X-Internal-Secret`) → the **identical** `POST /bots` retried → **201 accepted** (also proving the retry can't 409 on a leftover row). Gate checks backend *presence* (what the resolver returns), not liveness — a stub URL is sufficient and honest.

Run (MOCK_BOT=1, `BROWSER_IMAGE=mock-bot:dev`, full compose stack up/down inside the run):

```
$ MOCK_BOT=1 BROWSER_IMAGE=mock-bot:dev ./bin/stack-test wizard_stt_flow_test.py -x
1 passed in 50.49s
```

**What this does and doesn't claim:** it proves the wizard's *write path end-to-end at the API layer* — the terminal route is a verbatim proxy to the endpoint the test drives, so the config the UI writes is byte-identical. It does **not** drive the React form itself (no headless-browser leg here); the wizard form → route hop remains covered by the A5 live run / release witness pass, as your pr-value comment scoped.

## Also in the batch (your "minor" list)

- Probe + `SetupGate` messages: mechanism detail (`HOST_CLAUDE_CREDENTIALS` unset / file missing) restored — the PR's reword had broken `core/agent/tests/test_config_test.py::test_subscription_missing_file` — and both now point at the published https://docs.vexa.ai/configuration instead of the repo source path. Agent suite: `16 passed`.

Remaining from the review (A3/A4 rows) unchanged here. Merge follows on green checks + the witness pass — not merging from this thread.

ShashwatSinha03:
@DmitriyG228 — thank you for the review; this was the kind of review that makes a contributor better, not just a PR greener.

Two things in particular landed:

- **You caught two real bugs, not nitpicks.** The gate-after-insert orphaning `requested` meetings (finding 1) and the Makefile exporting `HOST_CLAUDE_CREDENTIALS` as the *host* path so the wizard went green while the worker stayed red (finding 2) — both are exactly the "lying green" class #502 exists to retire. Catching the second one on *reading* is the part I'd have missed until a human hit it live.

- **You fix-forwarded instead of bouncing.** Pushing `51a47da` / `44aa6c7` / `788a96b` — hoisting the resolver, correcting the mount target, and adding `wizard_stt_flow_test.py` as a red→green regression guard — meant the fixes came back proven, with the negative control (no-STT boot → 503 *and no row written*) encoded as a test. That's a higher standard than the diff I shipped, and the right one.

Also genuinely grateful for the patient write-up of the observation/attestation protocol in `docs/DELIVERY.md` — spelling out Layer 1 / Layer 2 up front turned "merge when it's done" into something I could actually execute against. The autonomous flow proof you added is better than the hand-made demo I'd have posted.

Learned a lot about this codebase from this one. Appreciate the standard you hold. 🤝

Issue #568, opened 2026-07-13
docs: GUARD_ENABLE_REDIS=false + uvicorn --workers N degrades guard to per-process buckets


DmitriyG228:
Found during the #555 fix-forward review (edge layer contributed by @rennf93 — thanks).

One documentation sentence is missing from the edge-guard docs (gateway README / GUARD_* surface docs): with `GUARD_ENABLE_REDIS=false` and `uvicorn --workers N`, rate-limit buckets and auto-bans are per-process — effective limit becomes N x GUARD_RATE_LIMIT_RPM and bans do not propagate across workers. Operators tuning limits without Redis need this stated where the knobs are documented.

rennf93:
Fixed in #573 (merged)

Issue #567, opened 2026-07-13
edge_guard _env_bool accepts only literal 'true' — unify truthiness with sibling ratelimit parser


DmitriyG228:
Found during the #555 fix-forward review (edge layer contributed by @rennf93 — thanks).

`_env_bool` in `core/gateway/services/gateway/src/gateway/edge_guard.py` treats only the literal string `true` (case-insensitive) as truthy, while the sibling per-user ratelimit config accepts `1`/`yes`/`on` as well. `GUARD_ENABLED=1` silently disables the guard the operator thinks they enabled.

Fix: single shared truthy-set helper (`{"1","true","yes","on"}`) used by both parsers; add a unit test pinning the accepted forms.

rennf93:
Fixed in #573 (merged)

DmitriyG228:
**Era-check (autonomous session, 2026-07-18 00:36): already delivered — merged PR #573 (commit `a606f9ba`, 2026-07-13) shipped exactly this issue's prepared solution.** #573 referenced #567 without `Closes`, so the issue stayed open and `state: ready` kept it in the claimable queue — the #712 class.

Verified on main, row by row:
- Single shared truthy set `{"1","true","yes","on"}`: `core/gateway/services/gateway/src/gateway/ratelimit.py:21-26` (`env_truthy`, strips + lowercases).
- `edge_guard._env_bool` delegates to it: `edge_guard.py:77-82` — `GUARD_ENABLED=1` now enables (the headline bug).
- Sibling parsers converge: `ratelimit.py:73`, `app.py:588-590` (WS reader).
- Pinning test: `tests/test_edge_guard.py:123-144` (docstring cites #567). Suite run on main: **90 passed, 1 xfailed**.

**Recommended disposition: close as completed-by-#573**, which also clears the stale ready-label from the claimable queue. (Adjacent fact, out of this issue's scope: `transcription/main.py:52-56` has its own broad `_env_bool` — a third copy of the idea in a different service, not the bug class.)

DmitriyG228:
Closed as completed by #573 (a606f9ba) — verification in thread. Credit @rennf93.

Issue #566, opened 2026-07-13
edge_guard WS guard runs after accept() — banned IPs still cost an upgrade + frames per connect


DmitriyG228:
Found during the #555 fix-forward review (edge layer contributed by @rennf93 — thanks).

The WS connect guard in `core/gateway/services/gateway/src/gateway/edge_guard.py` performs its ban/blacklist check after `websocket.accept()`. A banned IP therefore still gets a completed WebSocket upgrade plus at least one frame exchange per connection attempt before being closed — the guard pays the resource cost it exists to avoid under a reconnect-hammer.

Fix: check ban/blacklist before accepting (close with 403-equivalent via raising before accept, or `websocket.close()` pre-accept per Starlette semantics), so a banned IP is rejected at the cheapest possible point.

rennf93:
Fixed in #573 (merged)

DmitriyG228:
**Era-check (autonomous sweep, 2026-07-18): delivered in-tree by merged PR #573** (same PR that fixed #567; its title lists #565 #566 #567 #568 — the two siblings got closed, these two didn't).

Tree evidence (main @ 9e8ebd47): `core/gateway/services/gateway/src/gateway/app.py:581-597` — the full guard check runs **before** `ws.accept()`, closing pre-accept with 4401; `edge_guard.py:288` `ws_guard_check`. Banned IPs no longer cost an upgrade + frames.

**Recommended disposition: close, credit #573.**

DmitriyG228:
Closed as delivered by #573 — guard runs pre-accept (app.py:581-597), verification in thread.

PR #555, opened 2026-07-13
gateway: port fastapi-guard edge layer to 0.12 — per-IP throttle + auto-ban + static blocklist (closes #544)

rennf93:
Closes #544. Ports the fastapi-guard edge layer that #465 merged on the 0.10 line into the 0.12 gateway carve — per-IP pre-auth throttle (V1), auto-ban of repeat offenders (V2), and static IP allow/deny + optional geo/cloud blocking (V3), all env-driven behind a `GUARD_ENABLED` kill switch. One middleware-wiring diff carries all three values; each has its own acceptance row below.

**Authorship disclosure (up front, per the #465/#544 house standard):** I'm @rennf93 — author of #465 and #476, and the maintainer of `fastapi-guard` / `guard-core` (the dependency this diff adds). That's the conflict, disclosed in the open; the value-signing bar is a non-author rerunning the live legs (A9/A10) — see "How this issue closes" in #544. No agent co-author trailers; what ships here is mine.

## What changed

- `core/gateway/services/gateway/src/gateway/edge_guard.py` (NEW, 298 lines) — `build_guard_config()` (env surface), `apply_guard(app)` (kill switch on `GUARD_ENABLED`), `_WsGuard` in-memory per-IP limiter + auto-ban gated on `cfg.enable_ip_banning` with reset-on-set, `resolve_ws_client_ip()` mirroring guard_core's rightmost-XFF handling. Carries the 0.10.7 docstring rulings verbatim: request-body WAF OFF (the gateway proxies arbitrary user text), CORS/security-headers OFF in guard (one-writer), redis namespaced `vexa:guard:` on the existing `REDIS_URL`, `fail_secure=False`, `lazy_init=True`.
- `adapters.py` — `apply_guard(app)` wired into `build_production_app()` ONLY. `create_app()` stays pure so both harnesses (unit + conformance, A6) are unaffected by construction.
- `app.py` — the `/ws` multiplex gets an opt-in guard hook (`GUARD_WS_ENABLED`, default false, lazy import) since Starlette HTTP middleware doesn't intercept WebSocket scopes (open question 3 in #544, signed as found: WS guard is a standalone in-process limiter, not SecurityMiddleware's path — see the `edge_guard.py` module docstring's "ponytail" note for the ceiling).
[...]

PR #573, opened 2026-07-13
fix(gateway/edge_guard): post-merge review fixes — WS CIDR matching, check-before-accept, shared truthy helper, per-process-without-redis docs (#565 #566 #567 #568)

### LINKED ISSUE #565: edge_guard WS mirror: trusted-proxies/whitelist matching is exact-string, HTTP path accepts CIDR
Found during the #555 fix-forward review (edge layer contributed by @rennf93 — thanks).

In `core/gateway/services/gateway/src/gateway/edge_guard.py`, the WS connect guard mirrors the HTTP guard's IP checks but matches `GUARD_TRUSTED_PROXIES` / `GUARD_IP_WHITELIST` entries as exact strings, while the fastapi-guard HTTP path accepts CIDR ranges. An operator setting `GUARD_TRUSTED_PROXIES=10.0.0.0/8` gets correct HTTP behavior and silent no-match on WS.

Fix: parse entries with `ipaddress.ip_network(..., strict=False)` in the WS mirror and match by containment, same semantics as the HTTP layer.

rennf93:
Fixes the four post-merge review findings DmitriyG228 opened on #555 (#565, #566, #567, #568) — all in `edge_guard.py` (+ wiring in `app.py`, + docs). One commit, one PR as discussed.

**Authorship disclosure (same as #555):** I'm @rennf93 — the `edge_guard.py` author and the `fastapi-guard` / `guard-core` maintainer. The dependency this layer uses is mine; the findings are about code I contributed. No agent co-author trailers.

## The fixes

**#565 — WS CIDR matching.** The WS mirror matched `GUARD_TRUSTED_PROXIES` / `GUARD_IP_WHITELIST` / `GUARD_IP_BLACKLIST` as exact strings, while the fastapi-guard HTTP path accepts CIDR — so `GUARD_TRUSTED_PROXIES=10.0.0.0/8` worked for HTTP and silently no-matched on WS. Added `_ip_matches(client_ip, entries)` (`edge_guard.py`) parsing each entry with `ipaddress.ip_network(entry, strict=False)` and matching by containment (a bare IP becomes `/32`/`/128` and self-matches). Used at the three sites: whitelist bypass, blacklist deny, trusted-proxies resolution. Malformed entries are skipped (fail-open, consistent with `fail_secure=false`) — this deliberately diverges from the HTTP path, which raises on a malformed CIDR; a typo'd `GUARD_IP_BLACKLIST` entry silently does not block, the safe-direction tradeoff (called out in the docstring).

**#566 — WS guard check before accept.** The ban/blacklist check ran *after* `ws.accept()`, so a banned IP still got a full WebSocket upgrade + frames. Moved the full guard check (whitelist/blacklist/ban/rate-limit) to before `await ws.accept()`; on denial, `await ws.close(code=4401)` pre-accept. Order is now **guard → accept → auth** (auth not reordered relative to accept). The highest-risk claim — that a pre-accept `websocket.close` produces an HTTP 403 — was verified by source-reading, not trusting docs: Starlette's state machine accepts `websocket.close` while `CONNECTING` (it forwards it unchanged), and **uvicorn 0.51 turns that into a 403 to the upgrade** across all three WS protocol impls (`websockets_sansio_impl` / `websockets_impl` / `wsproto_impl`). No data frame is sent (a `send_text` before accept is invalid ASGI). 4401 is in the 4000-4999 app-defined range and already used by the auth-missing-key path.

**#567 — shared truthy helper.** `_env_bool` accepted only the literal `true` while the sibling `ratelimit.from_env` accepted `1/true/yes/on` — so `GUARD_ENABLED=1` silently disabled the guard (and `app.py`'s `GUARD_WS_ENABLED` check had the same bug). One shared `env_truthy` helper (`_ENV_TRUTHY = {"1","true","yes","on"}`) in `ratelimit.py`, used by `edge_guard._env_bool`, `ratelimit.from_env`, and `app.py`'s `GUARD_WS_ENABLED` check. Placed in `ratelimit.py` because it's the lowest-level module (no guard dependency) → no circular import, smallest diff, no new file. `_env_bool`'s `default`-when-unset fallback is preserved (`raw is None → default`; empty/`0`/`false` → False, not default — matching the old behavior exactly).

**#568 — docs.** With `GUARD_ENABLE_REDIS=false` + `uvicorn --workers N>1`, the HTTP limiter is per-process: effective limit `N × GUARD_RATE_LIMIT_RPM`, bans don't propagate across workers. One sentence added to `docs/docs/configuration.mdx`, `core/gateway/README.md`, and `deploy/lite/supervisord.conf` (the WS per-process ceiling was already documented; the HTTP one was not).

## How this was produced (drill)

sonnet develop → sonnet `fable-mode:critic` adversarial → I reviewed. The critic's job was to refute each fix; it confirmed all four survive attack with no blocks-merge / should-fix defects, verifying the load-bearing claims by source-reading (uvicorn→403, `env_truthy` default-fallback, guard_core HTTP CIDR semantics, and the gate baselines against clean `upstream/main`). Two nits it raised — a docstring that overclaimed "mirrors the HTTP path" and a comment that attributed the 403 to Starlette rather than uvicorn — were applied (both wording-only).

## Gates

- `pytest -q`: **61 passed, 1 xfailed** (was 58/1 at #555; +3 new tests — CIDR matching on the WS path, CIDR whitelist bypass, `env_bool` truthy-set pinning. `TestWsGuard` assertions updated for pre-accept rejection: `close_code == 4401`, `accepted is False`, `not ws.sent`).
- `mypy edge_guard.py app.py ratelimit.py`: 4 errors, **all pre-existing** (verified on clean `upstream/main` — `ratelimit.py:64/72` and `app.py:91/552`, the latter two outside the diff region). No new errors.
- `ruff check`: the 2 `E702` at `app.py:654-655` are **pre-existing** on `main` (from the #543 Jitsi merge — `plat = item.get("platform"); nid = ...`, not my lines). No new findings.
- `ruff format --check`: `edge_guard.py` and `test_edge_guard.py` are clean. `app.py` and `ratelimit.py` show pre-existing format drift on `main` (also from #543); I deliberately did **not** reformat them — my additions are hand-formatted to match surrounding style, and reformatting would bloat the diff with unrelated pre-existing changes. Flagging here so a reviewer running `ruff format --check` locally isn't surprised.

## Files

`core/gateway/services/gateway/src/gateway/edge_guard.py`, `…/app.py`, `…/ratelimit.py`, `…/tests/test_edge_guard.py`, `core/gateway/README.md`, `deploy/lite/supervisord.conf`, `docs/docs/configuration.mdx` — 7 files, +177/−33.

Closes #565, #566, #567, #568.


DmitriyG228:
**TAKE review — verdict: SOUND. Exemplary follow-up.** All four issues addressed exactly, each with discriminating tests (16/16 green in a real venv): CIDR matching at all three sites with the dangerous directions fail-safe (malformed whitelist/trusted-proxy entries skip → deny/ignore-XFF), pre-accept rejection proven by the `accepted`-flag instrument (and correctly explained — the old post-accept frame only worked because the guard ran late), and the #567 parser unification does NOT flip the ON-default (unset → ON pinned by test; `1/yes/on` now enable instead of silently disabling — the actual fix). Docs sentence present in all three surfaces. Zero scope creep; bundle ruled coherent (four sibling findings from one review of one module — splitting would have made four conflicting PRs on the same 30 lines).

Two nits for whenever (not blockers, no issue needed): an IPv6 + malformed-CIDR test case would pin `_ip_matches` edges; the function-local `env_truthy` import in app.py could be module-level.

Merging on green checks — this ships in v0.12.2, same day as your report. Thank you for coming back for the follow-ups this fast; this is the loop working exactly as designed.

Issue #592, opened 2026-07-13
Jitsi bot stuck 'waiting for admission' on an open meet.jit.si room — never enters, invisible to participants (live-witnessed v0.12.3)


DmitriyG228:
cc @AzimovS (Jitsi platform author, #543) — hitting this during the v0.12.3 **release witness** on a fresh lite deploy. The join adapter is genuinely working up to a point, then the **admission detection appears to false-positive** and the bot loops forever. Would value your eyes on `jitsi/admission.ts`.

## Symptom (live)
Bot is added to a public `https://meet.jit.si/testtest` room. It **navigates, enters the name, clicks Join**, then reports `Checking admission state…` → `Still waiting for admission… Ns elapsed` **indefinitely (>100s, never resolves, never times out)**. A human sitting in that same room **does not see "Vexa" join at all** (not as a participant, not as a lobby knock).

## Evidence (bot log, mtg-5, v0.12.3 published lite)
```
[Jitsi] Navigating to: https://meet.jit.si/testtest#config.startWithAudioMuted=true&config.startWithVideoMuted=true&userInfo.displayName=%22Vexa%22
[Jitsi] Name entered: "Vexa"
[Jitsi] Join clicked — waiting for the conference to load...
[Jitsi] Checking admission state...
[Jitsi] Still waiting for admission... 2s elapsed
... (loops to 109s+, never resolves)
[bot] lifecycle report 'needs_help' rejected: Error: lifecycle.v1: illegal transition joining → needs_help
```

## What works vs what's broken
- ✅ **Join path works**: navigate → prejoin → name → Join click (verified; on a *different* room earlier the bot correctly reached `Bot is in the lobby — waiting for a moderator to admit`, so the adapter CAN detect a real lobby).
- ❌ **Admission-state detection looks like a false positive** on an OPEN room: a fresh meet.jit.si room has **no lobby by default**, so `waiting for admission` should not trigger — yet the bot sits there and never appears as a participant. Either (a) the post-Join conference-load detection in `jitsi/admission.ts` mis-reads the joined-and-open state as "waiting", or (b) the bot never actually completes entry into the conference after the Join click.

## Two distinct findings
1. **Primary — admission detection / conference-entry** (`core/meetings/modules/join/src/jitsi/admission.ts`): on an open room the bot should transition to `active`/`joined` and become visible; instead it loops `waiting for admission` and is invisible to participants. Need to distinguish "real lobby" from "already in an open conference" from "stuck post-join".
[...]

PR #597, opened 2026-07-14
fix(jitsi): recognize members-only lobby authoritatively + typed admission timeout (#592)

AzimovS:
## What

Fixes the real defect behind #592. Investigating on the published v0.12.3 image against real meet.jit.si showed the issue's "open room / no lobby / admission false-positive" premise doesn't hold: **fresh meet.jit.si rooms are members-only** until a moderator arrives. The conference join fails with `MEMBERS_ONLY` and the error page renders none of the lobby affordances, so the DOM/text lobby check could miss it — the bot then treated a real lobby as an "unknown" state and escalated down the illegal `joining → needs_help` transition, which was swallowed, so it looped silently.

## Changes

- **Authoritative lobby detection** (`jitsi/admission.ts`): latch the `MEMBERS_ONLY` conference-failure host-side from the page console (reliable even when the DOM shows nothing) as the primary signal, with the app's redux verdict (`features/base/conference.membersOnly`) and the existing DOM checks as fallbacks — mirroring how admission already trusts `APP.conference.isJoined()`. Wired in from `join.ts` before navigation so the event is never missed.
- **Typed admission errors** (`shared/admission.ts` + `jitsi/admission.ts`): the Jitsi admission wait now throws a typed `AdmissionError` (moved to `shared/` so every platform throws the class the JoinDriver maps) — a lobby timeout → `awaiting_admission_timeout`, a decline → `awaiting_admission_rejected` — instead of a bare `Error` that collapses to a generic `join_failure`.
- **Test** (`jitsi/admission.test.ts`): fabricated `APP`/`Page` — members-only → `lobby`; timeout throws `AdmissionError('lobby_timeout')`.

## Testing

- Join module suite green (`npm test`), isolation gate green, `tsc` clean.
- **Live on the published v0.12.3 image against real meet.jit.si**: with no moderator, the members-only lobby is detected via the latch + redux for the full wait (the silent-loop path is gone); with a moderator present, the bot enters and produces a transcript.

## Follow-up (out of scope here)

The *accurate completion reason* for a stuck bot additionally needs a meetings-lifecycle change: a bot waiting in the lobby is torn down by the pre-active workload teardown (~5 min) **before** its own admission timeout fires, and that path stamps the fail-safe `join_failure` rather than `awaiting_admission_timeout`. Both are transient/retry, so behavior is unaffected — it's a precision issue, filed separately. This PR makes the bot-side path correct so it produces the right reason once the teardown is status-aware (or the timeouts are aligned).

### DIFF
--- a/core/meetings/modules/join/package.json
+++ b/core/meetings/modules/join/package.json
@@ -11,7 +11,7 @@
   "files": ["dist"],
   "scripts": {
     "build": "tsc",
-    "test": "tsx src/shared/selector-validity.test.ts && tsx src/googlemeet/admission.test.ts && tsx src/googlemeet/humanized/humanized.test.ts && tsx src/msteams/modals.test.ts && tsx src/zoom/join.test.ts && tsx src/jitsi/join.test.ts && tsx src/jitsi/password.test.ts",
+    "test": "tsx src/shared/selector-validity.test.ts && tsx src/googlemeet/admission.test.ts && tsx src/googlemeet/humanized/humanized.test.ts && tsx src/msteams/modals.test.ts && tsx src/zoom/join.test.ts && tsx src/jitsi/join.test.ts && tsx src/jitsi/password.test.ts && tsx src/jitsi/admission.test.ts",
     "check:isolation": "node scripts/check-isolation.js"
   },
   "dependencies": {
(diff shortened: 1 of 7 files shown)


AzimovS:
Follow-up filed for the meeting-api half (accurate completion reason on pre-active teardown): #598.

Issue #579, opened 2026-07-13
0.12 gateway dropped native-id-keyed api.v1 routes → breaks 0.10 dashboard/clients despite sealed contract


DmitriyG228:
### Why this matters

The 0.12 gateway is described as preserving a **sealed `api.v1`**, but several routes were **re-keyed from `(platform, native_meeting_id)` to numeric row-id, or dropped**, so a 0.10 `api.v1` client — including the shipped 0.10 dashboard (`vexaai/dashboard:0.10.6.3`) — 404s on everyday actions against a 0.12 core. Found while keeping the 0.10 dashboard against a 0.12 backend: it boots, logs in, lists meetings, and streams live transcripts (those contracts genuinely are sealed), but **rename meeting, delete meeting, chat, share-transcript, and deferred re-transcribe all 404**, and **bot-running detection silently reads false** because a response field was renamed. To an operator this reads as "the dashboard is broken," and any external `api.v1` integration built on native-id keying breaks the same way.

### Value this issue delivers

> A 0.10 `api.v1` client (incl. the 0.10 dashboard) can rename and delete a meeting, and correctly read bot-running state, against a 0.12 gateway.

(Per D5b these are one value — "native-id `api.v1` continuity" — realized by one PR of thin passthroughs. The chat/share/transcribe routes are listed as optional components to keep or consciously drop.)

### Where we are (honest)

- **PATCH / DELETE meeting — native path unrouted.** Gateway only has `PATCH /meetings/{meeting_id}` and `DELETE /meetings/{meeting_id}` typed `int` — `core/gateway/services/gateway/src/gateway/app.py:360-365`; meeting-api `collector/app.py:358-360`. A 0.10 client calls `.../meetings/{platform}/{native_id}` (3-segment) → no match → 404. Rename (notes/participants/languages) and delete are dead.
- **`/bots/status` field renamed.** 0.12 returns `{"running": [...], "count": N}` — `core/meetings/.../collector/app.py:244`. 0.10 clients read `running_bots`; missing field → bot-running indicator/stop-gating silently always false (HTTP 200, wrong shape, no error surfaced).
- **share-transcript moved.** Mint route is now `POST /meetings/{platform}/{native}/share` (`gateway/app.py:371-373`); the 0.10 `POST /transcripts/{platform}/{native}/share` path is gone → 404.
- **deferred transcribe dropped.** No `POST /meetings/{id}/transcribe` route in 0.12 → 404.
[...]

PR #672, opened 2026-07-16
test(compat): drop stale strict-xfail — #579 restored the 0.10 bots/status envelope

DmitriyG228:
The v0.12.7 `release-images` run failed only on `validate/validate-v010-compat`: **XPASS(strict)** on `test_05_bots_status_v010_envelope`. The test documented the V010-BREAK as a strict expected-failure; PR #666 (#579) restored `running_bots` as an additive alias, so the break is gone and the strict xfail fires.

This flips the test to assert the restoration: `running_bots` present AND the 0.12 `running`/`count` fields coexist (the alias is additive). A future regression of #579 fails it with a clear message.

The other V010-BREAK xfails (admin envelopes, stop status code, webhook echo) are genuinely still-broken shapes untouched by the batch — left as-is.

All other v0.12.7 validate legs are green; images published. The promote-time `release-validate` dispatch runs from main and picks this up.

### DIFF
--- a/deploy/compose/tests/compat/v010_rest_compat_test.py
+++ b/deploy/compose/tests/compat/v010_rest_compat_test.py
@@ -188,19 +188,20 @@ def test_04_bot_request_v010_flow(stack):
     print(f"\n[compat/bots] POST /bots (plain 0.10 body) → 201 · meeting {body['id']} live ({m['status']})")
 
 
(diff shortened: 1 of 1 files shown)


DmitriyG228:
docs: none — test-marker-only change (stale strict-xfail → plain assertion); no user-facing docs surface.

Issue #675, opened 2026-07-16
runtime meeting-bot profile command /app/vexa-bot/entrypoint.sh does not exist in the shipped bot image → every k8s bot spawn StartErrors (works by accident only on docker Cmd-append semantics)


DmitriyG228:
### Why this matters

The runtime kernel spawns a meeting bot from the `meeting-bot` **profile**, whose `Runnable.command` is hard-coded to `["/app/vexa-bot/entrypoint.sh"]` (`profiles.py:118`). But the shipped `vexaai/vexa-bot:v0.12.7` image has **no such path** — its `ENTRYPOINT` is `["/app/entrypoint.sh"]`, `WorkingDir=/app`, and there is no `/app/vexa-bot/` directory at all (docker-inspect verified; corroborated by the image's own `Dockerfile` below).

How this stays hidden on compose/lite but detonates on k8s is a **backend-semantics split**:

- **docker backend** (`docker_backend.py:255-256`) passes the profile command as the container **`Cmd`**. Docker semantics **append `Cmd` as arguments to the image `ENTRYPOINT`** → the container actually runs `/app/entrypoint.sh /app/vexa-bot/entrypoint.sh`. The extra arg is ignored by the entrypoint script, so **the bot works by accident**.
- **k8s backend** (`k8s_backend.py:89-90`) passes it as `kubectl run --command -- /app/vexa-bot/entrypoint.sh`. The `--command` flag **REPLACES the image entrypoint** (sets `Pod.spec.containers[].command`) → the container tries to exec a path that doesn't exist → **`StartError`, exit 128, `stat /app/vexa-bot/entrypoint.sh: no such file or directory`**.

So on **every stock helm install, every k8s bot spawn StartErrors** and the meeting never produces a word. This was **witnessed live today on the v0.12.7 helm witness** (LKE staging, ns `vexa-v012-staging`): pod `vexa-mtg-7-fd038627` StartError with exactly that message. Every subsequent spawn in the session had to be **hand-recreated with `command=["/app/entrypoint.sh"]`**, after which the bot joined and transcribed fine. Same family as #673 — **the k8s spawn path was never exercised end-to-end before ship.**

### Value this issue delivers

> On a stock helm install, the runtime spawns a real meeting bot Pod that **starts** (no `StartError`), joins the call, and streams transcript — instead of every spawn dying at container start on a path that doesn't exist in the shipped image.

### Where we are (honest)

[...]

PR #683, opened 2026-07-16
fix(release-validate): un-break YAML — #675 conformance snippet as one-liner

DmitriyG228:
PR #681's image-identity step embedded a multi-line `python -c` heredoc at column 1 inside a `run: |` block — YAML terminates the block scalar there (`could not find expected ':'` line 229). Since release-images **calls release-validate as a reusable workflow**, the parse failure blocked the entire v0.12.8 build: the tag push died as a 0-second failure and `workflow_dispatch` 422s.

Fix: the registry-dump snippet becomes a semicolon-chained one-liner (verified: YAML parses, `bash -n` clean, `ast.parse` clean). No behavioral change to the gate.

Unblocks the v0.12.8 candidate build. Part of the #675 batch follow-through.


DmitriyG228:
docs: none — CI workflow syntax repair only.

Issue #673, opened 2026-07-16
k8s spawn backend gives spawned bot/agent Pods NO tolerations/nodeSelector — every spawn Pending forever on an all-tainted cluster, meeting silently fails


DmitriyG228:
### Why this matters

The k8s spawn backend runs every workload — meeting bot, agent, worker — as a bare Pod via `kubectl run`, and builds that Pod's spec from exactly five inputs: image, `restart=Never`, adoption labels, env, and an `--overrides` blob that carries **volume mounts only**. There is no `nodeSelector`, no `tolerations`, no `affinity`, and no env-driven way to inject any of them.

The helm chart's `global.nodeSelector` / `global.tolerations` / `global.affinity` are applied to the long-lived **Deployments** (the runtime kernel Pod itself schedules onto a tainted pool just fine). But a `kubectl run` Pod is not a Deployment child — it inherits none of that. So on any cluster where **every** node is tainted — the standard pool-segregation pattern (dedicated node pools, spot/temp pools, GPU pools) — every bot and agent the runtime spawns sits `Pending` forever with `untolerated taint(s)`, and the meeting silently fails. No error surfaces to the user; the runtime believes it dispatched the workload.

This was **witnessed live on LKE today during the v0.12.7 release witness**. Our staging cluster carries `vexa.ai/pool=main|temp:NoSchedule` on every node. Evidence:

- `vexa-mtg-6-3d6ca2c2` — `FailedScheduling: 0/4 nodes are available: 4 node(s) had untolerated taint(s)` (a meeting bot).
- `vexa-agent-2-chat-main` — a **17-day-old** `Pending` Pod in the same namespace (an agent). Every spawned workload class is affected, and they accumulate as zombie Pending Pods.

The witness could only unblock the meeting by **manually untainting a node** — there is no supported knob. Same family as #656 / #650 / #636: **the shipped config doesn't provide what the runtime needs, and the failure mode is silence.**

### Value this issue delivers

> On a cluster whose nodes are all tainted (dedicated / spot / GPU pools), a spawned meeting bot or agent schedules and runs — by default onto the same pool the runtime kernel itself runs on — instead of sitting `Pending` forever while the meeting silently fails.

### Where we are (honest)

Anchors verified against `origin/main` (identical to tag `v0.12.7` for every file below).

[...]

PR #684, opened 2026-07-16
fix(runtime): scheduling-only pod overrides must not replace the containers list

DmitriyG228:
**Found by the v0.12.8 stock helm witness — the exact class this witness exists for.**

#678 emits `containers: [{name}]` in every override. `kubectl run --overrides` merges the containers **list by replacement** (json-merge, not strategic), wiping the generated container's image/env/command → API server rejects the Pod (`spec.containers[0].image: Required value`) → every spawn dies in <1 s, meeting stuck `requested`, no pod ever exists.

Live evidence (staging, stock v0.12.8): workload `mtg-18-03262323` starting→stopped in 0.3 s; server dry-run inside the runtime pod reproduces the rejection with the containers key and passes without it (image survives, tolerations + nodeSelector applied).

Fix: emit `containers` only when volumeMounts force it (the workspace-store seam, pre-existing shape); scheduling stays at pod level. Red→green: `test_k8s_pod_overrides_scheduling_without_store` now asserts `containers` absent (old code fails it). Runtime suite 155 passed.

Requires re-cut (runtime image changed): v0.12.9.

### DIFF
--- a/core/runtime/src/runtime_kernel/k8s_backend.py
+++ b/core/runtime/src/runtime_kernel/k8s_backend.py
@@ -93,10 +93,14 @@ def pod_overrides(env: dict[str, str], *, container_name: str) -> Optional[dict]
     node_selector = _scheduling_json(env, NODE_SELECTOR_ENV, dict)
     if not volumes and not tolerations and not node_selector:
         return None
-    container: dict = {"name": container_name}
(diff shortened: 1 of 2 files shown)


DmitriyG228:
docs: none — behavior fix inside the #673 mechanism; the k8s scheduling docs from #678 remain accurate.

PR #686, opened 2026-07-16
fix(helm-gate): here-strings kill the pipefail/SIGPIPE race that failed v0.12.9 preflight twice

DmitriyG228:
`set -euo pipefail` + `printf '%s\n' "$RENDER" | grep -q …`: on a MATCH, grep -q exits immediately, printf catches SIGPIPE (141), pipefail fails the pipeline — so the check fails **exactly when it should pass**. Deterministic on ubuntu runners once $RENDER_AUTH outgrew the pipe buffer (this batch's env additions did it); masked locally by macOS timing. Failed the v0.12.9 `release-images` preflight twice (`FAIL: CLAUDE_CODE_OAUTH_TOKEN missing from the Secret when set` + `printf: write error: Broken pipe`).

All 6 `printf | grep` sites → here-strings. `gate:helm PASS` locally; no assertion semantics changed.


DmitriyG228:
docs: none — test-harness plumbing fix, no assertion or product change.

Issue #528, opened 2026-07-11
meeting-api + runtime: Redis clients are built bare (no socket timeout, keepalive, or health checks) — a Redis outage leaves hung or zombie clients that only a service restart heals


DmitriyG228:
---

### Why this matters

On 2026-04-21 (hosted, 0.10) a storage failure put Redis into MISCONF for 46 minutes. When Redis
came back, **no service recovered**: every client pool silently kept broken connections — the
dashboard stayed frozen, stop commands never reached bots, webhooks silently dropped — for 25
additional minutes, until every Redis-client service was manually rolling-restarted
(`vexa-platform/docs/incidents/2026-04-21-redis-storage-cascade.md` §5: `health_check_interval`
defaults to 0/off and no service set it; liveness probes don't touch Redis; publish failures
swallowed without invalidating the pool). Five days later the same client gap produced the
04-26 silent consumer hang (a dead socket that never raises without `socket_timeout` — that
incident's §3). The postmortem's verdict: "once a client pool goes zombie, only a process restart
can heal it."

What an operator loses: a transient dependency blip converts into a multi-service outage that
only manual restarts end — in a self-host, that means an operator who doesn't know the runbook
has an indefinitely broken stack after any Redis restart, OOM-kill, or network flap.

### ERA NOTE (mandatory)

The incidents are **hosted-0.10-era**; the 3-line fix they specified was filed and closed as
#267 (COMPLETED) against the 0.10 tree, which is not on `main`. The 0.12 tree is a rebuild.
**This issue validates the NEW code against a KNOWN FAILURE MODE of the OLD code** — and the
inspection finds the new code re-introduced the exact construction the incidents indicted, while
also (in the gateway) carrying the exact fix. Every claim below is grounded at sha `0080471a`.

### Value this issue delivers

> **V1 — Kill and restore Redis under load and the stack heals itself:** on the compose stack,
> after a Redis outage (stop/start AND silent network drop, both) every service returns to green
> with ZERO service restarts — new segments ingest, `/ws` frames resume, bot stop commands and
> webhooks deliver — within a stated recovery bound (≤ socket_timeout + one tick), proven in the
> chaos lane.

> **V2 — It stays that way: a standing gate fails any long-lived Redis client built without the
> hardening,** so the bare-construction pattern that was rebuilt once cannot silently return a
> second time; and the chaos lane's fault range covers the incident's fault classes (refused +
> silent-drop), not only a 3-second pause.

[...]

PR #822, opened 2026-07-19
bot: idempotent memoized redis connect (fix "Socket already opened" race)

### LINKED ISSUE #820: bot: node-redis adapters have a first-use connect race → "Socket already opened"
### What happened

Under a normal bot start, the transcript pipeline intermittently faulted with
`Error: Socket already opened` from the bot's redis adapters. Expected: the lazy connect is
idempotent under concurrent first use. Got: two callers race the connect and node-redis throws on
the second `connect()`.

### Meeting platform

n.a. (all — it's in the bot's redis transport, not platform capture).

### Deployment — observed on / verified against

Self-hosted; code present on `main` (v0.12.14).

### Root cause (code-grounded, `main`)

Both adapters memoize connection with a **boolean flipped only after** `connect()` resolves:

- `core/meetings/services/bot/src/adapters/transcript-redis.ts:85-89`
- `core/meetings/services/bot/src/adapters/acts-redis.ts:85-89`

```ts
let connected = false;
const ensure = async (): Promise<void> => {
  if (!connected) {
    await client.connect();   // ← both racers get here
    connected = true;         // ← set too late
  }
};
```

Check-then-act race: the first `await client.connect()` yields the event loop before
`connected = true` runs, so a second concurrent first-use caller also sees `connected === false` and
calls `client.connect()` again — node-redis v4 throws `Socket already opened` on the second call.
The bot triggers concurrent first use routinely (first transcript `xAdd`/`publish` racing the acts
`subscribe`, or two segments in one tick).

### Why this is not #528

#528 hardened the **Python** long-lived clients (socket_timeout / keepalive / health_check) and
**explicitly scoped these bot node-redis clients out** ("auto-reconnect by library default … out of
the gate's (Python) scope, named for completeness"). That is about **reconnect after an outage**;
this is a different defect — a **first-use connect race** that faults before steady state.

### Fix (candidate)

Memoize the connect **promise** (not a post-hoc boolean) so concurrent first-use callers share one
`connect()`; clear it on failure so a later call can retry; make `quit()` settle any in-flight
connect and gate on `client.isOpen`. Extracting the shared memo into a small `makeLazyConnect()`
helper also removes the duplicated pattern across both adapters and makes it directly unit-testable.
Candidate branch ready with a red→green test (negative control reproduces the double-connect).

### Evidence

- Bot logs: `pipeline fault: Error: Socket already opened` on a self-hosted run.
- Deterministic test: two concurrent `ensure()` callers against a fake client whose `connect()` is a
  deferred promise — old flag-after-await memo calls `connect()` twice (red), memoized helper calls
  it once (green).

a-tokyo:
**Delivers issue:** #820 · Closes #820

Base `main` (a7f404ee, v0.12.14) · head ff765009 · 5 files, +192/-29
(new `redis-lazy-connect.ts` + test; `transcript-redis.ts` / `acts-redis.ts` refactored to use it).

The transcript writer and acts subscriber connect lazily on first use, memoizing with a boolean
flipped only **after** `connect()` resolves — a check-then-act race: two concurrent first-use callers
both see the flag false and both call `connect()`, and node-redis v4 throws `Socket already opened`
on the second (witnessed as a bot pipeline fault). The fix memoizes the connect **promise** so
concurrent callers share one `connect()`, extracted into a shared `makeLazyConnect()` that both
adapters use — removing the duplicated fragile pattern and making it directly unit-testable.

- **C1 — the fault, witnessed.** ran: a self-hosted bot through a live meeting. saw:
  `pipeline fault: Error: Socket already opened` from the redis adapter path. concluded: the lazy
  connect is not idempotent under concurrent first use.
- **C2 — the race, localized.** ran: read `transcript-redis.ts:85-89` and `acts-redis.ts:85-89`.
  saw: `if (!connected) { await client.connect(); connected = true; }` — the flag is set only after
  the awaited connect, so two first-use callers both pass the guard. concluded: classic
  check-then-act; node-redis v4 throws on the second `connect()`.
- **C3 — not #528.** ran: read #528 + its branch. saw: it hardens the Python clients and names these
  node-redis clients as out-of-scope (reconnect-by-default). concluded: distinct, unowned defect — a
  first-use connect race, not outage reconnect.
- **C4 — the fix, proven idempotent.** ran: the new `redis-lazy-connect.test.ts`, plus the existing
  `transcript-redis.test.ts` / `acts-redis.test.ts`. saw: one connect under concurrency; a negative
  control reproduces the old double-connect; all suites green. concluded: the shared memo removes the
  race with no happy-path behavior change.

## Acceptance floor

| Row | Evidence (`npx tsx src/adapters/redis-lazy-connect.test.ts`, head ff765009) |
|---|---|
| A1 — one connect under concurrent first use | 3 concurrent `ensure()` callers against a fake client with a deferred `connect()` → `connect()` called **exactly once** ✅ |
| A2 — negative control (the bug) | same harness drives the old flag-after-await memo → `connect()` called **twice** ✅ (proves the test discriminates the defect) |
| A3 — failed connect retriable | first `connect()` rejects → memo cleared → next `ensure()` reconnects (no permanent wedge) ✅ |
| A4 — clean teardown | `quit()` during an in-flight connect settles it, then `quit()`s only when `isOpen`; a never-connected `quit()` makes no `client.quit()` call ✅ |
| A5 — no regression | `transcript-redis.test.ts` + `acts-redis.test.ts` stay green on head ✅ |

## Docs diff (D6c)

`docs: none` — internals/reliability change with no user-visible surface (the fault it removes was an
internal flake). No changelog fragment; the fast static gates (isolation / exports / graph / schema /
contract-version / config-contract / test-isolation) pass locally on head.

No new deps, no secrets, no external surface. `gitleaks` clean.

## Validation request

Any competent non-author: `pnpm --filter @vexa/bot test` on head (new concurrency test green, adapter
suites green), and — if reproducible in their env — confirm the `Socket already opened` fault no
longer appears across repeated bot starts. Deployment validated: **self-hosted compose** (bot image).

## Authorship

Tooling disclosure (optional): drafted with agent assistance; all claims code-grounded and re-verified by the author.

### DIFF
--- a/core/meetings/services/bot/package.json
+++ b/core/meetings/services/bot/package.json
@@ -9,8 +9,7 @@
     "build": "tsc && node build-browser-utils.mjs",
     "build:browser-utils": "node build-browser-utils.mjs",
     "start": "tsx src/index.ts",
-    "test": "tsx src/config.test.ts && tsx src/orchestrator.test.ts && tsx src/signals.test.ts && tsx src/entrypoint.test.ts && tsx src/join-driver.test.ts && tsx src/stress.test.ts && tsx src/adapters/lifecycle-http.test.ts && tsx src/adapters/transcript-redis.test.ts && tsx src/adapters/acts-redis.test.ts && tsx src/pipeline.test.ts && tsx src/speaker-hints.test.ts && tsx src/capture-bridge.boundary.test.ts && tsx src/live-pipeline.test.ts && tsx src/recording.test.ts && tsx src/telemetry.test.ts && tsx src/telemetry-recorder.test.ts && tsx src/replay.test.ts && tsx src/zoom-speaker-wiring.test.ts && tsx mock/mock.test.ts",
-    "replay": "tsx src/replay.test.ts",
+"test": "tsx src/config.test.ts && tsx src/orchestrator.test.ts && tsx src/signals.test.ts && tsx src/entrypoint.test.ts && tsx src/join-driver.test.ts && tsx src/stress.test.ts && tsx src/adapters/lifecycle-http.test.ts && tsx src/adapters/transcript-redis.test.ts && tsx src/adapters/acts-redis.test.ts && tsx src/adapters/redis-lazy-connect.test.ts && tsx src/pipeline.test.ts && tsx src/speaker-hints.test.ts && tsx src/capture-bridge.boundary.test.ts && tsx src/live-pipeline.test.ts && tsx src/recording.test.ts && tsx src/telemetry.test.ts && tsx src/telemetry-recorder.test.ts && tsx src/replay.test.ts && tsx src/zoom-speaker-wiring.test.ts && tsx mock/mock.test.ts",    "replay": "tsx src/replay.test.ts",
     "check:isolation": "node scripts/check-isolation.js"
   },
   "dependencies": {
--- a/core/meetings/services/bot/src/adapters/acts-redis.ts
+++ b/core/meetings/services/bot/src/adapters/acts-redis.ts
@@ -13,6 +13,7 @@
 import { createClient } from 'redis';
 import { actsChannel, parseAct, type Act } from '../contracts.js';
 import type { ActsSource } from '../ports.js';
+import { makeLazyConnect } from './redis-lazy-connect.js';
 
 /** The minimal subscriber surface the source needs — injected so the adapter is offline-provable.
  *  `subscribe(channel, cb)` delivers each raw message string to `cb`. */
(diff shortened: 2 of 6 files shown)


DmitriyG228:
## Maintainer assist — rebased onto `main`, changelog fragment added. **The branch is yours; force-push freely.**

I have not taken custody of this branch. This was a one-shot unblock so it can be witnessed today — if you want to change anything, just push. (We have a PR elsewhere that stalled for two days because a contributor stood down while a maintainer held their branch; not repeating that here.)

### What I changed

1. **Rebased** `ff765009` → `4ae58e6c` onto `main` (`bf13dc54`). Your commit, your authorship (`Ahmed Tokyo <ahmed.tokyo1@gmail.com>`), unsquashed.
2. **One conflict**, `core/meetings/services/bot/package.json` — the test list. Resolved as a **union**, not either side: your branch predates `telemetry-recorder.test.ts` landing on main, so taking your side would have silently dropped it. Final list is main’s 19 + your `redis-lazy-connect.test.ts` = 20.
3. **Added `docs/changelog.d/820-redis-connect-race.md`** — this was the only real gap. `docs-current` was red because every user-visible change needs a per-PR fragment (never an edit to `changelog.mdx` itself — that file collides across parallel PRs). That is now the fragment convention satisfied.

### Observation bundle — controlled red→green, offline, no redis

Your test is unusually good and I want that on the record: **its first assertion is the negative control.** It asserts the *old* flag-after-await memo calls `connect()` twice, so the test proves both arms on every run and cannot silently stop discriminating.

I also ran the arm that the inline control does not cover — does it catch a *future* break of `makeLazyConnect` itself:

| arm | `makeLazyConnect` | result | exit |
|---|---|---|---|
| 1 | as submitted | **7/7 pass** | `0` |
| 2 | regressed to the naive `connected` boolean | **3 fail** — `connectCalls=3` (expected 1); `quit()` leaves `quitCalls=0` | `1` |
| 3 | restored | 7/7 pass | `0` |

Exit `1` under regression confirms it fails CI rather than just printing red — that distinction matters here, and it holds.

`node scripts/gates.mjs all` → **all 33 gates green** on the rebased head.

### What remains — and it is not code

Per the merge bar, this needs two things neither of which I can supply:
- **`state: value-signed`** — the D9 human sign-off, from a **non-author**. You reported #820 *and* fixed it, so it cannot be you.
- a non-author review approval on the current head sha.

The evidence above is meant to make that signature a five-minute confirmation instead of an investigation.

Thanks for this one — the shared `makeLazyConnect` extraction (rather than patching both adapters separately) and the deterministic race window in the fake client are exactly how we would have wanted it written.

DmitriyG228:
Closing the stale PR record as **delivered in v0.12.16**, not declined. Your change shipped via release commit `fc7b4e5a`; issue #820 is closed with the red→green receipt and contributor credit. Follow-up commit `60555fc0` bounded quit during an unreachable Redis connection. Thank you, @a-tokyo.

DmitriyG228:
Closing the stale PR record as **delivered in v0.12.16**, not declined. Your change shipped via release commit `fc7b4e5a`; issue #820 is closed with the red→green receipt and contributor credit. Follow-up commit `60555fc0` bounded quit during an unreachable Redis connection. Thank you, @a-tokyo.

Issue #797, opened 2026-07-18
speaker hints (Teams+Zoom): name resolution misses on live DOM and the no-name guard discards every detection SILENTLY — signal detection itself works (witnessed v0.12.13)


DmitriyG228:
## Goal

Teams and Zoom must resolve live DOM display names when their producer-specific
signal can support one, and must fail loud with a typed, sanitized
`name-unresolved` observation when identity cannot be resolved. An unresolved
observation never becomes a speaker hint: unknown is preferred to a fabricated
name.

This is the ordered producer companion in
`train/causal-speaker-attribution`, composed after #956. #956 owns durable
`captured-signal.v1` custody, causal binder/publication semantics, and Jitsi.
#797 owns only Teams/Zoom DOM extraction, producer diagnostics, sanitized
producer-DOM traces, and platform-specific validators.

## Where we are — corrected, evidence-backed

The original body said the Teams voice-level signal did not fire. That diagnosis
was retracted after the v0.12.13 witness logs were read hop by hop:

- the live watcher found participant tiles **with**
  `[data-tid="voice-level-stream-outline"]`;
- the observed tile identity printed an empty display name;
- `extractName` returned empty and the no-name guard discarded the detected
  edge before it could cross the bridge;
- Zoom had the same fail-silent class at a different producer seam: an active
  container with an absent/empty footer collapsed to `null`.

Current code anchors on the train branch:

- Teams name extraction and late render retry:
  `core/meetings/modules/teams-capture/src/msteams-speakers.ts:141-176`
- Teams pre-hint unresolved boundary:
  `core/meetings/modules/teams-capture/src/msteams-speakers.ts:235-256`
- Zoom active-container/footer classification:
  `core/meetings/modules/zoom-capture/src/zoom-speakers.ts:171-196`
- Zoom confirmed unresolved latch:
  `core/meetings/modules/zoom-capture/src/zoom-speakers.ts:198-245`
- sanitized page→Node validator/counters:
  `core/meetings/services/bot/src/capture-bridge.ts:182-228`
- required bridge registration and distinct Teams/Zoom wiring:
  `core/meetings/services/bot/src/capture-bridge.ts:512-665`

## Producer contracts — do not collapse them

| Platform | Signal | Multiplicity | Timing / edge contract | Name-risk surface |
|---|---|---:|---|---|
[...]

PR #831, opened 2026-07-19
feat(bot): record-always at the capture boundary — captured-signal.v1 recorder, hint capture, STT tap, distiller (#830)

### LINKED ISSUE #830: record-always at the capture boundary: persist captured-signal.v1 (audio + speaker hints) + STT round-trips per session, distill to offline replay fixtures
## Where we are (honest, code-grounded)

Transcription/attribution failures are undiagnosable retroactively because prod keeps only pipeline **outputs** (transcript segments): #807 (48% zero-segment meetings), #613/#616 (onset loss), #499/#539 (drop/repeat), #797 (silent hint discard). The diagnostic seam already exists but is unwired: `TelemetrySink` port + `makeTelemetryTap` tee (`core/meetings/services/bot/src/ports.ts`, `capture-bridge.ts`) with **no production adapter** and no `startCaptureBridge` telemetry argument at the composition root. The deterministic replay twin exists (`replay.test.ts` over `eval/replay-fixture/session.captured-signal.jsonl`).

## Prepared solution

Tee the raw capture stream at the boundary into a per-session `captured-signal.v1` JSONL (audio frames + — mixed lane — the out-of-band speaker hints as `type:"hint"` records), plus every STT round-trip beside it (`<session>.stt.jsonl`). Gate on a new `captureSignalEnabled` invocation.v1 field (or `VEXA_CAPTURE_SIGNAL=1`). A distiller (`eval/src/distill.mjs`) cuts a recorded session down to a minimal fixture by time window / speaker (mixed lane: hint-derived windows); `replay.test.ts` accepts `REPLAY_FIXTURE=<any session>` and runs the universal checks. Off ⇒ capture path byte-for-byte unchanged (single branch in the tap).

## Acceptance table

| # | Observation | Oracle |
|---|---|---|
| 1 | Tap→recorder→file: header + frames schema-conformant, seq monotone, PCM Float32-bit-exact | telemetry-recorder.test.ts (offline) |
| 2 | Mixed-lane hints recorded interleaved, epoch-clocked, pipeline still receives every hint | telemetry-recorder.test.ts §1b (offline) |
| 3 | Zero-frame session leaves attributable header-only file; writer faults never reach capture | telemetry-recorder.test.ts (offline) |
| 4 | A REAL meeting session records audio+hints and replays deterministically offline | live leg + `REPLAY_FIXTURE` replay |
| 5 | Distiller cuts full/per-speaker fixtures that replay green | distill.mjs on the live session |
| 6 | Negative control: toggle off ⇒ zero files, pipeline unchanged | test suite + gates |
| 7 | STT tap logs each round-trip (latency, status, typed fault) | live leg stt.jsonl |

## Human bar / setup
Operator run (TTS/scripted bots — no human speakers needed) · Compose or single VM.

Follow-ups deliberately out of scope: S3 SignalWriter + TTL lifecycle + pinning (Phase 1/2 of the plan), auto-flag wiring to `flagged-issue.v1`.

DmitriyG228:
Closes #830.

## The observation bundle

Rows map to #830's acceptance table. Every offline row is a committed test; the live rows were witnessed on a real Jitsi meeting (mixed lane — the same lane as the Teams/Zoom attribution bugs) with two scripted Chromium speakers whose microphones are ground-truth WAVs.

| # | Evidence | Result |
|---|---|---|
| 1 | `tsx src/telemetry-recorder.test.ts` — 50 tapped frames → file; header + every frame ajv-valid vs the SSOT schema; seq monotone; stored PCM byte-compared Float32-exact against the tapped PCM | ✅ |
| 2 | same suite §1b — real `makeSpeakerHintSink` drives 3 hints + 20 frames: hints land as `type:"hint"`, ajv-valid, epoch-clocked, interleaved in arrival order, and the pipeline still receives all 3 | ✅ |
| 3 | same suite — zero-frame session leaves a header-only file; a writer that throws on every append never throws into `captureFrame` | ✅ |
| 4 | **live**: real bot (this branch's `dist`, published image, hosted STT) in a real meeting → `session written: … (216 frames, 78 hints)`, exit 0; harvested and replayed with `REPLAY_FIXTURE=… tsx src/replay.test.ts` → deterministic, every segment transcript.v1-valid | ✅ |
| 5 | **live**: `distill.mjs` full cut 216 frames + 78 hints; `--speaker Boris` → 115 frames + 58 hints from 2 hint-derived windows; `--speaker Anna` → 127 + 20; both replay green | ✅ |
| 6 | toggle off ⇒ tap is one truthiness check (`makeTelemetryTap` short-circuit, asserted in the pre-existing `telemetry.test.ts`); full bot suite + `node scripts/gates.mjs all` green | ✅ |
| 7 | **live**: `stt.jsonl` captured 18 round-trips with latency and typed faults | ✅ |

### What the live leg found (and this PR fixes)

The first live run recorded **162 frames of real audio and zero attribution** while the bot's own counters read `bridge-crossed=28`. In the mixed lane the speaker hints arrive on a *separate* bridge channel that the audio tap never sees, so a recorded Teams/Zoom/Jitsi session could reproduce the sound of a misattribution bug but never the misattribution — useless for exactly #797/#499/#539. The offline test had missed it because it fed hints *through* the tap, which the real lane never does. Fixed in 73af234b (`HintEvent` in captured-signal.v1, `captureHint` port, tee in `makeSpeakerHintSink`); re-run recorded 216 frames + **78 hints** whose timeline matches the scripted Anna→Boris turn-taking. gmeet was never affected (it binds the glow name onto the frame).

### Not covered
No successful live STT text: the test deployment's transcription token is exhausted and 402s every request (a separate finding — see the STT-preflight PR). The STT tap recorded those 402s, which demonstrates the capture-vs-STT bisect working but is not the happy path.

## The diff
`contracts.seal.json` reseal is additive (`captured-signal.v1` gains `HintEvent`; `invocation.v1` gains optional `captureSignalEnabled`) — lane:contract review.

### DIFF
--- a/contracts.seal.json
+++ b/contracts.seal.json
@@ -13,7 +13,7 @@
   "core/gateway/contracts/ws.v1": "77ce41cb384600acaf34402f4daf8d04b32063c662831238ba7a2e4195e80287",
   "core/identity/contracts/identity.v1": "36902fe610115088c855e7af607c1d778dbd3ab75512d8c71c8bf78cf1767dc7",
   "core/meetings/contracts/acts.v1": "1eaa34d27c52b0eb987a278115a1966c058f615ca16e3391b00107a6779dec4f",
-  "core/meetings/contracts/captured-signal.v1": "78db9e8188a27ca88d3a07ad5a8d0ef458e63fa207538ae19429f86d59857332",
+  "core/meetings/contracts/captured-signal.v1": "6ab2c63f5164827d9735f51b0add4d30890972069f18b63d52255df33c36907d",
   "core/meetings/contracts/flagged-issue.v1": "4b868f57298131f8eec8246d86d57251f259927d8bfd7df5a0f6016a10f16211",
   "core/meetings/contracts/invocation.v1": "c7d7f6b95f11da595d0a9e27a429774cdefcc997831cbedff061ebcc2aafbbb8",
   "core/meetings/contracts/lifecycle.v1": "95392746da49c7c61ec81863e6b836af5acb879f979918753fc8fcf92583b820",
--- a/contracts.seal.json
+++ b/contracts.seal.json
@@ -15,7 +15,7 @@
   "core/meetings/contracts/acts.v1": "1eaa34d27c52b0eb987a278115a1966c058f615ca16e3391b00107a6779dec4f",
   "core/meetings/contracts/captured-signal.v1": "78db9e8188a27ca88d3a07ad5a8d0ef458e63fa207538ae19429f86d59857332",
   "core/meetings/contracts/flagged-issue.v1": "4b868f57298131f8eec8246d86d57251f259927d8bfd7df5a0f6016a10f16211",
-  "core/meetings/contracts/invocation.v1": "24ee452ac0ceb58f96422123a2139dc15d5bc1ce54c1a3f4aa45612d8ba77151",
+  "core/meetings/contracts/invocation.v1": "c7d7f6b95f11da595d0a9e27a429774cdefcc997831cbedff061ebcc2aafbbb8",
   "core/meetings/contracts/lifecycle.v1": "95392746da49c7c61ec81863e6b836af5acb879f979918753fc8fcf92583b820",
   "core/meetings/contracts/transcript.v1": "7d07d0fbac77f6562af64ae6c6286051e0e07763f427347d4e7b7735eaa0f9c8",
   "core/meetings/contracts/webhook.v1": "7d5b9d674544cf8ce7bdeda85b5156907be7c4d3e9be359366cc2ab1665bafba",
--- a/core/meetings/contracts/captured-signal.v1/captured-signal.schema.json
+++ b/core/meetings/contracts/captured-signal.v1/captured-signal.schema.json
@@ -2,11 +2,24 @@
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "$id": "https://vexa.ai/schemas/captured-signal.v1",
   "title": "captured-signal.v1",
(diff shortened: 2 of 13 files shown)


DmitriyG228:
Pushed the **mixed-lane attribution oracle** (2eda66bc) and merged `main` in — gates green on the merged state.

New row on the acceptance table, and the one that matters most for #797/#499/#539: a recorded **mixed-lane** session now replays through the real `@vexa/mixed-pipeline` to deterministic, hint-derived speaker names, offline and with no model. `replay-mixed.test.ts` replaces only the two non-deterministic dependencies (STT → a clock-keyed mock; the cut → an injected `BoundarySource`, since production downloads a pyannote model) and asserts *who spoke*: deterministic across two runs, every name traceable to a recorded hint, no segment anonymous, both ground-truth speakers present across the turn change. Wired into the package's `replay` script, so `gate:replay` covers it.

The golden is harvested from the live jitsi run in this PR and distilled to the ~18s Anna→Boris window (58 frames + 9 hints), keeping real frame/hint timing with a synthetic speech-level waveform — the real-audio cut and the synthetic one replay to identical output (`Anna:8.5s`, `Boris:6.3s`), which is what lets it ship at **27 KB gzipped** instead of 1.2 MB.

Two things the harness taught us, both now in the fixture README because they will bite the next person who regenerates it:
- PCM below the capture energy gate replays to **zero** segments (my first synthetic attempt), so the attribution checks are guarded on a non-empty replay — otherwise "every segment is named" greens vacuously on no segments.
- A too-tight leading cut makes the pipeline's short-UI-switch guard correctly reject the opening name as a spurious tile flip, replaying it as `seg_0`. That looked exactly like the reported bug and was a **fixture artifact** — worth knowing, because it means a too-short first turn reproduces the #797 signature for legitimate reasons.

Issue #807, opened 2026-07-19
Silent value failure: 48% of organic 'completed' meetings deliver ZERO transcript segments


DmitriyG228:
## Where we are (hosted production data, 2026-07-19)

Of the last 7 days' organic **completed** meetings (excl. 2 runaway integrations): **only 114 of 220 (52%) have any rows in `transcriptions`**. Of the transcript-less 106: 37 have NULL `end_time` (lifecycle bookkeeping gap), and ~29 ran 10+ minutes — real meetings where the bot was admitted, sat through the meeting, reported `completed`, and produced nothing.

End-to-end, the number a user feels: **~29% of organic meeting attempts result in a transcript** (397 attempts → 220 completed → 114 with transcript).

## The gap

`status='completed'` currently means 'the bot exited cleanly', not 'value was delivered'. A meeting that captures zero segments is indistinguishable from a success in every metric surface. Candidate causes to separate: transcription pipeline drop, bot present but no audio capture, legitimately silent meetings, completion_reason coarseness.

## Prepared direction

- Persist a delivery marker at completion (e.g. `data.segments_captured`) so 'completed-empty' is queryable and alertable.
- Audit the transcription path for the 10-min+ completed-empty cohort (session_uids in `meeting_sessions` — traceable).
- NULL `end_time` on 37 completed meetings is a lifecycle bug of its own — completion must stamp it.

Origin: users → prod → system logs (DB) → OSS issue.

DmitriyG228:
**Marker leg delivered on production** (v0.12.14 hc11, rev 85): meeting 24337 reached terminal and its `data` carries `segments_captured` — the completed-empty class is now queryable in prod. Remaining leg: the transcription-path audit of the completed-empty cohort (a week of marker data will scope it).

DmitriyG228:
**Era correction on the measurements in this issue** (caught by the owner, now O8 in the platform O-book).

The true 0.12-core cutover for organic traffic was **2026-07-19 ~12:12 UTC** — pinned by the `segments_captured` marker (#808), which appears on 8/18 of today's meetings and on **zero** rows before today:

```
day        meetings  has_segments_captured
07-13..18  ~1800     0
07-19        18      8   <- cutover
```

So the **48%→63% zero-segment shares and the 07-13 'regime change' were measured over prior-era hosted-compat rows** and say nothing quantitative about the current core. What DOES hold on the current build:

[...]

Issue #511, opened 2026-07-11
transcriber config: a wrong URL or token passes every check — preflight counts 404 as ok, the wizard test green-passes bad tokens, bot spawn never consults the probe


DmitriyG228:
`same-setup: first-run/config-truth` (batch anchor #502 — validate on the same fresh-clone Lite setup).

### Why this matters

Self-hosters pointing Vexa at an external OpenAI-compatible STT endpoint (#146: OpenAI Whisper
API on Lite via compose) get no verdict on their URL/key until a bot is already in a meeting:
the bot joins, captures, and every audio chunk fails with a per-chunk HTTP error while the
transcript stays empty. The reporter of #146 burned days cycling URL formats and env-var names.
On the current tree the failure is quieter, not louder: setup surfaces test green for
configurations that cannot transcribe, and the one probe that could catch a wrong URL calls a
404 "ok".

### Value this issue delivers

> **(V1)** An operator who configures a wrong STT URL or a rejected key learns it at
> configure/boot time from a named, actionable error — never from an empty transcript
> mid-meeting.
> **(V2)** Any correct OpenAI-compatible endpoint (base URL or full `/v1/audio/transcriptions`
> path) works identically across every consumer — wizard test, boot probe, dictation route, and
> the meeting bot.

(One diff carries both: the same probe/normalization seam produces the loud red and the
consistent green — D5b: one row + preferred validator per value below.)

### Where we are (honest — every claim from the current tree)

- **The #146 env vars no longer exist.** `TRANSCRIBER_URL`, `TRANSCRIBER_API_KEY`,
  `REMOTE_TRANSCRIBER_MODEL`, `SKIP_TRANSCRIPTION_CHECK` have zero hits in the tree. The current
  contract is `TRANSCRIPTION_SERVICE_URL` / `TRANSCRIPTION_SERVICE_TOKEN`
  (`core/meetings/services/meeting-api/src/meeting_api/config.v1.json`, capability `stt`;
  `deploy/lite/entrypoint.sh:41-42`; `deploy/compose/docker-compose.yml:301-302,431-432`).
  #146 contributes symptoms only (D4).
- **A boot-time live probe already exists** (ADR-0026): meeting-api's `stt` capability declares
  an http probe — POST `{url}/v1/audio/transcriptions` with the Bearer token
  (`config.v1.json`, `capabilities.stt.probe`) — run once at boot
  (`meeting_api/__main__.py:54-56` → `preflight()`) and lazily on `/health`
[...]

PR #832, opened 2026-07-19
fix(config): an exhausted STT token probes RED — the preflight transcribes real audio (#511 follow-up)

DmitriyG228:
Stacked on `fix/511-stt-preflight` (#511). Retarget to `main` once that lands.

## Why this is not already covered by #511

#511 regrades 404 → `invalid_endpoint`, but **402 still falls into the `ok: True` bucket**, so the failure below survives it. Live-witnessed on a deployment today: a token that authenticates everywhere and 402s every transcription, with every machine check green and the meeting completing with zero segments (the #807 shape).

**The probe cannot see the fault, structurally.** It POSTs an *empty body*; a metered backend answers that identically whether the credential is funded or worthless. Measured on the live service with the real dead token:

```
OLD empty-body probe -> {'ok': True,  'status': 422}          # greens the dead token
NEW audio round-trip -> {'ok': False, 'status': 402, 'kind': 'exhausted'}
```

**And no balance threshold can see it either.** `balance_minutes` reads **0.0** both for the exhausted token *and* for a billing-exempt service account that transcribes perfectly — which is why the one balance-aware check in the repo (the wizard) had to hardcode the exempt account's email to avoid condemning the working credential. That hardcode is an internal account name living in an OSS repo, and it is the reason the check could never be generalised.

## The change

Ask the real question: `payload: "audio"` posts a ~1s WAV — the same request shape a bot's first chunk makes — and declared `exhausted_statuses: [402]` grade the answer as a typed `exhausted` **config** fault (member of `CONFIG_FAULT_KINDS`, so bot spawn refuses and `/health` demotes, with the consequence in the reason). The wizard's Test runs the identical round-trip via a shared `audio_probe_body`; `/balance` survives only to *name* the account. No account identity is encoded anywhere — which account a deployment's token belongs to is the secret store's knowledge, via the existing `secret_ref`.

## Observation bundle

|---|---|---|
| 1 | Exhausted token: old probe green, new probe red-402 | live 2×2 above, real service |
| 2 | Negative control — zero-balance **exempt** token still green | live: `{'ok': True, 'status': 200}` |
| 3 | 402 → `kind: exhausted` ∈ `CONFIG_FAULT_KINDS` (spawn refuses, /health demotes) | `test_probe_402_is_exhausted_config_fault` (real local HTTP server) |
| 4 | The probe body actually carries a WAV in multipart | `test_probe_sends_real_audio_so_a_metered_backend_can_price_it` (asserts `RIFF` + `probe.wav` in the received body) |
| 5 | Wizard: exhausted red / zero-balance-but-transcribing green — the pair no balance rule can pass | `test_transcription_exhausted_token_fails_loud_despite_valid_auth`, `test_transcription_zero_balance_but_transcribing_token_is_green` |
| 6 | Metered probe does not run hot | `test_probe_declares_a_long_ttl_because_the_round_trip_is_metered` (ttl 60s → 900s) |
| 7 | **Live deployment**: patched preflight into a running meeting-api with the dead token → boot log `config.v1 capability stt: MISCONFIGURED — the token is VALID but cannot pay…`; `/health` stt row `state: misconfigured`, `kind: exhausted`, `status: 402` | witnessed on the test box |
| 8 | Suites | agent config-test **22 pass**, meeting-api config-contract **27 pass**, `gates.mjs all` green |

`internal@vexa.ai` is removed from `config_test.py`, both test fixtures, the compose comment, and `execution-targets.example.json` — whose note additionally claimed the exempt token needed topping up before a live transcript, the exact misreading that put a meterable throwaway token on a test box.

Contract reseal is additive (optional `payload` / `payload_model` / `exhausted_statuses` on the http probe) — lane:contract review.

### DIFF
--- a/contracts.seal.json
+++ b/contracts.seal.json
@@ -21,6 +21,6 @@
   "core/meetings/contracts/webhook.v1": "7d5b9d674544cf8ce7bdeda85b5156907be7c4d3e9be359366cc2ab1665bafba",
   "core/runtime/contracts/runtime.v1": "9d67b9379e0d34c4d715764a467e7f6e6543a5db8177fc284fb4a7418067bc18",
   "core/runtime/contracts/schedule.v1": "6f2c0277d3d186ca2a9333342930f135ea66bfd931135c451bd12b05ff862648",
-  "deploy/contracts/config.v1": "f30ab55cdf847714dd81a3fd64dc6bb1db845c70cf2b7dbc7db5380da895e061",
+  "deploy/contracts/config.v1": "b8c5c1f6344858eddd6509dcbf0a459a3df73866b467011a774245fb3c9ecb17",
   "deploy/contracts/execution-targets.v1": "0c644571b7823024aa2e7ec6a135a55202233155c6c60eef79f1f4ec063e6375"
 }
--- a/core/agent/control_plane/config_preflight.py
+++ b/core/agent/control_plane/config_preflight.py
@@ -149,24 +149,85 @@ def probe_url(base: str, path: str) -> str:
     return base if base.endswith(path) else base + path
 
 
+#: A ~1s 16 kHz mono WAV of a quiet tone — the smallest body that is unambiguously *audio*, so a
+#: metered backend must price it and answer 200 or 402 rather than rejecting it unparsed.
+_PROBE_WAV_SECONDS = 1
+_PROBE_WAV_RATE = 16000
+
+
+def _probe_wav() -> bytes:
+    """Build the probe's WAV in memory (stdlib only — this file is vendored into every service and
+    takes no dependencies)."""
+    import math
+    import struct
+
+    frames = b"".join(
+        struct.pack("<h", int(6000 * math.sin(i * 0.06)))
+        for i in range(_PROBE_WAV_RATE * _PROBE_WAV_SECONDS)
+    )
+    data_len = len(frames)
+    byte_rate = _PROBE_WAV_RATE * 2
+    return (
+        b"RIFF" + struct.pack("<I", 36 + data_len) + b"WAVEfmt "
+        + struct.pack("<IHHIIHH", 16, 1, 1, _PROBE_WAV_RATE, byte_rate, 2, 16)
+        + b"data" + struct.pack("<I", data_len) + frames
+    )
+
+
+def audio_probe_body(model: str = "whisper-1") -> tuple:
+    """The audio round-trip's (content_type, body) — the ONE definition of "ask the STT backend the
+    real question", shared with the terminal's Test button (``core/agent/control_plane/config_test``)
+    exactly as ``probe_url`` shares the URL rule. Both must ask identically, or the wizard greens
+    what the boot refuses."""
+    return _multipart({"model": model, "response_format": "json"}, "probe.wav", _probe_wav())
+
+
+def _multipart(fields: Mapping[str, str], filename: str, payload: bytes) -> tuple:
+    """Encode one file + simple fields as multipart/form-data (urllib has no encoder, and this file
+    takes no dependencies). Returns (content_type, body)."""
+    boundary = "----ConfigV1Probe0dcb1f7a"
+    out = bytearray()
+    for k, v in fields.items():
+        out += (f"--{boundary}\r\nContent-Disposition: form-data; name=\"{k}\"\r\n\r\n{v}\r\n").encode()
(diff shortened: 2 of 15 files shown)


DmitriyG228:
Reviewed and taking this into the v0.12.15 batch (I own main; sequencing #823 → #832 since this branch already contains #823's commit as its parent).

**Two things I verified rather than assumed:**

1. **Probe cost is bounded.** The audio round-trip is a real billable request, and `/health` refreshes on the probe's ttl — at the old 60s that would have been ~1440 transcription requests/pod/day. `ttl_s: 900` + `timeout_s: 15` brings it to ~96/day and matches the docstring's own promise. Good.

2. **The `/balance` finding is the sharpest thing here** — `balance_minutes` reads 0.0 for BOTH an exhausted token and a billing-exempt service account, so grading by it is structurally broken, not just imprecise. That belongs exactly where you put it (execution-targets), and it explains the recurrence.

**One seam defect found and fixed** (pushed as `66312b84`):

The C3 spawn gate bounds how stale a cached verdict may be before it refuses. That bound was hardcoded `60.0` — correct when the probe's ttl was 60s, wrong the moment this PR raises it to 900. `/health` then re-probes every 900s, so a cached verdict is normally minutes old and the 60s bound reads nearly all of them as "no opinion": **the gate goes inert for ~93% of each cache cycle and spawns against a backend already proven unable to transcribe.** Its own tests stayed green because they seed a fresh verdict.

The bound now derives from the declared `ttl_s` (×2 for the refresh gap) — they are the same fact, so a future ttl change can't silently disarm the gate again. Pinned by `test_exhausted_verdict_still_refuses_at_the_probes_own_cache_age`, which seeds a verdict from the middle of the current cache window; it fails against the hardcoded bound and passes against the derived one (negative control executed).

Full meeting-api suite: 758 passed. `gate:config-contract` + `gate:contract-version` green.

Note that `exhausted` joining `CONFIG_FAULT_KINDS` is a deliberate product change worth stating plainly: a deployment whose STT token runs out of balance now gets a typed 503 on `POST /bots` naming the remedy, instead of bots that join and produce nothing. That is the thesis of #511, applied to the billing case.

DmitriyG228:
**Pre-flight against both live environments — the `exhausted` gate will not misfire on upgrade.**

This PR makes `exhausted` (402) a CONFIG fault, and #823's C3 gate refuses `POST /bots` on config faults. Combined, a 402 from the deployment's STT token would 503 every spawn. Before shipping that to prod I checked what the real endpoints actually answer to the probe's own request shape (a ~1s WAV, multipart, deployment token, executed from inside each meeting-api pod):

| environment | audio probe status | body |
|---|---|---|
| **production** | **200** | `{"text":"Thank you.","duration":1.0,"language":"en",…}` |
| **staging** | **200** | same |

Both transcribe. No 402, so no `exhausted` verdict, so no spawn refusals on upgrade.

**Second finding, and it is the more interesting one:** both environments currently report their `stt` capability as

```
{'state': 'configured', 'probe': {'ok': True, 'status': 404}}
```

`ok: True` on a **404** — because both set the URL as the full endpoint (`…:8084/v1/audio/transcriptions`) and the pre-fix probe appends the path again. Probed both shapes from inside the pods:

| request | status |
|---|---|
| blind append (pre-fix) | **404** → reported `ok: True` |
| endsWith rule (#823 `probe_url`) | **422** → `ok: True`, honestly |
| audio round-trip (this PR) | **200** → `ok: True`, and *provably* able to transcribe |

So the ladder does what the issue said it should: 404-lie → honest-reachable → proven-working. The two live deployments were relying on a green that could not have gone red for any URL error, and this batch is what makes their green mean something.

DmitriyG228:
Rebased onto `main` now that #511 has landed (d57d5eab / #823) and **retargeted from `fix/511-stt-preflight` to `main`** — this is no longer stacked.

Re-verified against the as-merged #511, since the point of this PR is that #511 alone does not close the hole: with the merged preflight in place, the 402 path still needed `exhausted_statuses` + the audio payload. After rebase — agent config-test **22 pass**, meeting-api config-contract **27 pass**, `gates.mjs all` green, and all five vendored `config_preflight.py` copies byte-identical to the canonical one.

The live 2×2 in the PR body still stands (it was measured against the real transcription service, not against a branch): exhausted token `{ok: True, status: 422}` under the empty-body probe vs `{ok: False, status: 402, kind: 'exhausted'}` under this one; billing-exempt zero-balance token `{ok: True, status: 200}`.

Issue #552, opened 2026-07-12
terminal: a transcription-backend failure is shown as "stream disconnected; reconnecting" — the UI names the wrong layer (live-witnessed v0.12.1)


DmitriyG228:
### Why this matters

Live-witnessed on a fresh v0.12.1 lite deploy (2026-07-12): the bot's transcription failed and the
Terminal displayed **"Stream error: Meeting stream disconnected; reconnecting"** with a perpetual
**"Waiting for transcript…"**. The actual fault, which the bot logged precisely, was upstream and
unrelated to the stream transport:

```
[TranscriptionClient] stt unavailable (HTTP 503): {"detail":"Remote transcription fallback saturated: concurrency"}
[bot] pipeline fault: TranscriptionError: stt unavailable (HTTP 503)
```

The system HAS the truth (bot pipeline fault: STT backend rejecting) and the UI renders a fiction
about a different layer (the WS stream). A user follows the UI and debugs their network/connection;
the real problem is the transcription service. **The error surface points at the wrong layer.**
This is the live confirmation of the class #533 raised.

### Value this issue delivers

> When transcription fails, the Terminal names the actual fault ("transcription backend unavailable
> — HTTP 503", surfaced from the meeting's fault state) instead of a generic stream-reconnect
> message, so a user debugs the layer that is actually broken.

### Where we are (honest)

- The bot emits `pipeline fault: TranscriptionError` (bot workload log). Open question the fix must
  answer: does that fault reach the meeting record / stream channel, or die in the bot? (In the
  live repro the meeting stayed `status: active` with a healthy status_transition trail and no
  fault field in `meetings.data` — so the fault likely does NOT propagate to meeting-api; the
  terminal only sees the stream drop. If so, this is BOTH a propagation gap (#527 family) AND a
  render-site gap (#533).)
- Terminal render site: `clients/terminal` meeting stream view (the "Stream error / Waiting for
  transcript" surface). Ties to #533's 45 render sites.

### Acceptance table

| # | Observation | Negative control | Anchor |
|---|---|---|---|
| A1 | Induce an STT 503 (point TRANSCRIPTION_SERVICE_URL at a returning-503 endpoint); the Terminal shows a transcription-backend error, not "stream disconnected" | at base: same induced 503 → "Stream error: disconnected; reconnecting" + eternal "Waiting for transcript" | screenshots both shas, the induced-503 recipe |
[...]

PR #836, opened 2026-07-19
fix(bot,meeting-api): a meeting degraded by STT reports WHY, instead of completing silently (#552 partial, #807 class)

DmitriyG228:
Closes the bot half of the GAP row in `docs/test-scenarios/terminal-seams.md:147` (`stt_402_payment_required: surfaced_to_terminal — GAP: bot pipeline onError is console-only (index.ts) — dies before any carrier`). Related: #552, #807.

## The gap

The faults were never lost for lack of typing. `@vexa/transcribe-whisper` classifies every one (`payment_required` / `unauthorized` / `unavailable` / `timeout`, with the backend's own detail and the right retry semantics), and both pipelines hand them to `onError`. What was missing is the **last hop**: `bot/src/index.ts` logged them to a console that dies with the container.

So a meeting whose STT backend refused every chunk reached `completed` **indistinguishable from a silent room** — no reason in the DB, none on the webhook, nothing for a human to read afterwards. That is precisely why empty-transcript reports (the #807 class) could not be diagnosed after the fact. It is not hypothetical: the exhausted-token deployment in #832 produced 18 typed 402s in a 2-minute live run and completed clean and silent.

## The change

The bot accumulates STT faults **per kind** and reports once, on the **terminal** lifecycle event — the last thing anyone hears from a bot, and the only emit that is never an idempotent no-op at the receiver. A dead backend faults on every chunk, so counting and reporting once is both the honest summary and the only shape that cannot flood the control plane; the *first* fault of each kind still logs loudly for anyone tailing the container.

It rides `lifecycle.v1` as an additive field exactly as `infra_fault` already does (`additionalProperties: true` — **no schema edit, no seal change**). meeting-api ingests it into `meeting.data.stt_fault`, so it reaches the DB and the existing `meeting.status_change` webhook **without a new webhook event type**.

## Observation bundle

|---|---|---|
| 1 | 18 faults of one kind collapse to ONE summary carrying the count | `stt-faults.test.ts` |
| 2 | The backend's own words survive (HTTP 402, "Insufficient balance. Available: 0.00 minutes") — not a paraphrase | same |
| 3 | Only the terminal event carries it; no non-terminal event does | same |
| 4 | A healthy meeting emits **no** field at all (negative control) | same |
| 5 | A non-STT fault is ignored; `null`/`undefined` never throw | same |
| 6 | A reporter that throws still lets the bot report its terminal state | same |
| 7 | meeting-api persists it to `data.stt_fault` **and** the status_change webhook | `test_degraded_meeting_persists_why_the_transcript_is_empty` |
| 8 | A healthy meeting persists no `stt_fault` key | `test_healthy_meeting_carries_no_stt_fault` |
| 9 | The degraded terminal event still conforms to sealed `lifecycle.v1` | `conforms(terminal, "LifecycleEvent")` in the same test |
| 10 | Suites | bot suite clean · gates green · meeting-api **770 pass** (the 31 `test_calendar_sync` failures are **pre-existing on main**, which has 768 pass) |

## Deliberately not changed

- **The `no_op` gate on same-status re-emits.** Reporting mid-meeting would have meant weakening `machine.py`'s idempotency, which is load-bearing; the terminal event carries the same information without touching it.
- **The terminal's live banner.** #552's UX half — "stream disconnected; reconnecting" naming the transport instead of the backend — is still open: the banner reads an SSE fed by the collector, not by lifecycle, so that leg needs a producer on the stream the SSE forwards. This PR makes the fault *durable and attributable*; it does not yet make it *live on screen*. The GAP row stays in-flight and I've left it that way rather than marking it green.

### DIFF
--- a/core/meetings/services/bot/package.json
+++ b/core/meetings/services/bot/package.json
@@ -9,7 +9,7 @@
     "build": "tsc && node build-browser-utils.mjs",
     "build:browser-utils": "node build-browser-utils.mjs",
     "start": "tsx src/index.ts",
-    "test": "tsx src/config.test.ts && tsx src/orchestrator.test.ts && tsx src/signals.test.ts && tsx src/entrypoint.test.ts && tsx src/join-driver.test.ts && tsx src/stress.test.ts && tsx src/adapters/lifecycle-http.test.ts && tsx src/adapters/transcript-redis.test.ts && tsx src/adapters/acts-redis.test.ts && tsx src/pipeline.test.ts && tsx src/speaker-hints.test.ts && tsx src/capture-bridge.boundary.test.ts && tsx src/live-pipeline.test.ts && tsx src/recording.test.ts && tsx src/telemetry.test.ts && tsx src/telemetry-recorder.test.ts && tsx src/replay.test.ts && tsx src/zoom-speaker-wiring.test.ts && tsx mock/mock.test.ts",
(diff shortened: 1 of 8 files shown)


DmitriyG228:
v0.12.16 retirement receipt (2026-07-24): merged release PR #867’s recorded commit list contains this PR’s exact head `99d467e90989b00c67a7ea076fab9a1e8490a6e5`; #867 then shipped as squash merge `e617d8395b162df60fdc4bc5754bc1a887bd1bda` and tag `v0.12.16`. The producer-side work is therefore delivered through the batch. The remaining UI consumer work is tracked separately in #552 and was deliberately moved to v0.12.20. Closing this stale contributor PR is record reconciliation, not rejection of the delivered change.

Issue #539, opened 2026-07-12
mixed lane (Zoom): transcript drops and repeats never bounded — the Zoom leg of the counting oracle and the live verbatim check


DmitriyG228:
### Why this matters

Live mixed-lane transcripts were reported fragmented — gaps and repeated segments — side-by-side
against a clean Google Meet run on the same transcription model. Zoom shares the mixed-lane
engine with Teams (`services/bot/src/pipeline.ts:203-212` — one `createMixedBotPipeline` for
both), so the engine-side truth is established once; what Zoom needs is its OWN proof: the
Zoom-shaped fixture variant and a Zoom live leg a Zoom user can witness and sign.

**Per-platform scope (D5b platform rule — the perceiver bounds the atom):** this is the **Zoom**
sibling of the Teams issue #499. #499 carries the **shared engine work** — the 1..500 counting
oracle, the exactly-once golden, the per-stage attribution harness, and the stable segment
identity from publish to store. THIS issue references those and validates **only Zoom**: the
Zoom fixture variant + the Zoom live leg. A Zoom user will not debug Teams; a validator signs
only the platform they sit on.

**Same-setup batch (`same-setup: zoom-lane`):** #515 (zoom bot: in-meeting audio/consent
dialogs never driven — the overlay-driver), #518 (human-help path: sealed `/b/{token}` unserved
+ the noVNC address chain), and #538 (bot speaker wiring (Zoom) — this issue's twin, published
together). One debug rig serves all of them; claiming neighbors together is recommended.

### Value this issue delivers
> What is said in a Zoom meeting appears in the bot's transcript exactly once, in order — the
> shared oracle's guarantee, witnessed on Zoom by a Zoom validator in a real meeting.

### Where we are (honest)

**The era note (mandatory):** the drop/repeat reports predate v0.12 and describe the previous
engine — **we are validating new code against a known failure mode of the old code.** At head,
the 0.12 L4 baseline's live Zoom bot run on a real public meeting recorded `dup=0` over 39
segments (`core/meetings/eval/BASELINE.md:15`) — encouraging, but not an oracle: no truth file
bounded DROPS, and the run was not scripted. The counting oracle (built by #499) is what turns
"looked fine" into "proven exactly-once".

- The engine components (fixture builder, golden, per-stage runner, ingest upsert) are #499's
  C1–C4 — not duplicated here. If #499 has merged, this issue's floor is the two rows below; if
[...]

Issue #499, opened 2026-07-11
mixed lane (Teams): transcripts drop and repeat sentences — exactly-once has never been proven by an oracle


DmitriyG228:
### Why this matters

On Teams, live bot transcripts have come back fragmented, with gaps and with repeated segments —
observed side-by-side against a clean Google Meet run on the same transcription model, which
excludes the model as the cause. And on the previous engine, one Teams class failed absolutely:
a **two-party call** (bot + exactly one human) produced **zero transcript output for the entire
session** while recording worked (#481) — the worst possible drop, discovered only after the
meeting was over.

Silent loss and duplication are correctness failures of the record itself: a transcript that
drops or repeats sentences cannot be relied on for minutes, decisions, or compliance.

The validation approach is deterministic: scripted speakers count 1..500 through the real
engine. Every number must appear exactly once, in order — a missing number is a drop, a repeated
number a duplicate, an out-of-order number a segmentation fault, each attributed to its exact
stage. The oracle runs offline, in CI.

**Per-platform scope (D5b platform rule — the perceiver bounds the atom):** this is the **Teams**
atom and carries the shared engine work (the oracle, the golden, the identity fix). The **Zoom
sibling** — *"mixed lane (Zoom): transcript quality never placed under the counting oracle"* —
references this issue for the engine change and validates only the Zoom fixture variant + a Zoom
live leg (`same-setup`).

### Value this issue delivers

> **V1 — What is said in a Teams meeting appears in the bot's transcript exactly once, in order
> — proven by a deterministic counting oracle that runs in CI, and felt in a real meeting.**

> **V2 — A Teams call with the bot and exactly ONE human produces a transcript — the 1:1-layout
> class that silently produced nothing on the previous engine (#481) is proven transcribing on
> the current one.**

### Where we are (honest)

**The era note (mandatory):** the drop/repeat reports and #481 predate v0.12 and describe the
previous engine — **we are validating new code against known failure modes of the old code.**
Specifics:
- #481's logs (`[Unified] Scanned 1 participants… no voice-level-stream-outline signal found`)
  are 0.10-engine strings with zero occurrences in today's tree. The 0.12 engine transcribes the
  WebRTC-hook **mixed** stream regardless of any speaker-tile DOM signal
  (`services/bot/src/capture-bridge.ts:129-133`, `:207-241`;
  `services/bot/src/pipeline.ts:179-180`), so the gate that silenced two-party calls is removed
[...]

PR #837, opened 2026-07-19
test(bot): mixed-lane attribution oracle (re-land, dropped by #831's squash) + a latency instrument that measures something real

DmitriyG228:
## 1. The mixed-lane attribution oracle was reviewed but never landed

`replay-mixed.test.ts` and its golden are **absent from `main`**: #831's squash-merge (`2662ce25`) captured the PR *before* I pushed `2eda66bc`, so the single most on-goal piece of that PR silently didn't ship. Cherry-picked verbatim and re-verified green here. It's now in both the package test chain and the `replay` script, so **`gate:replay` covers the mixed lane** — the one where #797 / #499 / #539 actually live, and which had no offline oracle at all.

Worth a maintainer glance at merge strategy: a squash that races a push drops commits without any signal.

## 2. A latency instrument — and why the obvious one is worthless

"Great latency" was an unchecked claim. My first attempt wired a lag oracle into the **batch** replay and it produced `p50=3900ms / max=7800ms`. Those are **pure artifact**: exactly `session_end − segment_end`, because the harness feeds every frame in a tight loop while the pipeline's submit/confirm cadence is **wall-clock** (`speaker-streams.ts`: `setInterval(submitInterval*1000)`, `Date.now()` idle checks) — so nothing can confirm until all the audio is already in. **That number is not reported and not gated.** What the batch harnesses now assert instead is the invariant that *is* meaningful there: nothing may be confirmed before its own audio exists.

The real measurement needs the session fed at the rate it was spoken, so `replay-paced.test.ts` does exactly that — opt-in (`npm run replay:paced`, ~10s for the golden), wall-clock and therefore machine-sensitive, **gating nothing** unless `LATENCY_BUDGET_P95_MS` is passed.

Its causality check earned its keep on the first run: it read **−145ms**, because I paced frames to their *start* timestamp while a capture frame is delivered when its audio *ends*. Fixed, and the check stays as the guard.

### The number, and why I am not calling it our latency

It reports **~55ms p50** on the golden — with a **250ms** mock STT round-trip. That is not physically coherent for audio that must be transcribed before it can be confirmed, so the segment's reported `end` evidently tracks the confirmed *prefix's* audio timeline rather than the last chunk actually sent to STT. **That semantic needs pinning before any budget is declared**, and it's written into the harness header rather than papered over. Please don't quote this number anywhere user-facing yet.

So: this PR delivers the *instrument* and a truthful account of what it does and doesn't measure. It does not yet deliver a validated latency budget for gmeet/teams/zoom — that needs the `end` semantic resolved, plus paced runs per lane.

Gates green; bot suite clean.


DmitriyG228:
Pushed 9eae320b — the open question in the PR body is **resolved**, and latency is now a gated number for gmeet.

**The `end` semantic, pinned.** `speaker-streams.ts` computes a segment's `end` as `windowStartMs + totalSamples/sampleRate` — the buffer's *full extent at confirm time*, so it tracks the newest audio fed rather than the span the text covers. Any latency measured against it collapses to a constant (~55ms) no matter how slow STT is, which is exactly what the first version reported. The metric is now taken against the segment **start**: turn start → confirmed text, the delay a user actually experiences.

**Validated as an instrument, not asserted.** Mock STT 100ms → p50 **503ms** · 250ms → **856ms** · 600ms → **1509ms**. That is ≈ **2×STT + ~300ms**, because LocalAgreement `confirmThreshold: 2` requires two submissions — and two STT round-trips — before a prefix confirms. So *the confirm threshold, not buffering, is the dominant latency lever*: dropping it to 1 would roughly halve time-to-text at a cost in stability. A coherence check now fails the run if time-to-text ever drops below the STT round-trip it must pay, so the metric cannot silently degenerate again.

**gmeet: gated at p95 ≤ 1500ms** (observed 856ms — a regression guard with headroom, not an aspiration).

**Mixed lane (zoom/teams/jitsi): measured, deliberately NOT gated**, and the reason is the interesting part:
- It confirms per **turn**, so confirmed-text timing is bounded by how long a speaker holds the floor (6.8–11.3s on the golden), not by latency. First visible **draft** is this lane's comparable number.
- This harness injects a cut source that fires only on **speaker change**, while production's `PyannoteSegmenter` cuts continuously on speech/silence — so the pipeline is starved of the cuts that trigger early drafts (Anna's first turn produces no draft at all until it ends). The numbers are an **upper bound produced by an unrepresentative double**, and gating on that would be a green that proves nothing.

A real mixed-lane budget needs a boundary source with production cut density. The clean way to get one: record pyannote's boundaries into `captured-signal.v1` during a live run, then replay them — which the recorder in #831 is already the right place to do it from. Happy to take that as a follow-up.

Gates green (two intermediate `gate:compose` failures were flaky under parallel docker load — it passes standalone and in a clean full run).

DmitriyG228:
Pushed 8bced4ba — **the mixed lane is no longer approximated. It is measured, and gated.**

The blocker I flagged last comment ("a real mixed-lane budget needs a boundary source with production cut density") is closed by doing exactly that: `captured-signal.v1` gains a type-tagged `BoundaryEvent`, and the bot tees **production's own `PyannoteSegmenter` cuts** into the recorded session by wrapping the real segmenter through the existing `makeSegmenter` seam — production keeps its cut source, the recorder only observes it (`kind` + `tMs` + `confidence`, same epoch clock as frames and hints). Additive; sessions without cuts still replay, and both harnesses print which chunking they're using.

**Live-harvested on the lab: 216 frames, 76 hints, 27 real pyannote cuts.** With production's chunking:

| | gmeet | mixed (zoom/teams/jitsi) |
|---|---|---|
| attribution | — | both speakers named across the turn change, deterministic |
| time-to-first-text p50 | **856ms** | **2780ms** |
| max | 858ms | 4406ms |
| budget (regression guard) | 1500ms ✅ | 6000ms ✅ |

Same paced clock, same 250ms mock STT.

**The ~3× gap is real and it's a product question, not a test threshold.** The mixed lane waits for a pyannote cut, which needs a *pause*; gmeet submits per speaker every 100ms. Budgets are per-lane guards with headroom, and the harness prints the comparison on every run so the gap stays visible instead of normalized. Combined with the earlier finding that time-to-text ≈ 2×STT + 300ms (confirmThreshold=2), the levers are now explicit: STT round-trip and confirm threshold for gmeet, cut cadence for mixed.

**Two fixture artifacts found and excluded**, both of which mimic the #797 signature and would have produced a false bug report:
- a window starting **mid-turn** cold-starts the pipeline and emits `seg_0` for the opening turn — the full session attributes fine;
- a **too-short leading turn** trips the short-UI-switch guard.

The golden is now cut from the session's natural start, and both traps are written into the fixture README with the instruction to verify any suspicious cut against the undistilled session first.

Contract reseal is additive (`captured-signal.v1` gains `BoundaryEvent`) — lane:contract review. Bot suite clean, gates green.

DmitriyG228:
Superseded: the quality line (`quality/0.12.17-base`, heading to main as the v0.12.17 PR series) contains this PR's entire content — commit `2edc6966` re-lands it and every file probed (latency.ts, pipeline.ts, telemetry.ts, distill.mjs, captured-signal schema) is byte-identical between this branch and that line. Keeping open until the corresponding v0.12.17 PR merges, then this closes without landing separately.

DmitriyG228:
Retriaged out of v0.12.18 to **parked**. Reason: its content is carried by the open quality-line PRs; keep this branch only as a provenance/supersession record until those land. This is an explicit defer, not silent backlog drift.

PR #855, opened 2026-07-20
fix(meeting-api): bound native_meeting_id at intake — 422, not an asyncpg 500 (#843)

### LINKED ISSUE #843: fix(meeting-api): malformed native_meeting_id 500s from the DB layer instead of a typed 422 at intake
## Where we are (honest)

`POST /bots` validates *some* fields at the request boundary and returns a clean typed `422` (missing `platform`, unknown platform, `browser_session`, empty/null body, malformed JSON — all verified good). **`native_meeting_id` is not among them.** It flows unvalidated all the way into the Postgres INSERT and raises there.

Found during storm validation of the deployed **v0.12.15** (`0.12.15-260719-hc13`), staging, authenticated, 2026-07-19.

### Reproduction (both return HTTP 500 after ~5.6s)

```bash
# 1. oversized — platform_specific_id is varchar(255)
curl -X POST "$API/bots" -H "X-API-Key: $KEY" -H 'Content-Type: application/json' \
  -d '{"platform":"google_meet","native_meeting_id":"'"$(python3 -c 'print("A"*20000)')"'"}'
# → 500 Internal Server Error

# 2. NUL byte in the id
python3 -c "import json;print(json.dumps({'platform':'google_meet','native_meeting_id':'日本語'+chr(0)+'nul'}))" \
  | curl -X POST "$API/bots" -H "X-API-Key: $KEY" -H 'Content-Type: application/json' -d @-
# → 500 Internal Server Error
```

Server-side traceback (meeting-api):

```
sqlalchemy.dialects.postgresql.asyncpg.Error:
  <class 'asyncpg.exceptions.StringDataRightTruncationError'>:
  value too long for type character varying(255)
```

### Root cause — point of introduction

`native_meeting_id` → `meetings.platform_specific_id` is `character varying(255)`. Nothing bounds or sanitises the field in the `POST /bots` request model, so the DB is the first thing to object. The rejection happens at the **point of observation** (the INSERT) rather than the **point of introduction** (the request boundary) — the inverse of how `platform` is already handled.

### What is NOT broken (checked, so nobody re-chases it)

- **No SQL injection.** `'; DROP TABLE meetings;--` is stored as a literal string — queries are parameterised.
- **No orphan rows.** The failing INSERT is **atomic**: I verified `0` rows in `meetings` for both crashing inputs. #794's no-orphan guarantee holds *through* this crash.
- Only the caller is affected; no data corruption, no cross-tenant leak (tenant isolation verified 403 separately).

## Sub-finding — no format validation either

`native_meeting_id` accepts arbitrary junk and it is interpolated into `constructed_meeting_url`:

```
native_meeting_id: "'; DROP TABLE meetings;--"
→ constructed_meeting_url: https://meet.google.com/'; DROP TABLE meetings;--   (HTTP 201)
```

A bot is then dispatched at a nonsense URL and fails to join. Same root fix.

## Prepared solution

Bound + validate `native_meeting_id` in the `POST /bots` request model, at the same layer that already 422s `platform`:
- length ≤ 255 (mirror the column, or validate against the per-platform id shape)
- reject control characters / NUL
- optionally a per-platform format check (Meet `xxx-xxxx-xxx`, numeric Zoom id, …) — this also kills the sub-finding

## Acceptance

| # | Observation | Pass |
|---|---|---|
| A1 | `native_meeting_id` of 20000 chars → **422**, body names the field | red→green vs today's 500 |
| A2 | `native_meeting_id` containing NUL → **422** | red→green vs today's 500 |
| A3 | Negative control: a *valid* id still spawns (HTTP 201) | no green-on-empty |
| A4 | `meetings` row count unchanged for A1/A2 (no orphan) | preserves #794 |
| A5 | Rejection latency well under the ~5.6s the DB round-trip costs today | boundary, not DB |

## Deployment to validate
compose or lite — no live meeting required; this is a pure intake-boundary test.

DmitriyG228:
Closes #843

## Value
A malformed `native_meeting_id` is refused at the request boundary with a typed **422**, instead of travelling the whole spawn path and dying inside the Postgres INSERT as an opaque **500** ~5.6s later.

`meetings.platform_specific_id` is `varchar(255)`. An id longer than that, or one carrying a NUL/control byte, raised `asyncpg.StringDataRightTruncationError` at the INSERT. Every other malformed field on this route (missing platform, unknown platform, `browser_session`, malformed JSON) is already refused at this boundary — `native_meeting_id` was the exception.

Reproduced against **deployed v0.12.15 in production** during storm validation.

## Scope — deliberately narrow, and why
Bounds **length and control bytes only. The id's SHAPE is not validated.**

Production evidence forced this: a **bare-numeric Teams id transcribed a real meeting** (24368, 67 segments) while another id of the *same* shape failed. Shape does not predict success, so a format rule would refuse working meetings while fixing nothing. I had initially argued the opposite in #843 and [retracted it](https://github.com/Vexa-ai/vexa/issues/843#issuecomment-5021585354) when that counter-example appeared; a negative-control test now pins the boundary so the idea cannot creep back in.

## Diff
- `bot_spawn/router.py` — bounds check beside the existing 422s, applied *after* URL-derivation so a derived id is bounded too; `NATIVE_MEETING_ID_MAX_LEN = 255` mirrors the column.
- `tests/test_bot_spawn.py` — three tests.
- `docs/changelog.d/847-native-meeting-id-bounds.md`.

## Acceptance

| # | Observation | Result |
|---|---|---|
| A1 | 20000-char id → **422** naming the 255 limit | ✅ red→green (was **201**, then a 500 at the DB) |
| A2 | NUL-byte id → **422** naming control characters | ✅ red→green (was **201**) |
| A3 | **Negative control**: bare-numeric / hyphenated / 1-char / exactly-255 / spaced ids still **201** | ✅ passes before *and* after — the guard is not a format rule |
| A4 | No row written on refusal | ✅ asserted in A1/A2 |
| A5 | No regression | ✅ **128 passed**, 2 skipped (spawn + lifecycle suites) |

Note on A4: production verified the failing INSERT was already **atomic** (zero orphan rows), so #794's no-orphan guarantee held even through the crash. This PR fixes the *altitude* of the rejection, not a data-integrity bug.

## Deployment to validate
compose or lite — no live meeting needed; a pure intake-boundary test.


DmitriyG228:
## A6 — population backtest: what would this guard have refused in real traffic?

The acceptance table proves the guard fires on *chosen* inputs. It does not prove the
guard is safe — the risk of adding a rejection to the hot intake route is not "the 422
fails to fire", it's "a legitimate id now gets refused". So I backtested the predicate
against the entire production meetings table (read-only).

**n = 23,980 meetings.**

| probe | result | reading |
|---|---|---|
| `max(length(native_meeting_id))` | **62** | the 255 bound is 4× above anything ever seen |
| ids longer than 255 | **0** | the length check would have refused **nothing**, ever |
| ids containing a control character | **17** | the control-byte check *would* have refused these |
| …of those, `status='completed'` | 6 | — |
| …of those, **completed with ≥1 transcript segment** | **0** | **none of them ever produced a transcript** |

**Characterisation of all 17:** every one is a 12-character Google Meet id with a trailing
`\n` (ord 10) appended — two carry `\n\n`. Two accounts, all within a single 2-day span in
**August 2025**; none in the current era (O8). The signature is a shell/copy-paste capture
that kept its newline.

**Verdict:** the guard's blast radius over the full production history is *17 refusals and
0 lost transcripts*. It refuses only inputs that already produced nothing. This is the
negative control (A3) evaluated at population scale rather than on hand-picked strings.

It also inverts the value claim slightly, in the fix's favour: those 17 requests spawned
bots that burned a slot and returned an empty meeting. A 422 naming the control character
is strictly better service than the silent nothing those users actually got.

### What A6 does *not* cover — the remaining leg

A1/A2/A4 are proven with `TestClient(app)` against a **fake repository** — the request never
reaches Postgres. So A4 ("no row written on refusal") is currently asserted against a fake,
and the original 500 was `asyncpg` behaviour at the real DB. Before this ships I still owe
the deployed-artifact leg the issue names (compose/lite): POST over real HTTP to a real
Postgres, assert the 422 **and** an empty `meetings` table, plus a normal id still 201 and
spawning. That is the altitude the value claim lives at (P19).

DmitriyG228:
v0.12.16 retirement receipt (2026-07-24): merged release PR #867’s recorded commit list contains this PR’s exact head `a33a91090f45741645a6ea3fda6209c7a15e2a43`; #867 then shipped as squash merge `e617d8395b162df60fdc4bc5754bc1a887bd1bda` and tag `v0.12.16`. Issue #843 carries the 7-case staging-lite witness. Closing this stale PR as delivered through the release batch.

Issue #545, opened 2026-07-12
meetings bot: lifecycle.v1 seals `left_alone` but nothing emits it — a Meet bot whose humans have left (empty room, or parked with Fireflies/Otter) stays until the 4-hour cap, burning capacity


DmitriyG228:
### Prep revision note (2026-07-17)

Re-prepared after a maintainer ruling: the leave criterion is **silence** (no remote mic audio for the window), not a DOM census. The name-classifier component of the previous prep is removed from the floor and held as salvage for an optional veto adapter (Open question Q3 — resolved). Evidence base and seams below re-verified against the prior prep's anchors.

### D5b split/bundle verdict (stated first)

**One code change, two values → bundle.** The diff is a single silence-based aloneness monitor serving the sealed `left_alone` completion reason. Both the #270 ask (bots-only room) and its degenerate case (empty room) are the SAME condition under this criterion — nobody is speaking — so one mechanism carries both values; each keeps its own acceptance row and validator.

**Platform rule:** validation is **Google Meet only** (the perceiver bounds the atom), but the mechanism is deliberately platform-agnostic — it reads the audio tap, not the DOM — so Teams/Zoom/Jitsi siblings inherit the detector unchanged and validate their own platform's live leg when prepared (`same-setup`).

### Why this matters

Two independent production reports (#270). The second, from @m-tauqueer (self-hosted, building a meeting-memory app): after the humans leave a Google Meet, the Vexa bot and a Fireflies bot sit in the room together until the hard session cap, burning bot capacity on a meeting that ended long ago.

Present-tense on current main: v0.12 has no active-phase aloneness detection at all. A Meet bot whose humans all leave parks until an explicit stop or the max-active backstop (default 4 h, `core/meetings/services/bot/src/index.ts:86`). The `everyoneLeftTimeout` knob users can already set (`invocation.schema.json:14`, default 120 s) is parsed and then silently unserved.

### Value this issue delivers

> **V1 (the #270 ask — bots-only leave):** a Meet bot whose room has produced no human speech for the silence window leaves on its own and the meeting closes `completed(left_alone)` — a room containing only notetaker bots (Fireflies, Otter, …) is silent, so the #270 scenario is covered without identifying any bot by name.

> **V2 (the base case — empty-room leave):** an empty room is trivially silent; the same monitor, same window, same terminal.

[...]

PR #485, opened 2026-07-09
feat(meetings): leave after remote-audio silence (#545)

m-tauqueer:
## What changed

- Added a remote-audio activity tap to both capture lanes. Qualifying remote PCM energy resets the silence window; the bot's own speech never enters this path.
- Added a small silence aloneness source behind an adapter seam. It fires exactly once after the configured window and fails closed when capture is unavailable.
- Wired the verdict through the bot orchestrator to the existing lifecycle terminal `completed(left_alone)`.
- Forwarded the existing public `automatic_leave` fields into invocation.v1. `max_time_left_alone` wins when supplied; otherwise `BOT_ALONE_SILENCE_WINDOW_MS` and then the 10-minute module default apply.
- Added runtime, Compose, Helm, and Lite propagation, a Compose mock value scenario, user-facing documentation, and a changelog fragment.

## Why

Current v0.12 seals `left_alone` but does not emit it. A bot in an empty room, or parked with a silent notetaker such as Fireflies, therefore remains until an explicit stop or the hard cap. This change serves that existing lifecycle value using one platform-agnostic signal: no remote microphone audio for the configured window.

The earlier 0.10 name-classifier approach has been removed from the floor. This rewrite follows the re-prepared design in #545: silence-only now, with the adapter seam available for a future presence veto if production evidence requires one.

## Scope

- No participant-name matching
- No `startup_alone` behavior; that remains a follow-up
- No invocation.v1 or lifecycle.v1 schema changes
- Google Meet is the live validation platform; the detector itself is audio-path based

## Verification

- `pnpm gates` on rebased head `9846e78f`: green
  - 308 README checks
  - architecture: 82 nodes, 52 edges, 11 carriers
  - 24 sealed contracts unchanged
  - 12 Python packages and 18 Node packages green
  - real Compose and contract-conformance gates green
- Isolated Compose `silence-left-alone` scenario: green
- Deterministic audio timelines cover silence, resets, self-speech exclusion, fail-closed capture, exact-once firing, window precedence, and adapter veto
- Orchestrator tests validate schema-legal `completed(left_alone)` and terminal races

### Author live observations (full Compose, `W=20s`)

- **A2 — empty room:** last remote audio to browser leave: 21.253s; meeting-api stored `completed(left_alone)`.
- **A1 — Fireflies remains:** admitted with two real co-participant tiles; remote speech continued for 44.315s (>2xW); last remote audio to browser leave: 21.358s; meeting-api stored `completed(left_alone)`.

[...]

PR #885, opened 2026-07-21
fix(bot): share orchestrator port doubles and fail loud on missing deps (#865)

Ayush7614:
## Summary
- Export shared L2 port doubles (`noopAloneness`, `noopActs`, `noopPipeline`, `controlledAloneness`) from `test-doubles.ts` so every orchestrator construction site stops rediscovering the required `aloneness` seam by crashing.
- `createOrchestrator` now fails loud with `orchestrator: required port '<name>' is missing` instead of a raw `TypeError` on `.onAlone`.
- `orchestrator.test.ts`, `stt-faults.test.ts`, and `stress.test.ts` import the shared doubles (no local redeclares).

Closes #865.

## Observation bundle (acceptance)

| # | Expected | Actual | Verdict |
|---|---|---|---|
| A1 | noopAloneness (+ sibling noops) exported from one place; no test file redeclares them | `src/test-doubles.ts` exports them; `rg` shows only `noopAloneness =` definition there; tests import | green |
| A2 | orchestrator.test.ts + stt-faults.test.ts use shared doubles and pass | both suites pass locally | green |
| A3 | missing required port → error naming the port, not TypeError | new checks: `required port 'aloneness' is missing` | green |
| A4 | gates | local bot suites + pre-push static gates green; full `gates.mjs all` runs in CI | green (desk) |

## Test plan
- [x] `pnpm exec tsx src/orchestrator.test.ts`
- [x] `pnpm exec tsx src/stt-faults.test.ts`
- [x] `pnpm exec tsx src/stress.test.ts`
- [ ] CI gates on this PR

## Notes
- Branched from `rc/0.12.16` (aloneness + stt-faults batch). Not on `main` yet.
- Desk check only — no meeting / infra required (issue human bar).
- Authorship: Ayush7614 only (no agent co-author trailers).

Issue #754, opened 2026-07-17
merge-card-comment: fork PRs get a spurious red 'comment' check on every review event — read-only token 403s the card PATCH


DmitriyG228:
### The gap (field-grounded, PR #630, 2026-07-18)

Every review event on a **fork** PR produces a failed `comment` check run: `.github/workflows/merge-card-comment.yml` triggers on `pull_request_review` (`:15-16`), but GitHub caps the token at read-only for that event on fork PRs — the declared `pull-requests: write` permission cannot raise it. The card-update `PATCH /repos/Vexa-ai/vexa/issues/comments/{id}` returns 403 `Resource not accessible by integration` and the run fails.

Observed on PR #630 head `f4af257e`: run 29597824527 (`pull_request_review`, FAILURE, the 403) vs its `pull_request_target` twin 29597508356 (SUCCESS, card posted). The card itself is fine — the red is pure noise, **and it reads as the contributor's failure**: the author saw red checks on a governance-green PR.

### Why this matters

Fork PRs are the contribution path this repo optimizes for. A guaranteed-red check on every review event trains contributors to ignore red — the exact opposite of what the merge-card exists to teach. Zero-cost credibility leak, fires on 100% of fork-PR reviews.

### Prepared direction (small, pick at claim)

- (A, default) Move the review-event card refresh onto `pull_request_target` (which already has write and already posts the card) and drop `pull_request_review` from the write path — the event can stay as a no-op trigger or be filtered out entirely.
- (B) Keep the trigger but guard the PATCH: detect the fork/read-only case and exit 0 with a log line (honest skip, not a red).
- Either way: the check must never be red for a condition the contributor cannot influence.

### Acceptance

| # | Observation | Negative control | Anchor |
|---|---|---|---|
| A1 | A review event on a fork PR produces no failed check run; the card still updates (via the write-capable path). | Base: run 29597824527's 403 on PR #630. | check-run list on a fork PR after fix |
| A2 | Same-repo PRs unaffected (card posts/updates as today). | — | one same-repo PR exercised |

Docs surface: none (CI-internal). Found during the autonomous TAKE verification of PR #630; evidence in that PR's thread.

DmitriyG228:
[...]

PR #887, opened 2026-07-21
fix(bot): latch mixed-lane remote-audio ready on mix destination (#866)

### LINKED ISSUE #866: left_alone never fires on Zoom/Teams/Jitsi when the bot is alone from the start — the feature's headline case, broken on 3 of 4 platforms
Found while validating #545 / PR #485 for the v0.12.16 release candidate. **Blocks the platform-agnostic claim #545 makes**, and defeats the value it exists to deliver on every mixed-lane platform.

## The gap

Aloneness fails closed until capture reports readiness (`aloneness.ts:67` — no `available`, no `lastRemoteAudioAt` → verdict `unavailable`, never `alone`). That is correct: absent signal must not be read as silence.

But **readiness is latched by the first frame of audio, not by capture being attached**:

- mixed lane (zoom / teams / jitsi) — `capture-bridge.ts:359`, `await w.__vexaRemoteAudioReady?.()` fires from *inside* the `createMixedAudioCapture` callback, i.e. only once the combined stream produces a frame that clears the capture gate
- per-channel lane (google_meet) — `capture-bridge.ts:439`, in the per-speaker path

So on **zoom, teams and jitsi** (`config.ts:36-38` `isMixedLanePlatform`), a bot that joins a room where **nobody ever speaks** never receives a frame → never latches `available` → `left_alone` **can never fire**. It sits to the 4-hour cap.

That is precisely the scenario #545 names in its own title: *"a Meet bot whose humans have left (empty room, or parked with Fireflies/Otter) stays until the 4-hour cap, burning capacity."* An empty room is the case where the capacity saving is largest, and it is the case that cannot work.

Google Meet is unaffected — it takes the per-channel lane and latches at `:439`. So the shipped behaviour is: works on 1 of 4 platforms for the empty-room case, works on all 4 for the everyone-left-after-talking case.

## Why the docs are also wrong

`docs/docs/how-to/send-a-bot.mdx` (via #485) states the silence-based auto-leave rule with **no platform caveat**. A Zoom user reading it will expect an empty-room bot to leave, and it will not.

## Fix direction

Readiness means *"capture is attached and could hear"*, not *"capture has heard something"*. Latch `ready()` when the page-side hook is installed and the stream exists — before any frame arrives — for both lanes. `unavailable()` on teardown/failure already covers the fail-closed direction.

Care needed: latching too early on the mixed lane would start the silence clock before remote tracks are attached, so `ready()` should key on the mixed destination stream existing, not merely on the AudioContext being constructed.

## Acceptance

| # | observation |
|---|---|
| A1 | Zoom bot joins a room where nobody ever speaks → `left_alone` fires after the window (today: never fires) |
| A2 | Same for Teams and Jitsi |
| A3 | Google Meet behaviour unchanged (regression guard) |
| A4 | Capture torn down / never attached still fails closed — `unavailable`, never a spurious `left_alone` |
| A5 | `send-a-bot.mdx` states the rule accurately for every platform, or names the caveat |

## Related

- #545 / PR #485 — the feature this blocks
- #270 (@jbschooley) — "leave meeting if all other users are bots", which #545 delivers
- Sibling defect found in the same review and **already fixed** on `rc/0.12.16`: the aloneness tap re-judged delivered frames by RMS against the capture gate’s peak threshold, so quiet speech counted as silence.

## Human bar / setup

Small group (2-5) per platform for A1/A2 — or a solo join to an empty room, which is the cheaper repro since the case is "nobody speaks".

Ayush7614:
## Summary
- Mixed-lane (Zoom / Teams / Jitsi) now latches remote-audio **ready** when the mix `MediaStreamDestination` is created — i.e. capture is attached and *could* hear — instead of waiting for the first gated PCM frame.
- That unblocks the empty-room / alone-from-start case: previously no remote speech ⇒ no frame ⇒ never `available` ⇒ `left_alone` never fired (bot burned to the 4h cap).
- Docs (`send-a-bot.mdx`) state the silence auto-leave rule for **all four platforms**; changelog fragment added.

Closes #866.

## Observation bundle (acceptance)

|---|---|---|---|
| A1 | Zoom empty-room → `left_alone` after window | **Not yet live-witnessed** — needs solo empty Zoom join | pending live |
| A2 | Same for Teams + Jitsi | **Not yet live-witnessed** | pending live |
| A3 | Google Meet unchanged | gmeet path still latches after `createGmeetCapture.start()` (untouched) | green (code review) |
| A4 | Tear-down / never attached fails closed | Boundary test: `unavailable()` on stop; aloneness suite keeps fail-closed | green (desk) |
| A5 | `send-a-bot.mdx` accurate for every platform | Explicit Meet/Zoom/Teams/Jitsi wording; no caveat | green |

### Desk / offline evidence (done)
- `capture-bridge.boundary.test.ts` (Teams fixture, **zero PCM frames**): `mixed lane latches remote-audio ready without any PCM frames (#866)` ✅
- Same test: teardown calls `unavailable` (#866 A4) ✅
- `aloneness.test.ts` full suite ✅ (silence / reset / fail-closed / exactly-once)

## Test plan
- [x] Offline boundary + aloneness suites
- [ ] Solo empty-room Zoom → `left_alone` (A1)
- [ ] Solo empty-room Teams + Jitsi (A2)
- [ ] Meet regression smoke (A3)
- [ ] CI gates

## Notes
- Branched from `rc/0.12.16` (depends on #545 / #485 already in that rc).
- Live legs are the issue's human bar — please hold merge until A1/A2 are witnessed (or a maintainer co-witnesses).
- Authorship: Ayush7614 only (no agent co-author trailers).

### DIFF
--- a/core/meetings/services/bot/src/capture-bridge.boundary.test.ts
+++ b/core/meetings/services/bot/src/capture-bridge.boundary.test.ts
@@ -93,15 +93,32 @@ async function main(): Promise<void> {
     // page.evaluate-serialized functions; shim it page-side so the REAL bridge code
     // (which ships helper-free via tsc) runs unmodified under the test runner.
     await page.evaluate('globalThis.__name = globalThis.__name || ((t, v) => t);');
-    const stop = await startCaptureBridge(page, inv, pipeline);
+    let readyCalls = 0;
+    let unavailableCalls = 0;
+    const activity = {
+      ready() { readyCalls++; },
+      unavailable() { unavailableCalls++; },
+      observeRemoteEnergy() { /* unused — empty-room latch must not need energy */ },
+      snapshot() { return { available: readyCalls > 0 }; },
+    };
+    const stop = await startCaptureBridge(page, inv, pipeline, undefined, undefined, activity);
+
+    // #866 — mixed lane (Teams here) must latch ready when the mix destination exists, even with
+    // ZERO remote audio frames. Waiting on the first gated PCM frame made empty-room left_alone
+    // unreachable on Zoom/Teams/Jitsi.
+    await sleep(200);
+    check('mixed lane latches remote-audio ready without any PCM frames (#866)', readyCalls === 1,
+      `readyCalls=${readyCalls}`);
 
     // Fixture drives the speaking signal: occlusion class ON (start) → OFF (end).
-    await sleep(600);   // observer attach + initial silent state past the 200ms hysteresis
+    await sleep(400);   // observer attach + initial silent state past the 200ms hysteresis
     await page.evaluate(`document.getElementById('alice-outline').classList.add('vdi-frame-occlusion')`);
     await sleep(900);   // 200ms hysteresis + 300ms debounce + margin
     await page.evaluate(`document.getElementById('alice-outline').classList.remove('vdi-frame-occlusion')`);
     await sleep(900);
     await stop();
+    check('teardown reports unavailable (fail-closed; no spurious alone after stop) (#866 A4)',
+      unavailableCalls >= 1, `unavailableCalls=${unavailableCalls}`);
 
     // ── 3) the assertions: hints crossed with name, epoch clock, order ──
     check('page-side watcher started (hop 1 visible in page logs)', pageLogs.some((l) => l.includes('[TeamsSpeakers]')), JSON.stringify(pageLogs.slice(0, 3)));
--- a/core/meetings/services/bot/src/capture-bridge.ts
+++ b/core/meetings/services/bot/src/capture-bridge.ts
@@ -330,19 +330,30 @@ export async function startCaptureBridge(
   await page.evaluate(async ({ isMixed, isJitsi, isTeams, isZoom, botName }) => {
     const w = (globalThis as any) as Record<string, any>;
     if (isMixed) {
-      // Zoom/Teams: installRemoteAudioHook (installed pre-nav) mirrors each remote WebRTC audio
+      // Zoom/Teams/Jitsi: installRemoteAudioHook (installed pre-nav) mirrors each remote WebRTC audio
       // track into w.__vexaCapturedRemoteAudioStreams. Combine them into ONE live stream (an
       // AudioContext destination), keep connecting late-arriving tracks via a rescan (a participant
       // who speaks later), and feed that single mix to the mixed lane (pyannote re-separates speakers).
+      //
+      // Readiness (#866): `left_alone` fails closed until capture reports ready. Ready means
+      // "the mix destination exists — capture is attached and could hear", NOT "a gated PCM frame
+      // has arrived". An empty room (no remote tracks yet, nobody ever speaks) must still latch
+      // ready so the silence window can fire; waiting on the first frame made that headline case
+      // unreachable on every mixed-lane platform.
       const setupMix = (): void => {
-        const streams = (w.__vexaCapturedRemoteAudioStreams || []) as Array<{ id: string }>;
-        if (!streams.length) return;
         if (!w.__vexaMixCtx) {
           w.__vexaMixCtx = new (globalThis as any).AudioContext({ sampleRate: 16000 });
           w.__vexaMixCtx.resume?.();
           w.__vexaMixDest = w.__vexaMixCtx.createMediaStreamDestination();
           w.__vexaMixSeen = new Set();
+          if (!w.__vexaRemoteAudioReadyLatched) {
+            w.__vexaRemoteAudioReadyLatched = true;
+            void w.__vexaRemoteAudioReady?.();
+            w.logBot?.('[mixed] remote-audio ready (mix destination attached)');
+          }
         }
+        const streams = (w.__vexaCapturedRemoteAudioStreams || []) as Array<{ id: string }>;
+        if (!streams.length) return;
         for (const s of streams) {
           if (!s || w.__vexaMixSeen.has(s.id)) continue;
           try {
@@ -355,8 +366,7 @@ export async function startCaptureBridge(
           w.__vexaMixedCapture = true; // guard re-entry while the async create resolves
           Promise.resolve(w.VexaBrowserUtils.createMixedAudioCapture(w.__vexaMixDest.stream, (pcm: Float32Array) => w.__vexaPerSpeakerAudioData(0, Array.from(pcm))))
             .then((cap: any) => { w.__vexaMixedCapture = cap; return cap?.start?.(); })
-            .then(async () => {
-              await w.__vexaRemoteAudioReady?.();
+            .then(() => {
               w.logBot?.('[mixed] capture started over ' + w.__vexaMixSeen.size + ' stream(s)');
             })
             .catch((e: any) => { w.__vexaMixedCapture = null; w.logBot?.('[mixed] capture start failed: ' + String(e)); });
--- a/core/meetings/services/bot/src/capture-bridge.ts
+++ b/core/meetings/services/bot/src/capture-bridge.ts
@@ -340,17 +340,34 @@ export async function startCaptureBridge(
       // has arrived". An empty room (no remote tracks yet, nobody ever speaks) must still latch
       // ready so the silence window can fire; waiting on the first frame made that headline case
       // unreachable on every mixed-lane platform.
-      const setupMix = (): void => {
-        if (!w.__vexaMixCtx) {
-          w.__vexaMixCtx = new (globalThis as any).AudioContext({ sampleRate: 16000 });
-          w.__vexaMixCtx.resume?.();
-          w.__vexaMixDest = w.__vexaMixCtx.createMediaStreamDestination();
-          w.__vexaMixSeen = new Set();
-          if (!w.__vexaRemoteAudioReadyLatched) {
-            w.__vexaRemoteAudioReadyLatched = true;
-            void w.__vexaRemoteAudioReady?.();
-            w.logBot?.('[mixed] remote-audio ready (mix destination attached)');
+      //
+      // AudioContext construction is best-effort and MUST NOT abort the rest of this block: the
+      // Zoom/Teams/Jitsi WHO watchers install below and are independent of the mix graph. An
+      // in-process L3 shim (zoom-speaker-wiring) has no real AudioContext — throwing here used
+      // to skip createZoomSpeakers entirely and zero the boundary arrivals.
+      const ensureMixDestination = (): boolean => {
+        if (w.__vexaMixDest) return true;
+        try {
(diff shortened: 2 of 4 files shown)


DmitriyG228:
This is the one that stops cores burning to the 4h cap — thank you. Latching remote-audio ready on the mix `MediaStreamDestination` (capture attached, *could* hear) instead of waiting for a first gated PCM frame is the correct fix at the point of introduction: the readiness latch was the producer of the bug, not `left_alone` downstream. `awaiting-evaluation` now. Its acceptance needs a live witness of the headline case — a bot alone-from-start in an empty Zoom/Teams/Jitsi room firing `left_alone` instead of parking — which is the row a non-author signs. I'll route that; the offline red→green on the latch is your bundle's first row. Note for release notes: this is the fix that makes the 0.12.16 `left_alone` claim true beyond Google Meet.

Ayush7614:
Fixed the red `node` / `gates` failure.

**Cause:** latching mix readiness created `AudioContext` eagerly; in the in-process `zoom-speaker-wiring` shim that throws (`AudioContext is not a constructor`), which aborted the page evaluate *before* `createZoomSpeakers` and zeroed boundary arrivals.

**Fix:** best-effort `ensureMixDestination()` with catch — WHO watchers still install; real Chromium still latches ready on MixDest (boundary test green). Local re-check: `zoom-speaker-wiring`, `capture-bridge.boundary`, `aloneness` all pass. Push `dae96f0`.

Live A1/A2 empty-room legs remain for the value sign, as you routed.

Ayush7614:
cc: @DmitriyG228

DmitriyG228:
Good diagnosis — the eager `AudioContext` construction throwing in the in-process shim (`AudioContext is not a constructor`) aborting the evaluate before `createZoomSpeakers` is a real trap, and best-effort `ensureMixDestination()` with a catch is the right shape: WHO watchers still install, real Chromium still latches ready on MixDest. Green now. What remains is the live A1/A2 empty-room witness (alone-from-start → `left_alone` fires) — that's the non-author value-sign row and it's **our** step; your offline red→green is the bundle's first row and your lease holds while we schedule the live leg. This is queued to a human witness.

DmitriyG228:
## ✅ Validation evidence (agent-run, adversarial) — VERDICT: CONFIRMED (OFFLINE legs only; live leg still required)

*Evidence for a human value-signer (D13); a non-author adopts the offline rows as the D9 sign for those rows. The live A1/A2 leg is a separate human witness — see script below.*

**Red→green real:** `capture-bridge.boundary.test.ts` — base impl RED (`readyCalls=0`, bot never `available` = the #866 latch-on-first-frame defect); head GREEN (`readyCalls=1`, latched on mix-destination creation with **zero** PCM frames). A4 fail-closed green (`unavailableCalls>=1` on teardown). 23-suite bot module all green; `zoom-speaker-wiring` green; #452-era concurrency guards green.
**Adversarial — false `left_alone` risk: BOUNDED, by design.** Clock arms at join, but firing still needs a **continuous 10-min silent window** and any single frame resets it. Slow WebRTC negotiation and joined-muted-then-speaks are covered. The only room that leaves is one audio-silent for 10 min straight — the intended #545 semantic (silent notetakers), indistinguishable from empty by the feature's chosen oracle.
**One note for maintainers:** the best-effort `ensureMixDestination()` catch is a mild #698-class quiet path — if `AudioContext` ever failed in real Chromium, `left_alone` reverts to never-firing with only an L4 log. Mitigated because the same failure also kills all transcription (a far louder signal). Worth a one-line follow-up; not a blocker.
**gmeet lane untouched (A3):** diff confined to the `isMixed` block.

### Live-witness script the human runs (A1/A2) — ~2 min/platform
For **Zoom, Teams, Jitsi** (Meet = A3 control):
1. Short window: dispatch with `automatic_leave.max_time_left_alone` ≈ 120000 (or `BOT_ALONE_SILENCE_WINDOW_MS=120000`).
2. Empty meeting, admit the bot, no sound for the window → expect `completed(left_alone)`; grep bot log for `[mixed] remote-audio ready (mix destination attached)` (latched at join) + the silence verdict.
3. **Negative control:** at ~window/2 have a human say one word → clock resets, `left_alone` does NOT fire at the original deadline.
4. **A3 control:** same empty run on Google Meet → still fires (unchanged).

**Basis a signer can adopt:** the offline latch mechanism is proven and the semantic risks are analyzed and bounded; the value row that completes 0.12.16's Meet-only `left_alone` claim awaits the empty-room live witness above.

DmitriyG228:
## Status: offline CONFIRMED, held from the v0.12.17 milestone pending the live leg — here's what finishes it

@Ayush7614 — the offline work here is solid and the adversarial validation reproduced red→green. This is **not** a rejection: it's the one contribution in the batch whose value-sign is gated on a live witness, because of what the change does.

**Why this one needs a live leg (and #885/#891/#894 didn't):** #887 extends the `left_alone` claim from Google-Meet-only to Zoom/Teams/Jitsi by latching remote-audio-ready on the mix destination. The failure mode of a wrong latch is a **false `left_alone`** — a bot leaving a real meeting that is *not* actually empty. That's user-visible and costly, so "offline red→green" isn't enough to attest the value; it has to be witnessed not misfiring on a live stream. Your own PR analysis already bounds the risk (needs a continuous ~10-min silent window; any frame resets) — the live leg confirms that bound holds in a real client.

**What closes it — the empty-room witness already scripted in your PR comment**, per platform (~2 min each via the `max_time_left_alone` override):
1. **Jitsi** (meet.jit.si, no account) — bot joins an empty room → detects alone → leaves within the window. **The primary new-platform proof.**
2. **Zoom** and **Teams** — same, on a room for each.
3. **Negative control** — bot in a room with continuous audio (or a second participant) must **NOT** fire `left_alone`.
4. **Meet regression control** — the existing Meet path still fires correctly (no regression from the latch change).

Attach the run logs (the `left_alone` lifecycle event + timing per leg, and the negative control staying silent) as a comment and this earns its D9 value-sign and joins the v0.12.17 milestone.

**Offer:** the dev session can run the **Jitsi leg + negative control** now (no account needed) and post the evidence here to unblock the primary platform; the Zoom/Teams/Meet legs need meeting rooms. Say the word on the PR or Discord.

(Ignore the red `merge-card` — it's the #754 spurious-fork-token bug, not a real failure. Your code checks are green.)

DmitriyG228:
## Live witness — Jitsi POSITIVE leg GREEN (empty room → left_alone fires)

Ran the empty-room witness from source on an empty Jitsi room (public ffmuc instance; meet.jit.si had a lobby), `aloneness` window override = 90s. **The fix extends `left_alone` to Jitsi — witnessed:**

```
20:48:36  [bot] lifecycle.v1 active            (admitted, no lobby)
20:48:36  [page] [mixed] remote-audio ready (mix destination attached)   ← the #887 latch
          [bot] aloneness: silence adapter enabled (window_ms=90000)
20:50:06  [bot] aloneness: silence verdict (last_remote_audio_at=…, window_ms=90000)
20:50:06  [Jitsi] Leaving meeting (reason: left_alone)
20:50:06  [bot] lifecycle.v1 completed (left_alone)   → worker exit 0
```

Active→left_alone = **90s exactly** (the override window), fired only after the full silent window with `bridge-crossed=0` throughout. On a mixed-lane platform that previously never latched remote-audio-ready and so never detected aloneness, the bot now correctly leaves an empty room. This is the #866 extension working on Jitsi.

### Honestly, what is NOT yet witnessed
- **Live NEGATIVE control** (audio present → must NOT fire): attempted, but the headful audio-source browser stalled the run (resource limits on a live second-browser-in-Jitsi). The offline negative control in this PR (the false-`left_alone`-risk analysis + the latch unit test — any remote frame resets the window) stands as the safety evidence; the **live** negative control still owes a clean run. Since the failure mode of this change is a false `left_alone`, this is the row that most warrants a live re-run before the value-sign.
- **Zoom / Teams / Meet** legs — need a room each (owner).

So: Jitsi positive is live-green; negative control is offline-proven + live-pending; three platforms await rooms. The value-sign should wait on at least the live negative control. (`merge-card` red = the #754 fork-token bug, not real.)

DmitriyG228:
Release-assembly status (v0.12.18): offline-CONFIRMED + Jitsi positive live leg witnessed above. Held from merge pending the remaining live witness rows: **Zoom + Teams empty-room positive**, one **negative control** (occupied room, no false left_alone), and the **Meet regression control** — per the script above (`automatic_leave.max_time_left_alone`≈120000). Owner: these need a human/live run; everything else on this PR is ready. Retargeting to `rc/0.12.18` once the witness lands.

DmitriyG228:
Carried from the closed 0.12.18 milestone into the **v0.12.19 co-build batch**. Code/offline evidence and the Jitsi positive leg are already green. The remaining gate is explicitly human-owned: Zoom + Teams empty-room positives, one occupied-room negative control, and Meet regression. **Maintainer owns arranging/running those rooms**; @Ayush7614 is not blocked on more code and will not be asked to wait vaguely.

DmitriyG228:
Carried from the closed 0.12.18 milestone into the **v0.12.19 co-build batch**. Code/offline evidence and the Jitsi positive leg are already green. The remaining gate is explicitly human-owned: Zoom + Teams empty-room positives, one occupied-room negative control, and Meet regression. **Maintainer owns arranging/running those rooms**; @Ayush7614 is not blocked on more code and will not be asked to wait vaguely.

DmitriyG228:
Milestone coordination paired with #866: this PR is moving from v0.12.19 to v0.12.20 as a parked contributor car, not as an active-candidate expansion. @Ayush7614’s branch custody and credit remain intact; no extra contributor code is requested. @DmitriyG228 owns the remaining platform matrix, current-base compatibility, and non-author witness.

DmitriyG228:
@Ayush7614 — still wanted, and main was deliberately prepared for it. Not closing.

The deaf-leave guard landed as [#1195](https://github.com/Vexa-ai/vexa/pull/1195) (`43489d73`, **released in v0.12.24, in production**). Its commit message names this PR: latching remote-audio ready on mix-destination attach is "the right fix at the point of introduction", but on its own it removes the accidental protection that a never-started capture could never latch ready. Stream presence is the bit that separates the two cases, so #1195 landed first and pins both outcomes with two timelines that simulate your earlier latch directly. The guard touches no line inside the block this PR refactors.

So #866 is still open and this is still the fix for it.

The problem is the branch: it is now 574 files and −63,531 lines against main — it predates the repo restructure and would delete most of the tree. It needs re-cutting as a fresh, small branch off current main against `core/meetings/services/bot/src/aloneness.ts` and the mixed lane's `capture-bridge.ts`. Say the word if you want it and we will point you at the exact lines.

Issue #853, opened 2026-07-20
teams (dom-outline) + gmeet attribution validation — and the decision on #514 speak/TTS for autonomous multi-speaker truth (#847 I6)


DmitriyG228:
_Child of #847. Depends on: I2, and on the DECISION this issue exists to force. File-and-hold until I5 reports._

## Value this change must deliver

> **MS Teams attribution (the `dom-outline` branch Zoom never exercises) and Google Meet attribution (bound at capture, no binder) are both validated — and multi-speaker validation stops costing a scheduled human meeting each time.**

## Two legs, two different problems

**Teams — a branch Zoom does not cover.** `hintKindForPlatform` returns `dom-outline` for teams (voice-outline transitions, plus caption events), where zoom/jitsi use `dom-active`. Different latency and failure modes through the same binder: `ClusterNameBinder`'s own header documents `caption` at ~500–1500ms lag with flicker and absence when captions are off, `dom-outline` at ~200ms lag vanishing when video tiles change. I5's verdict does not transfer.

**gmeet — a regression check, not a fix.** The gmeet lane binds names at **capture** (the glow lights a channel; no binder, no cluster ids). The maintainer attests attribution worked correctly in 0.12.15. So this leg is: prove it still does, cheaply, and record a baseline — not hunt a defect. (#616, the glow re-lighting behind speech onset, is a separate open issue about leading *audio*, not about which name a bound turn carries.)

## The decision this issue forces

Both legs need **multiple speakers saying known text**. Two ways:

- **Human** — colleagues in a real meeting; truth is the maintainer's eyes. Unrepeatable, cannot regression-test, costs a scheduled meeting per iteration.
- **#514 speaking bots** (`state: ready`) — the TTS clip pool **self-identifies** ("Boris here, …", the `en` field `speakers.mjs` documents as what the attribution scorer keys on). The speaker label therefore rides INSIDE the content, so the single-pass reference scores attribution automatically, with no labelling and no human. That also closes framework gap **G1** (the content oracle is speaker-blind).

#514's cost is four hops, honestly stated on that issue: meeting-api has no `/speak` route (404 verified live through the gateway AND direct), the bot has no `speak_audio` dispatch branch, `voiceAgentEnabled` is dropped at spawn so the SpeakController is dead-gated, and **no audio-file playback path exists at all** (only text→TTS).

[...]

Issue #852, opened 2026-07-20
zoom: validate the shared name binder on real multi-speaker meetings — a verdict on #797 for the dom-active path (zoom + jitsi) (#847 I5)


DmitriyG228:
_Child of #847. Depends on: I2 (attribution metrics). Recording can start BEFORE I2 lands._

## Value this change must deliver

> **The shared name binder has a verdict on real multi-speaker meetings: we know how many platform hints it discards, how much speech it leaves unbound, and whether #797 is live on the `dom-active` path — for Zoom and, because the binder is shared, for Jitsi.**

## Why Zoom first

`ClusterNameBinder` is THE binding logic for every mixed platform; the four platform files are thin hint sources feeding it. `hintKindForPlatform` returns `dom-active` for **zoom and jitsi**, `dom-outline` for teams. So a Zoom verdict covers the `dom-active` branch for both zoom and jitsi. **Teams is a different branch and keeps its own leg (I6).**

Zoom is also the cheapest real multi-speaker source available: real meetings exist today, the extension already ships a `https://*.zoom.us/wc/*` content script importing `zoom-speakers`, and no bot, lobby or admission is involved. Requirement: join **from the browser** (Zoom Web), since the adapter reads Zoom Web's DOM.

## Recording is dependency-free — do it now

The desktop tape records the **verbatim** ingest stream, and the Zoom path emits `active-speaker` hint events into it. A meeting recorded today therefore already contains everything attribution scoring will need; scoring is retroactive. The scarce resource is live multi-speaker meetings, so nothing should sit in front of them.

Run: extension → desktop with `VEXA_RECORD_TAPE` + `VEXA_STT_TAP` (both already wired), production lane cadence, real STT.

## Prediction on the record (falsifiable before the meeting)

#797 (`state: prepared`) says of Teams+Zoom: *"name resolution misses on live DOM and the no-name guard discards every detection SILENTLY — signal detection itself works (witnessed v0.12.13)."* If true, expect hints **detected but discarded**: non-zero hint miss rate with a high provisional rate (`seg_N` names). If names bind correctly, #797 is stale on Zoom and this issue says so.

## Acceptance

| # | Observation | Arm |
|---|---|---|
| A1 | ≥2 real Zoom multi-speaker sessions recorded as corpus entries (tape → captured-signal + manifest + baselines) | live |
[...]

PR #896, opened 2026-07-21
fix(msteams): a name the resolver cannot select is recovered structurally — witnessed stale live (#853, #797)

DmitriyG228:
## What & why

Live #853 instrumentation ([the RED witness](https://github.com/Vexa-ai/vexa/issues/853#issuecomment-5036909274), 2026-07-21) caught the #797 red on today's Teams web DOM. Unlike Zoom #852 (where the bot is served **no** names), Teams **does** paint a name pill in every tile — but every `teamsNameSelector` missed. The lead selector `div[class*="___2u340f0"]` is an extension-era minified hash that has since **rotated**; the live name div now carries only atomic hashes like `___12zni01 f1cmbuwj fv6wr3j fz5stix` inside a container `___1504rl1 f1euv43f ftuwxu6` — no `data-tid`, no aria name, matching none of the explicit selectors. `extractName` returned `''`, so the no-name guard discarded every voice-outline transition: perfect on-screen names, an anonymous transcript.

## Fix mechanism

Rather than chase a new hash that will drift again (the exact failure #797 predicted), `extractTeamsSpeakerName` gains a **selector-agnostic structural fallback** that runs **only after** every explicit selector misses: it scans the tile's text-bearing leaf nodes for a name-shaped string — has a letter, isn't a timer, isn't a UI-control/forbidden word. It reads the name the way a human reads the tile, so it survives the next hash roll. The explicit selectors stay as the preferred fast path.

`extractName` is hoisted out of the `createTeamsSpeakers` closure to an exported pure function so it is testable offline; behavior for the live code path is unchanged.

## Observation bundle

|---|---|---|
| RED | Explicit selectors only, reconstructed live-shaped tile → `''` | `extractTeamsSpeakerName(tile, { structuralFallback: false }) === ''` |
| GREEN (offline) | Structural fallback → `"Dmitry Grankin"` | `extractTeamsSpeakerName(tile) === 'Dmitry Grankin'` |
| Precedence | A real selector hit still wins the fast path | legacy `___2u340f0` tile → `'Legacy Name'` with fallback on and off |
| Guard | Timer / UI-control-only tile yields no false name | `05:14` + `Microphone` tile → `''`; lone `00:30` → `''` |

`src/name-drift.test.ts` — 6 checks, wired into the module's `npm test`. Source-swap red→green proven: with the fallback body stubbed out the GREEN row **fails** (`assertion failed: GREEN: the structural fallback recovers "Dmitry Grankin"`); restored, all 6 pass. The fixture is **RECONSTRUCTED** to mirror the live-witnessed class *structure* (`___1504rl1` container / `___12zni01` leaf, atomic hashes only, no name attrs) — it is **not** a captured byte-for-byte DOM dump; what it faithfully reproduces is the property that broke the resolver (a name reachable by traversal but by no selector). Zero-dependency module preserved: hand-rolled DOM shim, no jsdom; `gate:isolation` green.

- Module suite: `npm test` (teams-capture) — teams-capture + blindness + name-drift all green.
- Full gates: `node scripts/gates.mjs all` → **✅ gates green** (all 33 gates).

## Honest scope — RED witness + fix + offline green, NOT the live green

This PR carries the **live RED witness**, the **fix**, and the **offline structural regression** (green). The **live green** — a real Teams meeting where the `dom-outline` hints resolve into **named** transcript rows — is **still pending a clean admit** and is not claimed here. It complements **PR #870 (Teams de-silence)**, whose instrumentation is what made this name-drift diagnosis visible in the first place.

Base = `main`, independent (teams-capture only) — **not** stacked on #797/#870.

Refs #853, #797.


DmitriyG228:
**Live GREEN witnessed** (see [#853 comment](https://github.com/Vexa-ai/vexa/issues/853#issuecomment-5036909274) thread): the fixed build resolved `SPEAKER_START: Dmitry Grankin` ×4 with `bridge-crossed 0→24`, `no-name reports → 0`, on the exact DOM that returned empty pre-fix. This PR now carries RED (live) + fix + offline red→green regression + live GREEN. Remaining open: a captured tile-DOM fixture + a WAV guest leg for binder-matched>0 (needs one clean rejoin; not blocking).

PR #905, opened 2026-07-21
fix(zaki-control): scope the terminal-409 shield to the usage settlement alone

ProjectNuggets:
## The live specimen (found by WP-M9 Phase-4 verification on staging)

`capture-7c420dd4`: `status-failed` landed at the Hub first; the reaper later forced `completed` locally and emitted a terminal `status-…-completed` echo. The Hub answers HTTP 409 `{code: invalid_state, retryable: false}` for that echo forever. B's guard (`if event.terminal: return False`) shielded ALL terminal events from the 409 dead-letter lane, so the echo retried every 5s sweep indefinitely — silently (the retry line logs at INFO under a WARNING-level logger) — at 2,400+ attempts and counting, **wedging erasure** for that capture: `terminal_callbacks_delivered` was waiting on an event the Hub will never accept.

## The fix

The shield exists to protect settlement seconds. A status echo carries none. Narrowed to `usage-` events only; terminal status echoes fall through to the existing local-staleness check (dead-letter iff the local capture is terminal or gone).

## Adversarial review (SURVIVES, executed probes)

- Prefix discriminator exhaustive: exactly two event-id mint sites (`status-{id}-{state}`, `usage-{id}-1`), one outbox writer, no other terminal event kinds.
- Erasure gate cannot open early: production drains the echo BEFORE the settlement (`ORDER BY created_at, event_id`, same-transaction tie → 's' < 'u'); `terminal_callbacks_delivered` requires total==delivered and `finalize_erased_capture` rechecks `pending == 0` transactionally. The new test pins the production order.
- Cross-sweep out-of-order 409: predecessors always sort earlier and re-establish the per-capture bar each sweep; probed with a 503-ing predecessor over two sweeps — the echo is never wrongly dead-lettered.
- Reverting only the guard fails exactly the new test (1 failed, 11 passed).

Known residual (by design): a Hub-final 409 on the **usage** event itself still retries forever — the money shield is intentional; that shape means settlement disagreement and must stay loud. Follow-up filed for the upstream reaper race (failed → completed re-emission).

On merge: image rebuild + staging repin; the wedged row dead-letters itself on the first sweep and erasure for meeting 5 unwedges.


ProjectNuggets:
Opened against the wrong base by tooling — please disregard.

PR #924, opened 2026-07-23
fix(admin-api): honor JSON body scopes on token mint (#922)

### LINKED ISSUE #922: admin-api: POST /admin/users/{id}/tokens silently ignores scopes in the body — mints ['bot'] and fails later with a confusing 403
Found while minting probe tokens during the v0.12.17 staging witness.

**Symptom:** `POST /admin/users/{id}/tokens` with body `{"scopes":["bot","tx"]}` **silently drops the body field** and mints scope `["bot"]`. Scopes must be passed as a query param (`?scopes=bot,tx`).

**Why it costs time:** the mint returns 200 with a usable-looking token; the failure only appears later and elsewhere as `403 {"detail":"Insufficient scope for this endpoint"}` on `/transcripts` and `/meetings` — while the *same* token returns 200 on `/bots/status`. Nothing points at the ignored field.

**Fix:** either honor `scopes` in the body, or reject the request (422) when an unknown/unsupported body field is supplied — a silently-dropped input is the anti-pattern here. Same class as 'the no-name guard discards silently' (#797): make the refusal say why.

Not a deploy defect; staging behaved correctly once the query param was used.

Ayush7614:
## Summary
- `POST /admin/users/{id}/tokens` with `{"scopes":["bot","tx"]}` was silently ignored and minted `["bot"]` — failure only showed up later as a confusing 403 on `/transcripts` / `/meetings`.
- Body now honors `scopes` / `name` / `expires_in` (body wins over query); unknown body fields → `422`; query `?scopes=` / `?scope=` still work.
- Docs + changelog fragment updated. Closes #922.

## Observation bundle

|---|----------|--------|---------|
| 1 | Body `{"scopes":["bot","tx"]}` → mint scopes `{bot,tx}` | `test_mint_scopes_from_json_body` green | pass |
| 2 | Unknown body field → 422 (no silent drop) | `test_mint_unknown_body_field_422` green | pass |
| 3 | Body wins over query when both present | `test_mint_body_scopes_win_over_query` green | pass |
| 4 | Query `?scopes=bot,tx` still works (compat) | `test_golden_identity_flow` green | pass |
| 5 | Invalid scope still 422 | `test_invalid_scope_422` green | pass |

**Evidence:** `cd core/identity/services/admin-api && uv run pytest tests/test_stack_admin_api.py -k 'mint_scopes or mint_unknown or mint_body or invalid_scope or golden_identity' -q` → **5 passed** (testcontainers Postgres).

**Not checked:** live staging mint against a running admin-api (desk-checkable at the seam; suite covers the contract).

## Test plan
- [x] admin-api mint body/query/unknown-field tests green
- [ ] Maintainer desk-check: mint with JSON body on staging/compose admin-api

### DIFF
--- a/core/identity/services/admin-api/src/admin_api/app/main.py
+++ b/core/identity/services/admin-api/src/admin_api/app/main.py
@@ -12,17 +12,19 @@ exercises:
   email, plus webhook_url/secret/events from user.data; rejects expired tokens; bumps
   last_used_at; FAILS CLOSED when INTERNAL_API_SECRET is unset (503) and on a bad secret (403).
 
-  Token mint: scoped {bot,tx,browser}, optional multi-scope `?scopes=bot,tx`, optional expiry
-  `?expires_in=<sec>`; an invalid scope → 422.
+  Token mint: scoped {bot,tx,browser}. Scopes via JSON body `{"scopes":["bot","tx"]}` or
+  query `?scopes=bot,tx` / `?scope=bot` (body wins when present). Optional `name` /
+  `expires_in` in body or query; an invalid scope → 422. A JSON body with unknown fields
+  is refused (422) — never silently dropped (#922).
 """
 import hmac
 import os
 from datetime import datetime, timedelta
 from typing import Dict, List, Optional
 
-from fastapi import Depends, FastAPI, HTTPException, Request, Response, Security, status
+from fastapi import Body, Depends, FastAPI, HTTPException, Query, Request, Response, Security, status
 from fastapi.security import APIKeyHeader
-from pydantic import BaseModel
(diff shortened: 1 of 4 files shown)


DmitriyG228 (approved):
First-class fix at the point of introduction: the mint endpoint owns its contract via a proper `TokenCreate` model with `extra:"forbid"` — a silently-dropped body field now returns a loud 422 instead of minting the wrong scope. Backward-compatible (query params still honored; body wins when both present). Empty-scope guard, three tests covering body/unknown-field/precedence, changelog fragment + authentication.mdx updated. Traced against #922 — satisfies it exactly. Value already signed; approving the diff.

PR #925, opened 2026-07-23
fix(mcp): sessioned GET /mcp starts SSE via ASGI passthrough (#921)

### LINKED ISSUE #921: mcp: GET /mcp with a VALID session id never starts a response — the SSE stream produces nothing and sse-starlette's ping never fires
## Origin

Measured during the #795 fix (PR #920, gateway MCP front door). **This is a different point of introduction from #795** — #795's 503 was manufactured by the 0.10 gateway's buffered client; this one is upstream of the gateway entirely, in the MCP service. Filed separately so the gateway fix isn't credited with resolving it.

## Observation (measured, not inferred)

A `GET /mcp` carrying a **valid `mcp-session-id`** receives **no response head at all**:

- **70 seconds with nothing** — no status line, no headers, no body.
- **Identical direct-to-`mcp:8010` and via the gateway.** The gateway is not implicated: it relays faithfully, and its `downstream_stream_opened` fires.
- The MCP service's **uvicorn logs no access line**. Uvicorn logs at `http.response.start`, so the application never starts the response.

## Where it is introduced

`mcp.server.streamable_http._handle_get_request` builds an `EventSourceResponse` over a stream that produces nothing. sse-starlette's **15s keep-alive ping never appears** either — which is the tell: even an idle-but-healthy SSE stream should emit pings, so the response is never being started, not merely silent.

Suspects, in order: the MCP service's GET/stream handling, or the pinned `mcp` / `sse-starlette` versions.

## Why it matters

The gateway now relays `GET /mcp` correctly (#920), so a sessionless GET returns MCP's own `400` verbatim in ~0.05s. But the *sessioned* GET — the actual long-lived SSE leg an MCP client uses to receive server-pushed messages — never opens. So the streamable-HTTP transport's server→client channel is effectively dead, even with the front door fixed.

## Acceptance sketch

- [ ] A `GET /mcp` with a valid session id starts its response (status + `content-type: text/event-stream` headers emitted) promptly, before any data is pushed.
- [ ] sse-starlette's keep-alive ping is observed on an idle stream.
- [ ] A server-pushed MCP message reaches the client over that stream.
- [ ] Negative control: a sessionless GET still returns MCP's own 400 verbatim (the #920 behaviour, unlaundered).

## Repro

Bring up the compose stack, complete an MCP handshake (`initialize` → `notifications/initialized`), then `GET /mcp` with the returned `mcp-session-id` — both through the gateway and directly at `mcp:8010`. Observe: no response head on either, no uvicorn access line, no 15s ping.

Evidence and the full measurement are in PR #920's body (signed gap section).

Ayush7614:
## Summary
- A `GET /mcp` with a **valid** `mcp-session-id` never started a response (no status/headers/body; no uvicorn access line; no sse-starlette 15s ping). Identical direct-to-mcp and via gateway — point of introduction is the MCP service, not the gateway (#920).
- Root cause: `fastapi-mcp` 0.4's `FastApiHttpSessionManager` buffers the full ASGI response and only returns after `handle_request` completes. A sessioned GET is an open `EventSourceResponse`, so that never happens and `http.response.start` never reaches uvicorn.
- Fix: after `mcp.mount_http()`, install an ASGI passthrough that relays the real `send`. Sessionless GET still returns MCP's own 400. Closes #921.

## Observation bundle

|---|----------|--------|---------|
| 1 | Sessioned GET emits `200` + `content-type: text/event-stream` promptly | Raw ASGI `send` captured `http.response.start` within 0.5s; `test_sessioned_get_starts_sse_headers_promptly` green | pass |
| 2 | Idle stream emits sse-starlette keep-alive ping (~15s) | Body `: ping - <ts>` at ~15s; `test_sessioned_get_idle_ping` green | pass |
| 3 | Sessionless GET still MCP's own 400 (negative control / #920) | `test_sessionless_get_still_400` green | pass |
| 4 | Tool surface unchanged | `tests/test_mcp_surface.py` green | pass |

**Evidence:**
```text
cd core/meetings/services/mcp && PYTHONPATH=src uv run pytest \
  tests/test_mcp_sessioned_sse.py tests/test_mcp_surface.py -q
# → 11 passed in ~18s
```

**Red (before fix):** same handshake → sessioned GET → `TimeoutError` at 5s with no `http.response.start`.

**Not checked:** live compose stack push of a server→client MCP message over the SSE leg (acceptance sketch row 3). Headers + idle ping prove the stream opens; a live push can be signed on compose if wanted.

## Test plan
- [x] In-process ASGI: sessioned headers, idle ping, sessionless 400
- [ ] Optional compose: handshake → sessioned GET via gateway and direct `:8010`

### DIFF
--- a/core/meetings/services/mcp/src/vexa_mcp/app.py
+++ b/core/meetings/services/mcp/src/vexa_mcp/app.py
@@ -3,7 +3,9 @@
 Port of 0.10.6 ``services/mcp/main.py`` reduced to the tools whose REST routes EXIST on
 the v0.12 public API (the gateway — ``core/gateway/services/gateway/src/gateway/app.py``).
 Every tool is a thin FastAPI route; ``FastApiMCP`` derives the MCP tool surface from them
-and mounts the streamable-HTTP MCP transport at ``/mcp``.
+and mounts the streamable-HTTP MCP transport at ``/mcp``. The mount uses an ASGI
+passthrough so a sessioned ``GET /mcp`` can start its SSE response immediately
+(#921) — fastapi-mcp's buffered adapter never completes for an open EventSource.
 
 Auth: the caller's credential (``Authorization: Bearer <key>`` / raw ``Authorization`` /
 ``X-API-Key``) is treated as the Vexa API key and forwarded to the gateway as ``X-API-Key``
@@ -27,6 +29,7 @@ from pydantic import BaseModel, Field, model_validator
 
 from .link_parser import ParseMeetingLinkResponse, parse_meeting_url
 from .prompts import PROMPTS, get_prompt_result
+from .streamable_http import install_streaming_http_transport
 
 _DEFAULT_GATEWAY_URL = "http://gateway:8000"
 
@@ -339,5 +342,8 @@ def create_app(
         return get_prompt_result(name, arguments)
 
     mcp.mount_http()
+    # fastapi-mcp 0.4 buffers the ASGI response; a sessioned GET is an open SSE
+    # stream and never completes that buffer — install a passthrough (#921).
+    install_streaming_http_transport(mcp)
     app.state.mcp = mcp
     return app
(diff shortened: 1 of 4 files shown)


DmitriyG228 (approved):
Non-author review + value-sign (D9): root cause verified — fastapi-mcp 0.4's buffered response adapter can never start an unfinished EventSourceResponse; the ASGI passthrough relays the real send while keeping FastApiMCP for tool derivation. All four #921 acceptance rows green (prompt SSE head, 15s idle ping, sessionless-400 negative control preserving #920, tool surface unchanged). Noting the version-pin fragility of patching _http_transport privates — documented in the module docstring and fails loud, acceptable.

Issue #857, opened 2026-07-20
test(join): DOM selector tests assert against hand-built elements — a guard that stayed green over a broken fix


DmitriyG228:
## Where we are (honest)

`core/meetings/modules/join/src/googlemeet/humanized/humanized.test.ts` guards the locale-agnostic Meet join-button selector against the Hungarian-lobby regression (prod ids 13951, 13952, 14018, 14153). It asserts:

```js
const huJoinBtn: El = {
  tag: "button",
  attrs: { jsname: "Qx7uuf" },          // ← no aria-label
  text: "Kérvényezés a csatlakozásra",
  children: [{ tag: "span", ..., children: [] }],
};
assert(matchesSelector(huJoinBtn, googleJoinButtonSelectors[0]));
```

The element is **hand-built in the test**. The selector matches it, the test passes, and the guard has been green ever since.

**Production says the selector never matches a real Meet lobby.** Across 7 successful joins on 2026-07-20, every single one fell through to the English literal `//button[.//span[text()="Ask to join"]]` (selector #3). The two structural selectors — the ones this test certifies — matched **zero** times. Evidence in #846.

So the fixture is the wrong shape, and the test has been asserting a property of *its own object literal* rather than of Google Meet.

## Why this is worth its own issue

This is not "one selector is wrong" — that is #846. This is: **a DOM-matching test built from an imagined DOM cannot fail when the real DOM disagrees**, and we have more than one of them (`session.test.ts` carries the same style for the lobby/auth-guard selectors).

The failure mode is particularly expensive because it is *confidence-inverting*: the Hungarian incident was diagnosed, a fix was written, a regression test was added, and everyone could reasonably believe the class was closed. It was not — and the green test is precisely why nobody re-checked for a year.

The likely concrete defect it masked: `:not([aria-label])` was added to exclude the lobby 3-dot menu, but a real primary CTA carries an `aria-label` for accessibility — the same file proves Meet buttons have them (`[aria-label*="Turn off camera"]`). The synthetic fixture omits `aria-label`, so the filter looks harmless in the test and excludes the real button in production. *(Hypothesis — it needs the captured DOM in #846 to confirm.)*

## Prepared solution

1. **Capture a real pre-join DOM** (English and one non-English) — the outerHTML of the lobby controls subtree — and commit it as a fixture. Anonymise: it contains no meeting content, only chrome.
[...]

Issue #600, opened 2026-07-14
MS Teams removal monitor false-positive: generic [role="alert"] evicts a legitimately-joined bot


DmitriyG228:
**Lane:** platform/teams · Witnessed live on a real Teams meeting (2026-07-14, during #593 fix verification).

## Symptom
A Teams bot that has **successfully joined and been admitted** is **evicted ~1.5 s later** and the meeting is recorded as `completed(evicted)` — even though nobody removed it. It leaves before capture even starts, so no transcript.

Verbatim bot log (admitted → false removal → graceful leave, no human action):
```
✅ Bot was admitted to the Teams meeting: Leave button confirmed
[bot] lifecycle.v1 active
Starting periodic Teams removal monitoring...
Step 6c: Handling post-join AV-confirmation modal (if shown)...
🚨 Teams removal detected: Found removal indicator "[role=\"alert\"]"    ← ~1.5s after admission
🚨 Teams removal detected from Node.js side. Initiating graceful shutdown...
[leaveMicrosoftTeams] Triggering leave action in browser context...
[bot] lifecycle.v1 completed (evicted)
```
Intermittent: a second bot in the *same* meeting held ≥66 s. It fires when a transient Teams alert/modal is present shortly after admission (here, right after the post-join AV-confirmation modal).

## Root cause (in the code as it is)
`core/meetings/modules/join/src/msteams/removal.ts` `checkForTeamsRemoval()` returns `true` on the **first** `teamsRemovalIndicators` selector that is visible, and the monitor maps that to `signalEnd('evicted')` → `leaveMicrosoftTeams`.

The indicator list mixes reliable, removal-specific **text** with **generic** role/class patterns — `core/meetings/modules/join/src/msteams/selectors.ts:112-117`:
```ts
// Generic error patterns
'[role="alert"]',
'[role="alertdialog"]',
'.error-message',
'.connection-error',
'.meeting-error'
```
`[role="alert"]` matches **any** Teams alert region — toasts ("you're muted", captions, network blips), the post-join AV-confirmation modal, notifications — none of which mean the bot was removed. So a benign transient alert triggers a false removal and self-leave.

## Impact
- A legitimately-joined Teams bot can silently leave within seconds of admission → `completed(evicted)`, no transcript.
- Distinct from #593 (that was `pipeline_start_failed` → `join_failure`; this is the removal monitor → `evicted`). Sibling of the #171 admission-false-positive family.

## Suggested fix
[...]

Issue #377, opened 2026-05-26
Google Meet: `admission_false_positive` bot exit on first join (verification re-check race)


DmitriyG228:
## Summary

On Google Meet, the bot occasionally exits with `reason: "admission_false_positive"` 5–6 seconds into a call, immediately after sending the `ACTIVE` status callback. Reproducible on first dispatch into a fresh meeting; rejoining ~30s later typically succeeds without code changes. The user is left waiting and has to re-issue the bot two or three times before it sticks.

## Symptoms

- Meeting record shows `status=completed`, very short duration between `start_time` and `end_time` (≈5–6s).
- Bot logs contain (in order):
  - `✅ Found Google Meet admission indicator: <selector>`
  - `Bot admitted - ensuring AWAITING_ADMISSION state is processed before sending ACTIVE...`
  - `Verifying bot is still in meeting after ACTIVE callback...`
  - `🚨 Bot is NOT in meeting after ACTIVE callback - false positive detected!`
  - `Initiating graceful shutdown sequence` with `"reason":"admission_false_positive","exit_code":0`
- User confirms the bot **was visibly in the meeting** before it left — i.e. it really did admit, then incorrectly self-diagnosed as not admitted.
- Pattern: first dispatch fails, second dispatch (within seconds) fails, third dispatch sticks for the full meeting. Suggests a cold-cache / page-settle timing issue.

## Root cause hypothesis

In `services/vexa-bot/core/src/platforms/shared/meetingFlow.ts` around line 140:

```ts
log("Bot admitted - ensuring AWAITING_ADMISSION state is processed before sending ACTIVE...");
await new Promise(resolve => setTimeout(resolve, 1000));

try {
  await callStartupCallback(botConfig);

  log("Verifying bot is still in meeting after ACTIVE callback...");
  const stillAdmitted = await strategies.checkAdmissionSilent(page);
  if (!stillAdmitted) {
    log("🚨 Bot is NOT in meeting after ACTIVE callback - false positive detected!");
    await gracefulLeaveFunction(page, 0, "admission_false_positive");
    return;
  }
```

Flow:

1. `waitForGoogleMeetingAdmission` finds an admission indicator → returns `true`.
2. Sleep 1s.
3. Send `ACTIVE` callback.
4. Re-run the **same** selector check via `checkForGoogleAdmissionSilent` → `checkForGoogleAdmissionIndicators` in `services/vexa-bot/core/src/platforms/googlemeet/admission.ts:42-72`.
[...]

Issue #318, opened 2026-05-05
Zoom Web per-speaker pipeline broken on modern Zoom client: bot scrapes <audio>.srcObject DOM elements that no longer exist — 0 transcripts on real meetings (confirmed live mid 11717)


DmitriyG228:
## Symptom

A Zoom Web bot in a multi-participant meeting reports `0 streams` to its per-speaker audio pipeline, regardless of:
- whether audio is successfully joined ("Leave computer audio" present in the Zoom audio menu);
- which output sink Zoom is routed to (`BotSink_<mid>` / `ZoomAudioSink` / `Default` all yield identical 0-stream result);
- whether the bot is observed-as-active by Zoom's UI (DOM-based `SPEAKER_START` / `SPEAKER_END` events fire normally for every participant — the speaker observer works fine).

VAD stays at `0/0` indefinitely. Whisper is never called. No transcripts written. PulseAudio recording works (master.webm captures the mixed audio). End user gets a playable audio file but **the meeting transcript is empty for the entire run**.

## Confirmed live (mid 11717, 2026-05-05)

Real Academy Software Foundation working group meeting (4 named speakers — ILM, Sony Imageworks, plus host). Bot got admitted, audio joined, multi-speaker exchange ongoing for 25+ minutes. Bot logs:

```
17:58:24  [PerSpeaker] Browser-side audio capture started with 0 streams
…  vad=0/0 (checked/rejected) — every 30s telemetry tick, indefinitely
…  🎤 [Zoom Web] SPEAKER_START: James Spadafora (Industrial Light & Magic)
…  🎤 [Zoom Web] SPEAKER_START: Cameron Target
…  🎤 [Zoom Web] SPEAKER_START: Olga Avramenko (Sony Pictures Imageworks)
```

Live experiment at 18:22 UTC: switched Zoom speaker output from `BotSink_11717` → `ZoomAudioSink` via VNC into the bot, watched two subsequent telemetry ticks (30s + 60s post-switch). VAD remained 0/0. **This rules out the bot's own audio-sink routing as the cause.** Zoom Web in this client version does not create `<audio srcObject={MediaStream}>` DOM elements at all.

## Root cause

`services/vexa-bot/core/src/platforms/zoom/web/prepare.ts:217-223` (the diagnostic only — same selector is the basis of `SpeakerStreamManager` initialization):

```typescript
const elements = Array.from(document.querySelectorAll('audio, video'));
const withStreams = elements.filter((el: any) =>
[...]

PR #967, opened 2026-07-25
fix(zoom): rank in-meeting DOM evidence above the lobby-text probe in isAdmitted()

### LINKED ISSUE #966: Zoom Web: admitted bot exits with `admission_false_positive` ~3s after being let in — toolbar auto-hide + generic lobby substrings
# Zoom Web: admitted bot exits with `admission_false_positive` ~3s after being let in — toolbar auto-hide + generic lobby substrings

## Summary

On Zoom Web, a bot that has been genuinely admitted can immediately conclude it is **not** in the meeting and leave. `isAdmitted()` consults the lobby-text probe **before** the DOM fallbacks, and `zoomWaitingRoomTexts` contains substrings as generic as `'Please wait'` and `'waiting room'` — which Zoom also renders *in-meeting*. Combined with Zoom's footer toolbar auto-hiding a few seconds after admission (which defeats the primary `isVisible()` check), a live meeting is abandoned seconds after the host clicks Admit.

This is the **Zoom** instance of the family already fixed on other platforms: #377 and #166 (Google Meet), #171 and #600 (Teams). I could not find an existing Zoom report.

Present in **v0.12.18** at `core/meetings/modules/join/src/zoom/admission.ts`.

## Production evidence

From a self-hosted deployment, 2026-07-24 (times UTC):

```
22:58:38.503  [Zoom Web] Still waiting for admission... 177s elapsed
22:58:40.506  [Zoom Web] Bot admitted — Leave button now visible
22:58:41.506  🔥 UNIFIED CALLBACK: ACTIVE - reason: none
22:58:41.525  Verifying bot is still in meeting after ACTIVE callback...
22:58:43.532  🚨 Bot is NOT in meeting after ACTIVE callback - false positive detected!
22:58:43.533  [Graceful Leave] Initiating graceful shutdown sequence
22:58:43.539  [Zoom Web] Leave button selectors all missed — forcing page navigation
```

The bot was in the meeting for **3 seconds**. Result: `status=failed`, zero transcript segments, and no recording — the recording never started, so there was nothing to recover afterwards.

Note the internal contradiction at `22:58:40` vs `22:58:43`: admission was detected *by* the Leave button, and 3 seconds later "Leave button selectors all missed." That is the toolbar auto-hide.

## Root cause

`isAdmitted()` (v0.12.18, `zoom/admission.ts:73-117`) evaluates in this order:

```ts
const leaveBtn = page.locator(zoomLeaveButtonSelector).first();
if (await leaveBtn.isVisible({ timeout: 500 })) return true;      // 1: defeated by auto-hide

const inWaitingRoom = /* body.innerText includes any zoomWaitingRoomTexts */;
if (inWaitingRoom) return false;                                   // 2: SHORT-CIRCUITS

const meetingApp = page.locator(zoomMeetingAppSelector).first();
if (await meetingApp.isVisible({ timeout: 500 })) return true;     // 3: unreachable when 2 hits

// 4: live <audio> + srcObject ...
```

Two independent problems:

**(a) Ordering.** `zoomWaitingRoomTexts` (`zoom/selectors.ts:46-54`) contains bare `'Please wait'` and `'waiting room'`. Zoom renders these in-meeting — connecting-audio toasts, host-control notices — and can leave lobby copy in the DOM after admission. Step 2 therefore vetoes steps 3 and 4 for a bot that is demonstrably in the call.

**(b) Step 4 is dead code.** The `<audio>` + `srcObject instanceof MediaStream` fallback cannot fire on modern Zoom Web, which routes audio through the Web Audio API — this repo's own **#318** documents exactly that. So admission detection presents as four layers while effectively being two, and the thinness is invisible to a reader.

Both signals that *should* have caught this — the footer controls — are matched with `isVisible()`, which is precisely what auto-hide breaks. They are present in the DOM the entire time.

## Suggested fix

Match in-meeting footer controls on **DOM presence** rather than visibility (presence survives auto-hide, and none of these render in the lobby), and rank them **above** the text probe. Keep the text probe as a veto for the weak `.meeting-app` fallback, since Zoom renders the waiting room inside `.meeting-app` (the meeting_id=36 case in the existing comment). Drop the dead `<audio>` branch.

PR follows, with regression tests that fail against current `main` and pass after.

## Note on the existing test harness

`zoom/admission.test.ts`'s `makePage()` already returns `isVisible: () => false` for every locator — it *is* the auto-hidden-toolbar state — but its fake `document` exposes only `body.innerText`, with no `querySelector`. So the harness could never distinguish "footer absent" from "footer present but hidden", which is the entire bug. The PR extends `makePage` with a declared set of present selectors. This is a small instance of what **#857** describes.

## Environment

- Bug verified by reading **v0.12.18** (`core/meetings/modules/join/src/zoom/admission.ts`).
- The production logs above come from a self-hosted **0.10.6.3.14** deployment, where the same code exists at `services/vexa-bot/core/src/platforms/zoom/web/admission.ts`. The defect and the surrounding logic are unchanged between the two; only the path differs.
- Platform: Zoom Web (default path; `ZOOM_SDK` unset), meeting with waiting room enabled.

mikobits:
# fix(zoom): rank in-meeting DOM evidence above the lobby-text probe in `isAdmitted()`

Fixes #966.

An admitted Zoom Web bot could conclude it was **not** in the meeting and leave ~3 seconds after the host clicked Admit — `admission_false_positive`, zero segments, no recording. This is the Zoom instance of the family already fixed for Google Meet (#377, #166) and Teams (#171, #600).

## What was wrong

`isAdmitted()` ran the `zoomWaitingRoomTexts` innerText probe **before** both DOM fallbacks. That list contains bare `'Please wait'` and `'waiting room'`, which Zoom also renders in-meeting (connecting-audio toasts, host-control notices) and can leave in the DOM after admission.

Meanwhile the two footer signals that should have overruled it were matched with `isVisible()` — and Zoom's footer toolbar **auto-hides** a few seconds after admission. So the only checks that could prove "we are in the meeting" were the ones auto-hide breaks, and a fuzzy text match got the last word.

Separately, the fourth fallback (live `<audio>` with `srcObject instanceof MediaStream`) is dead on modern Zoom Web, which uses the Web Audio API — this repo's **#318**. It made admission detection look four-deep when it was effectively two-deep.

## What this changes

- **`selectors.ts`** — adds `zoomInMeetingMarkers`: footer controls matched on **presence**, not visibility. Deliberately broad and redundant, since Zoom ships unannounced DOM changes and A/B tests them; any one match is sufficient. `.meeting-app` is **deliberately excluded** — Zoom renders the waiting room inside it.
- **`admission.ts`** — single `page.evaluate` snapshot, then: footer markers (presence, and no pre-join form) → admitted; else lobby text → not admitted; else `.meeting-app` without lobby copy or pre-join form → admitted. Removes the dead `<audio>`/`srcObject` branch.
- **`admission.ts`** — widens `checkForZoomAdmissionSilent` from 3 attempts/2s to 6/~7.5s. This check only ever gates a decision to **leave** a meeting we believe we are in, so being slow to conclude "not admitted" is much cheaper than being wrong: a real ejection persists and is caught on a later attempt, whereas a premature `false` reads to the user as the bot abandoning a live call.
- **`admission.test.ts`** — extends `makePage()` with a declared set of present selectors, and adds three cases against the shipped `checkForZoomAdmissionSilent`.

## Invariants deliberately preserved

Two earlier bugs are the reason the original ordering existed. Both are covered by new tests so this fix cannot silently undo them:

- **meeting_id=36** — Zoom renders the waiting room inside `.meeting-app`, so the container alone must not report admitted. Lobby text still vetoes that weak fallback.
- **meeting_id=31** — bot at the "Enter Meeting Info"/passcode screen. The pre-join form outranks everything, including footer markers.

## Test evidence

`npx tsx src/zoom/admission.test.ts`

Against current `main`, the new case fails and the guards hold:

```
PASS  timeout → AdmissionError('lobby_timeout')
PASS  anti-bot wall → AdmissionError('denial')
PASS  host rejection → AdmissionError('denial')
FAIL  footer marker present outranks a generic lobby substring — got false
PASS  lobby copy + .meeting-app but no footer → NOT admitted (no regression of #36)
PASS  pre-join form vetoes footer markers → NOT admitted (no regression of #31)
5 passed, 1 failed
```

After this change:

```
6 passed, 0 failed
```

The failing case reproduces the reported incident exactly: every locator invisible (auto-hidden toolbar), a footer control present in the DOM, and `'Please wait'` on screen from an in-meeting toast.

## Note on the harness

`makePage()` already returned `isVisible: () => false` for every locator — it *is* the auto-hidden-toolbar state — but its fake `document` exposed only `body.innerText`. With no `querySelector`, it could not distinguish "footer absent" from "footer present but hidden", which is the whole bug. Adding a declared present-selector set is a small step toward what **#857** asks for.

## Verification status

The fix has been running in production on a self-hosted deployment since 2026-07-24, where the identical defect exists in the pre-0.12 layout (`services/vexa-bot/core/src/platforms/zoom/web/admission.ts`). The bot now logs `✅ Bot verified to be in meeting after ACTIVE callback` at the point it previously left, and completed a subsequent 48-minute Zoom meeting with 115 transcript segments across two speakers.

I have **not** been able to run this against a live Zoom meeting on the 0.12 tree itself — the port is a faithful translation of a fix verified on 0.10.6.3.14, plus the unit tests above. Worth a maintainer smoke-test on a real waiting-room meeting before release.

### DIFF
--- a/core/meetings/modules/join/src/zoom/admission.test.ts
+++ b/core/meetings/modules/join/src/zoom/admission.test.ts
@@ -14,7 +14,7 @@
  * Run: npx tsx src/zoom/admission.test.ts
  */
 
-import { waitForZoomMeetingAdmission } from "./admission";
+import { waitForZoomMeetingAdmission, checkForZoomAdmissionSilent } from "./admission";
 import { AdmissionError } from "../shared/admission";
 import { resetEscalation } from "../shared/escalation";
 
@@ -25,8 +25,19 @@ function check(name: string, ok: boolean, detail?: string) {
   else { console.log(`  \x1b[31mFAIL\x1b[0m  ${name}${detail ? ` — ${detail}` : ""}`); failed++; }
 }
 
-/** A page whose DOM is just body text: evaluate() runs the probe in-node, locators see nothing. */
-function makePage(bodyText: () => string): any {
+/**
+ * A page whose DOM is body text plus a declared set of present selectors.
+ *
+ * `locator().isVisible()` always returns false — which is not a limitation but
+ * the single most important property of this harness: it reproduces Zoom's
+ * auto-hidden footer toolbar, the state in which every visibility probe lies
+ * about a bot that is genuinely in the meeting.
+ *
+ * `present` lists selectors for which `document.querySelector` returns an
+ * element. Exact-string matching is enough here because isAdmitted() queries
+ * the selector constants verbatim.
+ */
+function makePage(bodyText: () => string, present: string[] = []): any {
   const locator = (sel: string): any => ({
     first: () => locator(sel),
     isVisible: async () => false,
@@ -35,7 +46,11 @@ function makePage(bodyText: () => string): any {
   });
   return {
     evaluate: async (fn: any, arg?: any) => {
-      (globalThis as any).document = { body: { innerText: bodyText() } };
+      (globalThis as any).document = {
+        body: { innerText: bodyText() },
+        querySelector: (sel: string) => (present.includes(sel) ? ({} as any) : null),
+        querySelectorAll: (sel: string) => (present.includes(sel) ? [{} as any] : []),
+      };
       try { return fn(arg); } finally { delete (globalThis as any).document; }
     },
     locator,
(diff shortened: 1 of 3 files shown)


DmitriyG228 (changes requested):
Maintainer routing verdict at exact head `2eafbc953b26360b7634c9a32d47a10f1be1a6b2`: changes requested, with contributor custody preserved.

Please keep this PR as the implementation car and make one bounded repair:

1. Build the pre-join veto from the shared pre-join selector contract, including the documented classic `#inputname` and `#joinBtn` shapes, and retain the text-only pre-join guard. Add red→green cases showing broad in-meeting markers cannot admit either classic or text-only pre-join pages.
2. Rewrite the historical “former fallback/removal” comment as the designed-present invariant only.
3. Add `docs/changelog.d/967-zoom-admission-false-positive.md` so the product fix rides the changelog.

Then run the focused @vexa/join tests plus static gates and request a fresh review on the new exact head. Do not add speaker-attribution or browser-product scope; PR #965 and PR #957 are file-disjoint and remain independently owned.

The current merge-card red is expected before value sign-off/approval. The separate comment-job 403 is the known fork-permission issue #754 and is not yours to repair here. Maintainers will arrange the exact-candidate live Zoom waiting-room witness after source green.

PR #985, opened 2026-07-28
Feat/dac v1

sharmaritvik2007:


DmitriyG228:
@sharmaritvik2007 — sorry this sat. Not closing it, but it needs a decision from you.

None of the four features is in main: attendance timeline (`core/meetings/services/bot/src/attendance.ts`), terminal white-label (`brand.ts`, `BrandLogo.tsx`), video recording (`RECORDING_VIDEO_ENABLED`), recording playback (`RecordingPlayer.tsx`). All absent.

The branch is based on v0.12.16 (2026-07-22). Since then main added ~1,400 lines to exactly the files this rewrites — `capture-bridge.ts` alone is +965. It is `CONFLICTING` and a rebase is substantial.

Two notes: the HTTP `Range`/206 substrate behind the playback work already landed independently (`cc4badd3`), so #913's contribution there is the terminal player and proxy allow-list, not the API. And your changelog fragments are numbered `910`–`913`, which collide with unrelated merged PRs — they should be renamed to this PR's number.

Tell us which of the four you still want to land and we will scope a rebase around it. Splitting them would help — white-label and attendance are independent.

PR #999, opened 2026-08-04
fix(bot): point the fake camera at a decodable y4m — kills Meet's permanent "Camera not found" toast

### LINKED ISSUE #998: Meet: permanent "Camera not found" toast over the meeting stage — fake camera source is /dev/null (undecodable)
### What happened

Every Google Meet meeting our self-hosted bot joins shows a permanent **"Camera not found / Make sure your camera is plugged in"** toast, bottom-center over the meeting stage. It never times out — observed still up 11 minutes into a real meeting. Expected: no error toast (the bot joins with its camera off, so nothing should be complaining about a camera at all).

Besides looking broken to every human participant, the toast sits exactly where shared-screen content renders, so any downstream consumer of the bot's display (screenshots, recordings of the stage) gets it burned into the pixels.

### Root cause

Both Chromium launch-arg lists pass `--use-file-for-fake-video-capture=/dev/null`:

- `core/meetings/modules/join/src/browser-args.ts:67`
- `core/meetings/modules/remote-browser/src/args.ts:41`
- (third copy in `core/meetings/modules/join/scripts/debug-join.ts:59`)

Chromium registers a fake camera from that flag, but `/dev/null` is not a decodable y4m, so the device exists and permanently fails to produce frames — which Meet reports as "Camera not found", forever.

### Meeting platform (if relevant)

Google Meet

### Deployment — observed on which version / verified against which tag?

Self-hosted compose, `v0.12.18`. Re-verified present on `main` @ `1f6898cf` (all three sites unchanged).

### Logs, screenshots, meeting shape

Ordinary 2–4 person Meet calls, bot admitted from the lobby, camera off. The toast is present from join and never clears; confirmed via periodic captures of the bot's X display across the meeting.

### Fix, tested

Pointing the flag at a real one-frame black y4m (generated by `entrypoint.sh` via the `ffmpeg` already in the bot image) removes the toast completely — verified live on the same deployment. One frame is enough: the bot never turns its camera on, so the frame is never published.

Deliberately **not** switching to `--use-fake-device-for-media-stream`: that flag also fakes the microphone, which would displace the `tts_sink` → `virtual_mic` PulseAudio chain the speak path depends on.

PR with the fix + regression assertions follows shortly.

sashazykov:
Fixes #998.

### What

`--use-file-for-fake-video-capture=/dev/null` registers a fake camera Chromium can never decode frames from, so Google Meet shows a permanent **"Camera not found / Make sure your camera is plugged in"** toast bottom-center over the meeting stage — for the whole call (observed still up 11 minutes in), overlapping shared-screen content. All three flag sites are fixed:

- `join/src/browser-args.ts` and `remote-browser/src/args.ts` now point at `/tmp/blank-camera.y4m`, generated as a one-frame black 640×360 y4m by the bot `entrypoint.sh` via the `ffmpeg` already in the image. The bot turns its camera off at join, so the frame is never published — it only has to *exist and decode*.
- `join/scripts/debug-join.ts` (the hot debug container, which has no ffmpeg) writes the same frame directly with Node `fs`. This is load-bearing, not cosmetic: the debug runner also consumes `getJoinBrowserArgs()`, which now expects the file.

### Why not `--use-fake-device-for-media-stream`

It also fakes the microphone, which would displace the `tts_sink` → `virtual_mic` PulseAudio chain the speak path depends on. Keeping the file-based flag changes nothing about audio.

### Evidence

- **Red→green pair:** new assertions in `localePin.test.ts` and `auth.smoke.test.ts` — the arg lists must carry the y4m path and must NOT carry `/dev/null` (no existing test covered this flag). `modules/join` suite: 12 passed; `modules/remote-browser` suite: both files pass.
- **The generated y4m decodes:** verified with the image's own ffmpeg — `ffprobe` reports `rawvideo, 640x360, yuv420p`, and a full `ffmpeg -i … -f null -` decode pass is error-free. Both the entrypoint (ffmpeg) and debug (Node-written) variants checked.
- **Validated live** on a self-hosted compose deployment at `v0.12.18` + this patch: the toast is gone from join onward on a real Meet call (confirmed via periodic captures of the bot's X display). Before the patch the toast was present on every Meet join.
- Changelog fragment included (`docs/changelog.d/998-fake-camera-decodable.md`).

Happy to have a non-author validate per the delivery loop — reproduction is one Meet join with any bot, looking at the bot's display (or a stage screenshot) for the toast.

### DIFF
--- a/core/meetings/modules/join/scripts/debug-join.ts
+++ b/core/meetings/modules/join/scripts/debug-join.ts
@@ -14,7 +14,21 @@
  */
 import { chromium } from "playwright-extra";
 import StealthPlugin from "puppeteer-extra-plugin-stealth";
+import { existsSync, writeFileSync } from "fs";
 import { joinMeeting, startDebugView, leaveGoogleMeet, leaveMicrosoftTeams, leaveZoomMeeting, getJoinBrowserArgs } from "../src/index";
+
+// The fake camera must point at a DECODABLE y4m — /dev/null registers a camera Chromium can
+// never read frames from, and Google Meet shows a permanent "Camera not found" toast over the
+// meeting stage (#998). The bot image generates this file in its entrypoint (ffmpeg); the debug
+// image carries no ffmpeg, so write the one black 640x360 yuv420p frame directly.
+const BLANK_CAM = "/tmp/blank-camera.y4m";
+if (!existsSync(BLANK_CAM)) {
+  const w = 640, h = 360;
+  const header = Buffer.from(`YUV4MPEG2 W${w} H${h} F30:1 Ip A1:1 C420jpeg\nFRAME\n`, "ascii");
+  const luma = Buffer.alloc(w * h, 0);
+  const chroma = Buffer.alloc((w * h) / 2, 128);
+  writeFileSync(BLANK_CAM, Buffer.concat([header, luma, chroma]));
+}
 
 
 if (process.platform !== "linux" || !process.env.DISPLAY) {
@@ -56,7 +70,7 @@ if (!isMeetUrl && !isTeamsUrl && !isZoomUrl) {
       "--no-sandbox", "--disable-setuid-sandbox", "--disable-blink-features=AutomationControlled",
       "--disable-infobars", "--disable-gpu", "--disable-features=IsolateOrigins,site-per-process",
       "--disable-site-isolation-trials", "--in-process-gpu", "--use-fake-ui-for-media-stream",
-      "--use-file-for-fake-video-capture=/dev/null", "--disable-features=VizDisplayCompositor",
+      "--use-file-for-fake-video-capture=/tmp/blank-camera.y4m", "--disable-features=VizDisplayCompositor",
       "--password-store=basic", "--remote-debugging-port=9222",
     ];
     const ctx = await chromium.launchPersistentContext(AUTH_PROFILE, {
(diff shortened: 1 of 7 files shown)


sashazykov:
I can not join Discord for some reason. I get "Unable to accept invite" error.

sashazykov:
**Closing this as invalidated — a second live run falsified the PR's evidence, and the mechanism it claimed was wrong.**

Second run, same deployment, patch fully in place — verified inside the live bot container: the y4m file exists and Chromium's command line carries the flag. The toast came back anyway.

The actual mechanism: `--use-file-for-fake-video-capture` is only honored when `--use-fake-device-for-media-stream` is enabled (per Chromium's `content_switches` docs). Neither of the bot's arg lists sets that flag, so the file value — `/dev/null` or a decodable y4m, it makes no difference — is **inert**. The container exposes no video device at all, `getUserMedia({video})` finds no camera, and Meet's "Camera not found" toast is literally true.

Why the first validation looked green: the initial run after deploying showed no toast, and I attributed that to this change. The variance (what makes the toast appear on one join and not another) is not understood — which is exactly why a single-run validation was insufficient, and that's on me.

Why not just add `--use-fake-device-for-media-stream`: it fakes the *microphone* as well as the camera, which would displace the `tts_sink` → `virtual_mic` PulseAudio chain the bot's speak path depends on — the same reason the PR itself gave for not using it.

I've posted a corrected root-cause analysis on #998 — the symptom is real and worth fixing, just not this way. One residual cleanup from this work stands on its own if you want it: `--use-file-for-fake-video-capture=/dev/null` is dead weight on all three sites and could simply be removed for clarity (no behavior change).

PR #1065, opened 2026-08-07
fix(gateway): let a bot-scoped key read its own meetings + transcripts (#1062) — TEMP PATCH pending #1066

### LINKED ISSUE #1062: P0: bot-scoped API key cannot read the transcript it captured — 'send a bot → get transcript' is broken with one key (bot vs tx scope split)
## The primary developer flow is broken with a single-purpose key

**"Send a bot to a meeting, then read the transcript"** — the headline Vexa use case — **cannot be completed with a `bot`-scoped API key.** Sending the bot works; reading the transcript it produced returns `403 Insufficient scope`.

### Live reproduction (hosted, api.cloud.vexa.ai, 2026-08-08)
A key with `scopes: ["bot"]` (verified via `GET /auth/me`):

| Call | Result |
|---|---|
| `GET /bots` (lists meetings) | ✅ 200 |
| `POST /bots` (send a bot) | ✅ (bot scope) |
| `GET /recordings` (the **audio**) | ✅ 200 |
| `GET /transcripts/by-id/{meeting_id}` (the **text**) | ❌ **403 Insufficient scope** |
| `GET /meetings` | ❌ **403 Insufficient scope** |

**The asymmetry that makes this a bug, not a policy:** a `bot` key can read the **recording (audio)** of a meeting but not its **transcript (text)** — reading transcripts requires the separate `tx` scope (`core/identity/services/admin-api/src/admin_api/token_scope.py:15` — `VALID_SCOPES = {"bot","tx","browser"}`; a full token is `["bot","tx","browser"]` per `schema/sync.py:128`). So the key scoped for the obvious first action (drive a bot) cannot perform the obvious second action (get its transcript), while it *can* fetch the heavier audio artifact.

### Why P0
This is the first thing every developer does. If onboarding/quickstart mints or documents a `bot` key (or a key missing `tx`), the developer sends a bot, then hits 403 on the exact thing they came for — the transcript. It reads as "the API doesn't work."

### Fix options (pick one, OSS)
1. **`bot` scope implies read of the transcripts/meetings the same key's bots produced** (own-resource read) — least surprising; the key that captured it can read it.
2. **Quickstart/onboarding mints a `bot,tx` key by default**, and every "send a bot → get transcript" doc example uses a key with both — and says so loudly.
3. At minimum, make the 403 body name the missing scope and how to mint a key that has it (currently just "Insufficient scope for this endpoint").

### Acceptance
- A key created via the documented quickstart flow can complete `POST /bots` → `GET /transcripts/by-id/{id}` end-to-end with no 403.
- OR: the docs quickstart explicitly mints `bot,tx` and a `bot`-only key's 403 names the missing scope + the mint command.
- Regression: a bot-scope key reading transcripts of meetings it did NOT create still correctly 403s (own-resource boundary preserved).

### Notes
Related surface drift (separate, minor): `GET /bots` returns **meetings** (naming inverted vs `Create Bot`); `GET /transcripts/{platform}/{native_id}` returns 403 but is **not in `/openapi.json`** — a compat shim the docs still reference (docs-parity, #534). Canonical is `/transcripts/by-id/{meeting_id}`.

DmitriyG228:
## The bug (#1062)

The headline flow — **send a bot → read its transcript** — cannot be completed with a single API key, because `bot` and `tx` are mutually-exclusive scopes at the gateway. A user who holds a `bot`-scoped key to spawn bots gets 403'd when they try to list the resulting meetings or read the transcripts those bots produced.

### Reproduction (verified live 2026-08-08)

| Request | `bot`-scoped key | `tx`-scoped key |
|---|---|---|
| `GET /bots` | ✅ 200 | ❌ 403 |
| `POST /bots` | ✅ ok | ❌ 403 |
| `GET /recordings` | ✅ 200 | ✅ 200 |
| `GET /meetings` | ❌ **403** | ✅ 200 |
| `GET /transcripts/by-id/{id}` | ❌ **403** | ✅ 200 |

There is no single key that can both spawn a bot **and** read what it captured. That is the P0.

## Root cause

`ROUTE_SCOPES` in `core/gateway/services/gateway/src/gateway/app.py` gated `/meetings` and `/transcripts` on `{tx}` only:

```python
ROUTE_SCOPES = {
    "/bots":        {"bot", "browser"},
    "/transcripts": {"tx"},          # ← bot key 403s here
    "/meetings":    {"tx"},          # ← bot key 403s here
    "/recordings":  {"tx", "bot"},
}
```

`bot` is **intended** to read the meetings + transcripts of the bots it created (`tx` is a separate transcription-only scope). So this gate is a bug, not the intended boundary.

## The fix

Add `bot` alongside `tx` on `/meetings` and `/transcripts`:

```python
ROUTE_SCOPES = {
    "/bots":        {"bot", "browser"},
    "/transcripts": {"tx", "bot"},   # +bot
    "/meetings":    {"tx", "bot"},   # +bot
    "/recordings":  {"tx", "bot"},
}
```

### Why this is safe — the routes are already user-scoped

`/meetings` and `/transcripts` are **user-scoped downstream**: the gateway resolves the key to a `user_id`, injects `X-User-Id`, and meeting-api returns **only that user's** rows (the `/transcripts/by-id/{id}` route is owner-scoped per row — the same P0 cross-tenant fix already in the tree). So accepting `bot` on these prefixes lets a user read the data their **own** bot captured; it does **not** let a bot key read another user's meetings or transcripts. The scope gate was redundantly strict *on top of* the user-scoping that already enforces ownership.

`tx` stays a valid transcription-only reader (added `bot`, did not replace `tx`). `/bots` is unchanged and still requires `bot`/`browser` — a `tx`-only key correctly continues to 403 there.

> Note: the historical `services/api-gateway/main.py` this file was carved from is no longer in the tree, so `app.py`'s `ROUTE_SCOPES` is now the single live source — there is no second copy to drift against.

## Tests (all green)

Added to `core/gateway/services/gateway/tests/test_proxy.py`:

- ✅ `bot`-scoped token → `GET /meetings` now **passes** (was 403), forwards to meeting-api with `X-User-Id` injected.
- ✅ `bot`-scoped token → `GET /transcripts/by-id/{id}` and `GET /transcripts/{platform}/{native}` now **pass** (were 403).
- ✅ `bot`-scoped token → `GET /bots` still passes (regression guard).
- ✅ `tx`-only token → `/bots` still **403** (regression guard; preserves the existing `test_insufficient_scope_is_403` boundary).
- ✅ `tx`-scoped token → `/meetings` + `/transcripts` still pass (the fix adds `bot`, it does not remove `tx`).

Verified the new positive tests go **RED** against the unfixed `ROUTE_SCOPES` (`request_denied_scope … required:["tx"]`) and **GREEN** with the fix.

```
gateway suite:     117 passed, 1 xfailed
conformance suite:  84 passed
```

Run: `cd core/gateway/services/gateway && uv run pytest -q` (and `../conformance` likewise).

## Follow-up (design note — NOT in this PR)

The next step the founder wants is for **agents to attach custom metadata to meetings** so an agent can "own the data" for retrieve + search. The natural home is the existing `meeting.data` JSONB blob (`core/meetings/services/meeting-api/src/meeting_api/`), which already holds recording/config keys.

Proposed follow-on feature (separate issue, not this PR): a namespaced **`PATCH /meetings/{id}/metadata`** that writes under `data.custom.*` only — never clobbering the recording/config keys already in `data` — scoped so the **owning user's key** can write (same user-scoping this PR relies on for reads). This keeps agent-authored metadata isolated from system-managed meeting state and makes it addressable for later retrieve/search.

---

Fixes #1062. **Draft** — founder reviews before ready-for-review.

## Contribution rights

### DIFF
--- a/core/gateway/services/gateway/src/gateway/app.py
+++ b/core/gateway/services/gateway/src/gateway/app.py
@@ -60,10 +60,18 @@ def _auth_unavailable_response(exc: Exception, *, span: str) -> Response:
 
 # Route-prefix → required scope set. Mirrors main.py ROUTE_SCOPES (main.py:59-65) for the CORE
 # surface the gateway lane carves; multi-scope tokens pass for any of their domains.
+#
+# #1062 — the headline flow "send a bot → read its transcript" must work with ONE key. A `bot`
+# scope therefore reads /meetings + /transcripts too (not only /recordings): these routes are
+# already USER-scoped downstream (the key resolves to a user_id and meeting-api returns ONLY that
+# user's rows), so accepting `bot` here lets a user read the data their OWN bot captured — it does
+# NOT let a bot key reach another user's meetings. The scope gate was redundantly strict on top of
+# the user-scoping. `tx` stays a valid transcription-only reader; /bots still requires `bot`/`browser`
+# (a `tx`-only key correctly 403s on /bots).
 ROUTE_SCOPES: Dict[str, Set[str]] = {
     "/bots": {"bot", "browser"},
-    "/transcripts": {"tx"},
-    "/meetings": {"tx"},
+    "/transcripts": {"tx", "bot"},
+    "/meetings": {"tx", "bot"},
     "/recordings": {"tx", "bot"},
 }
 
--- a/core/gateway/services/gateway/tests/test_proxy.py
+++ b/core/gateway/services/gateway/tests/test_proxy.py
@@ -80,6 +80,73 @@ def test_insufficient_scope_is_403():
     assert "scope" in r.json()["detail"].lower()
 
 
+def _bot_only():
+    """A `bot`-scoped token (no `tx`) — the key a headless bot-runner holds."""
+    return FakeAuthorizer(user={"user_id": 7, "scopes": ["bot"], "max_concurrent": 1})
+
+
+def test_bot_scope_reads_own_meetings():
+    """#1062: the SAME key that spawns a bot must read that bot's meetings. A `bot`-scoped token on
+    GET /meetings now passes (was a 403 — /meetings required `tx` only), forwarding to meeting-api,
+    which owner-scopes the rows to this user_id. Fixes the "send a bot → can't list its meetings" gap."""
+    client, downstream = _client(authorizer=_bot_only())
+    r = client.get("/meetings", headers=AUTH)
+    assert r.status_code == 200
+    assert downstream.last["method"] == "GET"
+    assert downstream.last["url"].endswith("/meetings") and "meeting-api" in downstream.last["url"]
+    assert downstream.last["headers"]["x-user-id"] == "7"  # owner-scoped downstream, not client-set
+
+
+def test_bot_scope_reads_own_transcript_by_id():
+    """#1062: a `bot`-scoped token on GET /transcripts/by-id/{id} now passes (was 403 — /transcripts
+    required `tx` only). The by-row-id read is owner-scoped downstream, so the bot key reads ONLY the
+    transcript of a meeting this user owns. Completes the one-key "send a bot → read its transcript" flow."""
+    client, downstream = _client(authorizer=_bot_only())
+    r = client.get("/transcripts/by-id/1234", headers=AUTH)
+    assert r.status_code == 200
+    assert downstream.last["method"] == "GET"
+    assert downstream.last["url"].endswith("/transcripts/by-id/1234")
+    assert "meeting-api" in downstream.last["url"]
+
+
+def test_bot_scope_reads_own_native_transcript():
+    """#1062: same carve for the native-keyed transcript read GET /transcripts/{platform}/{native}."""
+    client, downstream = _client(authorizer=_bot_only())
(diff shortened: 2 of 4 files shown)


DmitriyG228:
**Founder ruling 2026-08-08: this is a TEMPORARY PATCH, to be treated as such.** It repairs the broken headline flow (`POST /bots` → `GET /transcripts` with one key) by widening the hardcoded `ROUTE_SCOPES` union — the minimal tactical fix. The structural fix is #1066 (registry/config-driven scopes: OSS ships a minimal capability baseline where the user key covers the whole meeting capability; the platform overlays `tx` in its own deployment config). **#1066's acceptance includes retiring this hardcoded union** — when the registry lands, this widening is subsumed by the baseline scope definition and the literal `{tx, bot}` union in `ROUTE_SCOPES` goes away with the rest of the hardcoded map. Design rationale: https://github.com/Vexa-ai/vexa/issues/1066#issuecomment (2026-08-08 design-position comment).

PR #1067, opened 2026-08-08
feat: agent-owned meeting metadata (attach/amend/filter data.custom) — flywheel iteration 2 (#1064)

### LINKED ISSUE #1064: feat: agent-owned meeting metadata — attach custom metadata in-call (POST /bots), amend (PATCH), filter (GET /meetings?custom.*) — flywheel iteration 2
## Agent-owned meeting metadata — attach in-call, filter/search by it

**Vision (flywheel iteration 2):** let an agent OWN a meeting, not just capture it — by attaching custom metadata it can later filter and reason over. Turns the transcript store into the agent's own queryable corpus. Follows #1062 (iteration 1: agents can read what they captured).

### Capability
1. **Attach in-call:** `POST /bots` accepts an optional `metadata` object (`{project, purpose, agent_owner, tags:[...], ...}`), persisted under `meeting.data.custom.*` (JSONB). The `meeting.data` field already exists and holds recording/config — write custom keys under a reserved `custom` namespace so nothing clobbers `data.recording` / `data.config`.
2. **Amend post-hoc:** `PATCH /meetings/{platform}/{native_meeting_id}/metadata` (or `/meetings/{id}/metadata`) merges into `data.custom.*`. Owner-scoped (the user's own key only; there is already a native-meeting PATCH surface — extend it).
3. **Filter/search:** `GET /meetings?custom.<key>=<value>` (and/or `?tag=`) filters the meeting list by attached metadata — so an agent retrieves *its* meetings by its own classification. Postgres JSONB `@>` / GIN-index makes this cheap.

### Why it compounds (the loop)
An agent that classifies meetings as it dispatches them can later filter, retrieve, and reason over its corpus — the transcript store becomes the agent's memory. Each iteration (better metadata surface → agents attach more → larger owned corpus → more retrieval/reasoning) spins the discovery flywheel.

### Scope / auth
Metadata write + filtered read are OWN-RESOURCE (the key's user only), consistent with the #1062 user-scoping model. A `bot`-scoped key (which dispatched the meeting) should be able to attach + read its own metadata.

### Acceptance
- `POST /bots` with `metadata:{project:"X"}` → the meeting persists `data.custom.project="X"`, no clobber of existing `data` keys.
- `PATCH .../metadata {tags:["sales"]}` merges (doesn't replace) `data.custom`.
- `GET /meetings?custom.project=X` returns only matching meetings.
- Regression: `data.recording` / `data.config` untouched by any metadata write; cross-user isolation holds.

### Notes
Data model already supports it (`meeting.data` JSONB, `core/meetings/services/meeting-api/`). Depends on / sequences after #1062 (same files). Also expose the metadata fields in `/openapi.json`.

DmitriyG228:
## Agent-owned meeting metadata — flywheel iteration 2 (#1064)

Let an agent **OWN** a meeting, not just capture it: attach custom metadata it can later filter, retrieve, and reason over. The transcript store becomes the agent's own queryable corpus. Follows #1062 (iteration 1: agents can read what they captured).

### Stacking — READ THIS FIRST
This branches **off `fix/1062-bot-scope-transcript-read`** (PR #1065), not `main`, and is opened as a **stacked PR against that branch** so the diff is *only* this feature (627 insertions). It depends on #1065: agents must be able to READ their meetings (`GET /meetings` reachable by a `bot`-scoped key) to filter them — the `custom.*` filter is worthless if the read 403s.

**Merge order:** land #1065 first, then **retarget this PR to `main`** (GitHub will auto-shrink the diff once #1065 is in `main`).

### The 3 endpoints + response exposure
Everything writes under a **reserved `meeting.data['custom']` namespace** — a sibling of `data.recording` / `data.config` / `data.docs`.

1. **Attach in-call** — `POST /bots` accepts an optional `metadata` object → persisted at spawn under `data.custom.*` (in the same row `create`, so it lands atomically with the spawn keys).
2. **Amend post-hoc** — `PATCH /meetings/{platform}/{native_meeting_id}/metadata` **MERGES** the body into `data.custom.*` (shallow merge, never replace). Owner-scoped (the token's `user_id` only).
3. **Filter** — `GET /meetings?custom.<key>=<value>` (repeatable, AND-combined; `custom_<key>=` also accepted) → Postgres JSONB containment `data @> {"custom": {...}}`.
4. **Expose** — `data.custom` surfaces in the `/meetings` (it's a *light* list key, kept by `project_list_data`) and `/transcripts` responses; `recording`/`config` stay cleaned as before.

### Atomic no-clobber approach
The metadata write reuses the recordings flow's discipline (`recordings/service.py` → `mutate_recordings`): a **`SELECT … FOR UPDATE` row lock spanning read-through-commit**, with a mutator that touches **only** the `custom` key. So a concurrent `data['recordings']` chunk/finalize write (or a config write) can neither be lost nor clobber the metadata — proven at unit level (`test_store_merge_preserves_sibling_keys`, `test_patch_metadata_no_clobber_of_recording_and_config`).

### The filter query + index
```sql
WHERE meetings.data @> CAST($1::JSONB AS JSONB)   -- $1 = {"custom": {"project": "apollo", ...}}
```
Containment on the **whole `data` column** (not the `data->'custom'` sub-object) so the **existing** whole-column GIN index `ix_meeting_data_gin` (`sessions/models.py:75`) serves it — **no migration required**.

### Auth
Metadata write + filtered read are **OWN-RESOURCE** (the token's `user_id`), consistent with #1062's user-scoping. `/meetings` is in the gateway `ROUTE_SCOPES` as `{tx, bot}` (the #1062 change), so a **`bot`-scoped key** — the one that dispatched the meeting — can attach + amend + filter its own metadata. Enforced downstream in meeting-api by `X-User-Id`.

### Validation (boundary, typed 422 — never a 500)
`meeting_api/metadata.py` (new, pure): the object must be a **flat-ish** JSON object — string keys → scalar (`str`/number/`bool`/`null`) or a **flat list of scalars** (`tags: ["sales"]`); **no nested objects**; ≤ **8 KB** serialized; ≤ 64 keys.

### OpenAPI
`core/gateway/contracts/api.v1/api.schema.json` updated (identity unchanged: `Vexa API Gateway` `1.5.0`):
- `metadata` property on `MeetingCreate` (POST /bots),
- new `PATCH /meetings/{platform}/{native_meeting_id}/metadata` path,
- `custom.<key>` filter param + note on `GET /meetings`.

The sealed contract was **re-sealed** (`pnpm seal:contracts`) — a one-line `contracts.seal.json` diff for a back-compatible additive change (**lane:contract**). `validate.mjs` green; `gate:contract-version` green.

### Test evidence (all green)
- **meeting-api:** `913 passed, 5 skipped` (25 new in `test_agent_metadata.py`).
- **gateway:** `119 passed` (2 new proxy tests: metadata forward + bot-scope amend).
- **gateway conformance:** `84 passed`.
- **api.v1 validate.mjs:** all checks pass.
- **gate:contract-version:** green.

New tests cover: attach persists `data.custom.project="X"` with spawn keys intact · PATCH merges (doesn't replace) · no-clobber of `data.recording`/`data.config` · `?custom.project=X` returns only matches · non-matching filter returns none · **cross-user isolation** (user A's filter never returns user B's meetings) · exposure in list + transcript responses · abuse (nested/oversize/non-object) → 422.

**DRAFT — do not merge.** Land #1065 first, then retarget to `main`.

Closes #1064 (after #1065).

## Contribution rights


DmitriyG228:
## Two production-killing defects in the `custom.*` filter — fixed and proven on real Postgres (`cf77c560`)

A real-Postgres run of **this branch's own adapter** (using the PG fixture built for #1074) found the metadata filter was broken in two independent ways. Both were invisible to the in-memory fake, which is the whole reason they survived two rounds of green CI.

### Defect 1 — double-serialized JSONB bind: *every* filter returned zero rows

The probe was built as:

```python
Meeting.data.op("@>")(cast(_json.dumps({"custom": dict(custom_filter)}), JSONB))
```

SQLAlchemy's JSONB bind processor serializes the parameter itself, so the already-serialized string bound as a jsonb **string scalar**. `object @> string` is silently `false` — no error, no warning, just an empty list. On Postgres 16, against a meeting stored with `data.custom = {"score": 42}`:

```
no filter                  -> [1]
custom_filter={'score':42} -> []      # and likewise every other filter
```

**Fix:** bind the dict itself so SQLAlchemy serializes exactly once.

### Defect 2 — the nested array-membership claim was false

The code (and its comment) claimed `{"tags":["sales","q3"]} @> {"tags":"sales"}` is TRUE. It is FALSE — Postgres' array-contains-primitive exception applies **only at the top level**, never nested under a key:

```sql
SELECT '{"tags":["sales","q3"]}'::jsonb @> '{"tags":"sales"}'::jsonb;    -- f
SELECT '{"tags":["sales","q3"]}'::jsonb @> '{"tags":["sales"]}'::jsonb;  -- t
```

So the headline agent use case — filter my meetings by a tag — matched nothing even once defect 1 was fixed.

**Fix:** each key becomes an OR of two containment probes, AND-ed across keys:

```python
stmt = stmt.where(
    and_(*[
        or_(
            Meeting.data.op("@>")(cast({"custom": {k: v}}, JSONB)),    # stored SCALAR, type-strict
            Meeting.data.op("@>")(cast({"custom": {k: [v]}}, JSONB)),  # stored ARRAY, element membership
        )
        for k, v in custom_filter.items()
    ])
)
```

Both probes are whole-column containments, so `ix_meeting_data_gin` still serves them — no migration, no new index.

The fake's existing `_jsonb_scalar_contains` semantics (type-strict scalar equality · array → element membership · object → no match) are *exactly* what the OR-of-two-probes now means in SQL. That alignment was verified case-by-case against PG 16 (scalar, array, object, `null`, bool-vs-number, empty array, absent key) rather than assumed — the fake needed no change. The wrong comment is corrected in `adapters.py`, `fakes.py` and `ports.py`.

### The proof (this is what fake-green hid twice)

New `tests/test_pg_custom_filter.py` runs the shipped `SqlAlchemyTranscriptStore.list_meetings` against a real schema in a real Postgres, env-gated on `MEETING_API_TEST_DATABASE_URL` via `tests/pg_fixtures.py` (temp schema per session, dropped in teardown; CI without a database skips cleanly):

| assertion | before | after |
|---|---|---|
| `{'score': 42}` returns the row stored with the number 42 | `[]` | ✅ |
| string probe matches stored string, never stored number (type-strict) | `[]` | ✅ |
| `{'tags': 'sales'}` matches `["sales","q3"]`, not `["q3"]` | `[]` | ✅ |
| absent key matches nothing · multi-key AND | `[]` | ✅ |
| user B's filter never returns user A's meetings | `[]` | ✅ |

```
# pre-fix adapter, same tests, real PG:  5 failed in 0.98s   (first failure: assert [] == [1])
# post-fix:                              5 passed in 0.84s
```

### Gates

- meeting-api: **919 passed / 10 skipped** (no PG) · **927 passed / 2 skipped** (with `MEETING_API_TEST_DATABASE_URL`)
- gateway: **119 passed / 1 xfailed**
- `node scripts/gates.mjs all`: **green** (35 gates, incl. the real compose stack-readiness proof)

Still draft, not merged.

Issue #1072, opened 2026-08-08
release: Join-Evidence Instrument train (v0.12.22 scope) — typed join-failure evidence, end to end


DmitriyG228:
## The train

Converts \`train/observability\` from investigative to a merge batch: make every join failure carry **typed, structured evidence with an attribution dimension**, so the **system-failure rate** becomes a measurable, trendable instrument instead of an anecdote.

**The metric this train unblocks (founder ruling 2026-08-08):** the reliability gate metric is the **SYSTEM-failure rate** — failures where the system failed to do its job (join/navigation failure, auth loss, crash) over meetings where it had a fair chance — **never** the raw failed-status rate. A bot the user stopped, or a host never admitted, is a funnel outcome, not a system failure.

**Why now (fundability + flywheel):** today the system-failure rate is *uncomputable* — all production failure rows carry \`reason: None\` (#1059), and \`failure_stage\` alone cannot attribute (e.g. \`awaiting_admission\` = host choice OR a stuck admission flow — indistinguishable). The quotable trend (segmented, ordinary users: ~8.2% May → ~32% July outcome-rate approximation) cannot be decomposed into system vs user/host vs platform-exogenous until this instrument exists. The reliability release-gate and the endogenous-vs-exogenous regression split are unbuildable without it. Prior context: #806, #807, #826.

> Correction note: this issue's original body cited "~32% of hosted meetings fail to join (confirmed twice)". That framing was struck — the API-probe sample was a single owner key (dev-polluted) and the count included user/host actions; the numeric match with the segmented July number was coincidental, not confirmatory.

## Cars

| Car | Issue | Acceptance |
|---|---|---|
| 1 | #1059 | every join failure persists structured evidence (typed reason, failure_stage, stage timings, raw detail) on the meeting row; unit-tested; replayable fixture per failure class |
| 2 | #1058 | typed failure taxonomy **with attribution dimension: system_fault / user_action / host_action / exogenous_platform / unknown**; platform-rejection (≈20s Teams) ≠ admission-timeout (≈13min Meet) — distinct terminal states, no conflation |
| 3 | #1061 | Zoom \`auth_session_missing\` detected + surfaced as a typed reason, attribution system_fault (self-heal = follow-up) |
| 4 | #1068 | env-gated Postgres test fixture for meeting-api (also unblocks #1067's isolation-test follow-up) |

[...]

Issue #1068, opened 2026-08-08
test(meetings): #1064 cross-user isolation + 404-when-not-owned are only fake-covered — add a Postgres-backed isolation test


DmitriyG228:
## Summary

The agent-owned-metadata (#1064) **cross-user isolation** and **404-when-not-owned** guarantees are asserted **only against the in-memory fake** (`InMemoryTranscriptStore`). The real SQL that actually enforces the boundary in production is not exercised by CI.

This is **test-coverage debt, not a code defect** — the production adapter (`SqlAlchemyTranscriptStore`) is correct today. The gap is that the thing CI guards (the fake) is not the thing that ships (the SQL).

## What is only fake-covered

`core/meetings/services/meeting-api/tests/test_agent_metadata.py`:
- `test_filter_cross_user_isolation` — user A's `?custom.*` filter must never return user B's meetings.
- `test_patch_metadata_404_when_not_owned` — PATCH `.../metadata` on a meeting owned by someone else → 404.

Both drive `InMemoryTranscriptStore`, whose isolation is hand-written Python (`m["user_id"] == user_id` predicates), not the real SQL.

## The real SQL the boundary actually rides on

`core/meetings/services/meeting-api/src/meeting_api/collector/adapters.py`:
- **PATCH scoping** — `merge_custom_metadata` selects `WHERE Meeting.user_id == user_id AND platform AND platform_specific_id … FOR UPDATE` (owner-only write).
- **GET filter scoping** — `list_meetings` builds an access **UNION** (owner / transcript-share viewer / bound-workspace member), and the `custom_filter` JSONB `@>` predicate is **ANDed onto the outer query** over that union. So isolation depends on the union being correct *and* the filter not widening it.

A regression in either (e.g. an accidental `OR`, a missing `user_id` predicate, or the filter applied before the union) would keep every fake-based test green while leaking across tenants in production.

## Ask

Add a **Postgres-backed** isolation test (gated exactly like the existing real-PG conformance test — `@pytest.mark.skipif(not os.getenv("MEETING_API_TEST_DATABASE_URL"))`, see `tests/test_single_flight.py::test_pg_advisory_lock_runs_on_real_postgres`) that:
1. seeds two users' meetings (same native id, same `custom`) in a real `meetings` table,
2. asserts user A's `?custom.<k>=<v>` list returns only A's rows,
3. asserts PATCH `.../metadata` on B's meeting as A → 404 / no write,
[...]

Issue #926, opened 2026-07-23
zoom: bot exits code 1 with reason:None ~12s after spawn — 2 users back-to-back, no captured cause


DmitriyG228:
## What prod is showing

During the v0.12.16 prod-owner watch, two **Zoom** bots from two **different** users failed with an identical, uninformative signature within ~24s of each other:

| meeting | user | created (UTC) | failed (UTC) | lifetime |
|---|---|---|---|---|
| 24672 | 419 | 2026-07-23 12:11:32 | 12:11:43 | ~11s |
| 24674 | 421 | 2026-07-23 12:11:56 | 12:12:08 | ~12s |

`data.last_error` for both, verbatim:

```json
{"reason": null, "exit_code": 1, "error_details": "Bot exited with code 1; reason: None"}
```

## Why this is its own finding (not #862 / #915)

- **Not #862** (`workload destroyed while awaiting admission`) — no admission phase reached; the bot exits code 1 almost immediately (~12s), not the ~20–325s lobby-death profile.
- **Not #915** (`no meeting indicators found after polling`) — it never gets far enough to poll.
- **Not [#897](https://github.com/Vexa-ai/vexa/issues/897)** (URL parser drops the pasted URL so `meeting_url` is never sent) — here the meeting IS created and a bot IS spawned; it then dies.

Two *distinct* users failing the same way within 24s reads as systemic (a shared dependency / zoom-side change / spawn-path regression), not a single bad link.

## The fail-loud gap (cf. #698)

`reason: None` + `exit_code: 1` is exactly the silent/uninformative failure #698 is about: the bot process exited non-zero but nothing captured *why*, so the outcome is unactionable from the DB alone. Root-causing this needs the bot pod's own stderr at exit (which is discarded here).

## Asks
1. Capture and persist the Zoom bot's exit reason/stderr on `exit_code != 0` (so `reason` is never `None`).
2. Investigate the early exit-1 on the Zoom spawn path for uid 419/421 around 2026-07-23 12:11 UTC.

## Provenance
Filed from the v0.12.16 prod-owner watch (file-don't-fix). Observed via `meetings.data->>'last_error'`, internal users (2586/3175/3487) excluded. Only 2 occurrences so far — flagging early because two different users hit it simultaneously.

DmitriyG228:
[...]

Issue #806, opened 2026-07-19
Bot join reliability: joining-stage failures are the top organic failure (198 in 10 days, google_meet-dominant) — and failure detail is unstructured bot_log blobs


DmitriyG228:
## Where we are (hosted production data, 2026-07-19, prod on v0.12.12)

Excluding two runaway integration users, the organic bot failure picture over the last 10 days:

- **42% of organic meeting attempts failed** (167/397 in the last 7 days).
- Top failure stages: `joining` (198 across the window, google_meet-dominant: 113 before 07-13, 85 after), then `awaiting_admission` (18 organic).
- Daily failure rate stepped from ~19% (07-12) to 62–81% (07-13 onward) — a regression consistent with a Meet UI change breaking join selectors (same pattern as the 06-15/06-22 spike that recovered on 06-29). Prod core was UNCHANGED (0.10) across 07-13, so the trigger is external (platform UI), not a deploy.

## Two gaps

1. **Join reliability (the value path)**: `joining`-stage failures mean the bot never entered the meeting — the product delivered nothing. This is the single largest organic value-loss driver.
2. **Failure observability**: `meetings.data` for a failed meeting carries only a raw `bot_logs` array — there is no structured `failure_reason` taxonomy (selector_not_found / admission_timeout / captcha / removed / crash). Aggregation across failures requires parsing log blobs, so regressions like 07-13 are invisible until someone reads the DB by hand.

## Prepared direction

- Add a structured failure taxonomy at the point the join brick gives up (`>>> [JOIN-STATE]` transitions already exist) and persist it as `data.failure_reason`.
- Alerting/regression detection belongs downstream of that field (hosted side will consume it — see vexa-platform SLO issue).
- Reproduce the current google_meet joining failure in the join debug harness (`modules/join`, noVNC) against a live meeting to identify the broken selector/flow.

Origin: users → prod → system logs (DB) → OSS issue.

DmitriyG228:
**Correction after deeper evidence — the initial interpretation was wrong about the cause.**

Aggregating `data.last_error` on the organic google_meet joining failures (07-14→07-19) reframes this:

| error class | count | meaning |
|---|---|---|
| `joining callback failed: server rejected after 3 attempts` / `All 3 callback attempts failed` | 32 | **bot was fine — meeting-api rejected/failed the status callback** |
| no error detail at all | 40 | observability gap (this issue's taxonomy ask, confirmed) |
[...]

PR #1075, opened 2026-08-08
feat(bot,meetings): typed join-failure evidence — reason, stage, timings persisted (#1059, #1058)

### LINKED ISSUE #1058: join_failure conflates two different failures: a 20s Teams rejection and a 13min Meet timeout
Measured on hosted production, current release only (helm rev 106, since 2026-07-25): 924 meetings, 83 `join_failure`.

**The two populations have nothing in common except the label.**

| platform | count | avg time-to-death | min | max |
|---|---|---|---|---|
| teams | 37 | **19.6s** | 14.8s | 47.6s |
| google_meet | 46 | **793.2s** | 4.8s | 1236.8s |

Teams clusters tightly at ~20s — a deterministic, immediate rejection. Meet averages 13 minutes with a max of 20.6 min — a bot that sat until something reaped it, i.e. a timeout, not a refusal. Only one Meet row (4.8s) looks like a true join failure.

**Cause:** `_pre_active_completion_reason` (`core/meetings/services/meeting-api/src/meeting_api/lifecycle/reconcile.py:464-478`) returns `join_failure` for ANY pre-active stage that is not `awaiting_admission`:

```python
return "awaiting_admission_timeout" if status == "awaiting_admission" else "join_failure"
```

So "refused by the platform while joining" and "never got anywhere and was reaped" collapse into one reason. They have different owners and different fixes, and both are TRANSIENT in `retry.py`, so both get retried identically — including the ones that will deterministically fail again.

**Impact:** the platform-fault rate cannot be decomposed. Teams reads 17.8% platform-fault vs Meet 6.7%, but the Teams number is real rejections while a large share of Meet's is reaping. Any alert or SLO built on `join_failure` measures two unrelated things.

**Fix shape:** distinguish by the evidence the reconciler already holds — elapsed time in stage and which pre-active stage it died in. Suggested split: `join_rejected` (died in `joining` well inside the budget — the platform refused us) vs `join_timeout` (exhausted the join/lobby budget without a verdict). Keep both TRANSIENT if that is the desired retry policy, but let them be counted separately.

Related: #<this repo's> attribution work — the reasons are good, the axis is missing (see the companion issue on fault ownership).

### LINKED ISSUE #1059: join_failure records no evidence — all 83 production rows carry reason: None
Measured on hosted production, current release (since 2026-07-25): **every one of the 83 `join_failure` meetings has `data.reason = None`.**

So the system records THAT a bot died on its way into the meeting and nothing about WHY — no platform error, no HTTP status, no page state, no container exit code. The bot container knew; the database does not. Diagnosis stops at the shape of the failure.

**This path is the odd one out.** `_escalate_untracked_zombie` (`lifecycle/reconcile.py:226-240`) writes a real evidence note:

```python
reason = (f"workload {bot_container_id} untracked for >{int(grace)}s; presumed lost — "
          "runtime restart or external removal")
```

and the confirmed-teardown path immediately below it sets `completion_reason` but leaves `reason` unset. The file's own comment states the principle — *"ATTRIBUTE, never manufacture (#862). The reason is DERIVED from the stage the bot died in, and the note carries the probe's own answer (workload state, exit code) — the only things this sweep actually knows."* — but on this branch the note is not carried.

**Impact:** 9.4% of production meetings fail with platform attribution and none of them can be diagnosed from data. The next investigation repeats the same analysis and lands in the same place. This is the same invisible-failure class as the billing defects found in the t066 train: the system knows more than it records, and a number nobody can decompose is a number nobody can act on.

**Fix shape:** carry the probe's own answer into `data.reason` on the confirmed-teardown branch, exactly as the zombie-escalation branch does — workload state, exit code, and whatever the bot reported before it died. Where the bot itself observed a platform error (a Teams/Meet refusal string, an HTTP status), that belongs in the record too.

**Evidence:** `SELECT data->>'reason', count(*) FROM meetings WHERE created_at >= '2026-07-25' AND data->>'completion_reason'='join_failure' GROUP BY 1` → `(None, 83)`.

DmitriyG228:
Part of the Join-Evidence Instrument train (#1072) — Cars 1+2, built together because they are the same failure path.

Closes #1059
Closes #1058

## What was broken

**#1059 — `reason: None` on all 83 rows.** Not a missing report: a missing *projection*.

Every producer already stamped a human reason. The bot writes one on every terminal it emits (`orchestrator.ts` — `#926` made sure of it), the reconcile sweep writes the probe's own answer, and `MeetingRecord.reason` carried it into the FSM. But `MeetingRecord.data` — the JSONB projection that becomes `meeting.data` — only ever wrote it into **`last_error.reason`**, a nested key that (a) `SELECT data->>'reason'` cannot see, and (b) `collector/projection.LIST_OMIT_KEYS` *drops from every list response* as a heavy forensics key. So the reason was recorded in a place no query and no list view looks. `data.reason` is now a top-level key.

**#1058 — one label, two populations.** `completion_reason` is the sealed retry-classifier input; it has no value for "the platform refused us" that is distinct from "nobody answered the door". Both arrive as `join_failure`, are both TRANSIENT, and are both retried — including the ones that will deterministically fail again.

## The taxonomy — two independent axes

Carried **alongside** the sealed lifecycle.v1 `CompletionReason`, which is untouched (it drives `retry.py`; this is diagnostic, not a retry input).

| `reason` — WHAT happened | Meaning |
|---|---|
| `platform_rejection` | The platform/host actively denied us (moderator deny, automated-join block, captcha). |
| `admission_timeout` | Reached the waiting room; nobody ever answered; the issued lobby budget ran out. |
| `auth_session_missing` | Authenticated mode found a signed-out profile. |
| `never_reached_lobby` | The page loaded but no admission signal ever appeared (selector rot / platform UI change — the #806 class). |
| `navigation_failure` | Never got as far as the meeting page (nav error, off-origin redirect, closed target, control plane down at boot). |
| `stopped_before_admission` | The **user** ended the run while the bot was still knocking. |
| `unknown` | The signals do not support a verdict. `detail` carries what was observed. |

| `attribution` — WHO it belongs to | |
|---|---|
| `system_fault` | We failed to do our job: join flow, navigation, session, crash, or a wedged admission flow. |
| `user_action` | The user ended it themselves. Excluded from the metric's denominator entirely. |
| `host_action` | The host denied us or never answered. A funnel outcome, not a defect. |
| `exogenous_platform` | The platform itself refused or changed under us, **positively identified**. |
| `unknown` | No owner is supportable. Counted separately — never folded into `system_fault` (which would inflate the metric with our own ignorance) nor into `host_action` (which would hide real defects). |

The axes are genuinely independent, and two evidence-driven overrides prove it: a `platform_rejection` whose detail names an automated-join policy block (`zoom_requires_rtms`, a captcha wall) reattributes `host_action → exogenous_platform`; an `admission_timeout` whose detail shows **our** join layer went blind (`no meeting indicators found`, a selector miss) reattributes `host_action → system_fault` — the host never got the chance to decline. That is the founder's "UNLESS evidence shows the admission flow itself broke" carve-out, made mechanical.

### The metric this enables

```
system_failure_rate = failures(attribution = system_fault) / fair-chance meetings
```

per **release** and per **platform**, straight off the persisted rows — no log parsing. A *fair-chance* meeting is one the system was actually given the opportunity to serve, which excludes `user_action`: a run the user ended before admission was never ours to win. Both `#1058` populations are now countable separately, and only one of them is ours.

### How the #1058 split is made

From the timing signature, against the budget the control plane itself issued (`LOBBY_BUDGET_MS`, the spawn's `waitingRoomTimeout`) — a wait is only an "expiry" relative to a deadline we wrote:

- burned **≥ 90%** of the issued budget → `admission_timeout` (the ~13min Meet population)
- out of the lobby **well inside** it, with no denial reason → `platform_rejection`
- **never saw a lobby at all** → `never_reached_lobby` (the ~20s Teams population)

## Evidence shape

Persisted at `meeting.data.join_evidence`, alongside the new top-level `meeting.data.reason`:

```jsonc
{
  "reason":      "admission_timeout",
  "attribution": "host_action",
  "source":      "bot",              // bot (first-hand) · reconcile · runtime_destroy · derived
  "stage":       "awaiting_admission",
  "detail":      "Bot is still in the Google Meet waiting room after timeout — host did not admit",
  "timings":     { "time_to_lobby_ms": 7000, "time_in_lobby_ms": 780000, "total_ms": 787000 },
  "lobby_budget_ms": 600000
}
```

**An absent timing is absent, never zero** — "nobody measured the lobby" and "zero ms in the lobby" are different facts, and the reconcile sweep (which never saw the platform) emits none rather than inventing them.

The bot classifies **at the source**, where the DOM verdict and the clock actually live (`services/bot/src/join-evidence.ts` + a stopwatch in the join driver). meeting-api validates that verdict and **derives** one when a producer sends none (`lifecycle/join_evidence.py`), so reconcile-driven terminals, runtime-confirmed destroys, and any bot too old to classify are evidenced too. `reached_lobby` is derived from the FSM's own history, not the payload — same FM-003 discipline as `failure_stage`, because the orchestrator historically stamped `awaiting_admission` on every non-admitted verdict including bots that never saw one (also fixed here).

## Fail-open guarantee

Evidence capture describes a run that has **already ended**; no fault in producing it may alter the terminal being recorded. Every entry point on both sides is total, every call site is additionally wrapped, and this is **proven rather than asserted**:

- a driver whose `lastSignals()` throws → identical terminal, identical exit code, conformant event;
- a monkeypatched classifier that raises → the meeting still fails, `data.reason` still lands, callback still 200s;
- a hostile `join_evidence` payload (wrong types, NaN timings, nested junk, `null`, `[]`) → same terminal status, same 200, only the report degrades;
- an unrecognised *reason* from a newer bot is preserved into `detail` rather than dropped; an unrecognised *attribution* is discarded rather than counted, because it feeds the gate.

## Contract

**No re-seal needed.** `join_evidence` rides lifecycle.v1 additively (`additionalProperties: true`) — the same no-seal-bump path `infra_fault` and `stt_fault` already take — and the contract seal hashes only `*.schema.json`. api.v1's `MeetingResponse.data` is already `additionalProperties: true`, so the evidence surfaces on `GET /meetings` and `GET /bots` with no schema change. `gate:contract-version` and `gate:contract-conformance` are green. The field is documented in `contracts/lifecycle.v1/README.md` with a new golden (`LifecycleEvent.failed-join-evidence.json`).

## Tests — all green

| Suite | Result |
|---|---|
| meeting-api `uv run pytest -q` | **963 passed, 5 skipped** (baseline 888/5 → **+75 new**) |
| bot `pnpm --filter @vexa/bot test` | green — `join-evidence.test.ts` alone: **109 passed, 0 failed** |
| `@vexa/join` | 10 passed, 0 failed (unchanged) |
| `pnpm turbo run test` (whole workspace) | **34/34 tasks successful** |
| pre-push gate suite (14 static gates) | green |

Coverage includes a **replayable fixture per failure class** — recorded terminal tapes replayed through the shipped HTTP callback (meeting-api) and through the real orchestrator with fake ports (bot), classifying deterministically with no browser, no meeting, no platform. A meta-test asserts the tape library exercises every typed reason and every attribution value, and a **cross-language parity test** pins the TypeScript and Python vocabularies (and the shared `LOBBY_EXPIRY_FRACTION` / `DETAIL_MAX_CHARS` constants) to the same words, so the two implementations cannot drift silently.

`tests/conftest.py` is untouched — the new fixtures live in `tests/test_join_evidence.py` (avoiding conflict with the parallel #1068 Postgres-fixture work).

## Two pre-existing gaps this surfaced (not fixed here)

1. **A row stuck at `requested` cannot be reconciled at all.** The FSM's only legal first edge is `None → joining`, so the sweep's synthetic `failed` is rejected 409 and the row keeps its status forever — and therefore gets no evidence. Pinned as an explicit test (`test_a_row_stuck_at_requested_still_cannot_be_reconciled_at_all`) so it is not mistaken for a taxonomy bug. Fixing it is an FSM-entry change.
2. **`reconcile.py`'s `body["data"] = {"stop_requested": True}` is a dead channel.** `LifecycleSink.apply_change` never reads `event["data"]`, and `update_meeting_status` merges only `rec.data` — so a sweep-driven completion after a meeting-api restart loses the user-stop marker. That is why this PR routes evidence through a **top-level** `join_evidence` key instead.

## Not verified

Which typed reason the production ~20s Teams cohort actually lands in is **not** verifiable offline. The classification is exercised against both candidate signal shapes (`AdmissionError('denial')` → `platform_rejection`; the bare `no meeting indicators found` throw → `never_reached_lobby`), but the raw platform signal for those 83 rows was never recorded — that is precisely the defect being fixed. Which bucket they fall into will only be known once this ships and rows accumulate. The 90%-of-budget expiry threshold is likewise a reasoned default (production's Teams cohort died at ~3% of budget, Meet's at ~130%), not a value tuned against labelled data.

## Contribution rights


DmitriyG228:
**Delivered — rung: installed and verified in production.**

Merged into main via the 0.12.26 train drain (#1371, one movement for all cars per the one-drain rule), then shipped in estate entry `0.12.23-estate-20260829-seq12` and pinned to production 2026-08-31. Prod conformance re-checked after the pin: CONFORMS 6/6.

Live evidence, meeting 27335 in production this evening — the first prod failure that explains itself:

```
reason:      zoom_requires_rtms — meeting/account blocks automated browser joins
stage:       joining              (never reached a lobby)
attribution: exogenous_platform   (not our software)
source:      bot                  (reported, not inferred)
timings:     failed in 19s of a 900s lobby budget
```

Before this, that meeting would have been an opaque `failed`, indistinguishable from the other nine ExitReasons.

Shipped with three maintainer commits on top: the pre-#1251 `needs_help` classifier hardening, vocabulary alignment with the sealed `MeetingCompletionReason` enum, and the stale brick docstring corrected. One follow-up visible in the evidence above: a pre-lobby policy refusal is labelled `awaiting_admission_rejected` though it never reached admission — a vocabulary nit worth its own fix.

Closing as delivered; the branch's commits are in `main`.

PR #1096, opened 2026-08-09
fix(helm): honor explicit runtime image references and configure minio-init (#1006)

DmitriyG228:
**Delivers issue:** Closes #1006

A cluster that pulls only from a private mirror could not run Vexa no matter what it configured. Three of the images the chart causes a cluster to pull ignored the operator entirely.

Reported by an engineer running the official chart on a quota-controlled, mirror-only OpenShift cluster. In his words, spotting it *"cost me a while before I spotted the Docker Hub path sitting in the render."*

## What was broken

- `runtime.agentImage` and `runtime.agentWorkerImage` were **discarded whenever `global.imageTag` was set** — the chart rendered `vexaai/v012-agent-api:<tag>` and a hardcoded `vexaai/v012-agent-worker:<tag>` instead. Both are *spawned* Pods, so they fail **late**: the control plane goes green and the first meeting dies on `ImagePullBackOff`.
- The `minio-init` hook Job hardcoded `minio/mc:latest` — unconfigurable, and mutable.

## What changed

Every image now resolves by one rule, `vexa.imageRef`:

1. an explicit full reference wins, and **`global.imageTag` never rewrites it**;
2. else `repository@digest`;
3. else `repository:tag`.

Ambiguous identities (a full reference *and* a digest) and malformed digests **fail the render closed** rather than silently resolving to a mutable tag. A typo'd digest must not deploy bytes the operator did not choose.

## Observation bundle

**C1 — the same test, unchanged, is RED at base and GREEN at head.**

At base (`fde7dfb7`), `gate:helm-mirror` exits 1 with 11 failures. The load-bearing ones:

```
AGENT_WORKER_IMAGE = vexaai/v012-agent-worker:v0.0.0-mirror   (explicit reference discarded)
runtime.agentImage discarded
image: minio/mc:latest
+ all six fail-closed cases rendering instead of erroring
```

At head (`ae3bb7ae`): **20 OKs**, and all three A3 negative-control probes go red as required.

**C2 — the byte-diff proves the change is surgical.** Rendering base vs head across four overlays, **exactly one line differs anywhere**: the mc tag. Every other image byte-identical. The `minio/mc` default moved from mutable `latest` to the pinned `RELEASE.2025-08-13T08-35-41Z`, verified against Docker Hub as the same digest — `sha256:a7fe349ef4bd8521fb8497f55c6042871b2ae640607cf99d9bede5e9bdf11727`. **The bytes are unchanged; the identity is now explicit.**

[...]

PR #1093, opened 2026-08-09
fix(runtime): spawned Pods declare CPU/memory requests and limits (#1005)

DmitriyG228:
**Delivers issue:** #1005

Quota-enforcing clusters reject every runtime workload Vexa spawns, because the spawned Pods declare
no CPU/memory requests or limits. Reported by an engineer running the official chart on a
quota-controlled OpenShift cluster. This makes the two runtime profiles — meeting-bot and
agent-worker — declare requests **and** limits, sized independently, exposed as chart values.

It also closes a latent bug the same path carried: **any** k8s spawn with a workspace volume was
silently rejected, because the volume was attached without its container `volumeMounts`.

## Contribution rights

- [x] **Independent:** I created this contribution, or otherwise have the right to submit it
  under Apache-2.0, and it is not owned or controlled by an employer, client, or other entity.
  
- [ ] **Employer/client authorization required:** an employer, client, or other entity owns or
  may control this contribution. I am requesting Vexa's private corporate-authorization process.
  
- [ ] **Unsure:** I need a private rights review before merge.
  

> ⚠️ **Unticked deliberately.** The branch, the validation and this description were prepared by an
> agent; the rights declaration is a legal certification a human must make. @DmitriyG228 — tick one
> box. Both commits already carry a DCO `Signed-off-by`.

## Observation bundle (the record of your harnessed loop)

Full bundle in-tree: [`OBSERVATION-BUNDLE-1005.md`](https://github.com/Vexa-ai/vexa/blob/1005-runtime-resources/OBSERVATION-BUNDLE-1005.md)
(460 lines, every run quoted with its negative control shown red first).

- **C1 — real quota cluster:** ran a k3d cluster (`k3d-vexa1005`, Kubernetes v1.35.5+k3s1) with a
  `ResourceQuota` naming `requests.cpu`, `requests.memory`, `limits.cpu`, `limits.memory` — the
  shape that forces all four to be declared · saw the pre-change spawn rejected by the API server
  with `must specify limits.cpu … requests.memory for: vexa-mtg-red-unsized`, and the same spawn
  admitted `Running`/`Guaranteed` once sized · concluded the reporter's failure reproduces exactly
  through the real `Runtime` + `K8sBackend`, not a mock.
- **C2 — the two profiles are genuinely separate:** ran both profiles in one run, one namespace ·
  saw `1/512Mi` on the bot and `250m/256Mi` on the agent-worker · concluded they read their own
  knobs rather than one shared value.
[...]

Issue #1006, opened 2026-08-05
[DEV] Mirror-only clusters pull every Vexa workload — honor explicit runtime images and configure minio-init


DmitriyG228:
### Value this issue delivers

> A mirror-only operator can select exact meeting-bot, agent, agent-worker and MinIO-init image references, and the rendered/runtime configuration contains no undeclared public-registry fallback.

### Why this matters

Private Kubernetes and OpenShift clusters commonly deny public-registry egress. The chart exposes explicit runtime image values, but a global tag can override two of them and the MinIO init Job hardcodes a mutable public image. An operator therefore needs a patch overlay despite supplying complete mirror references.

### Where we are (honest)

Reproduced with `helm template` at `6a50a2320b797919f54af9a434ae8ecdf35973fe` using `global.imageTag=v0.12.18` and three distinct `harbor.invalid` runtime references:

- `BROWSER_IMAGE` retained the explicit Harbor digest.
- `AGENT_IMAGE` rendered `vexaai/v012-agent-api:v0.12.18`, ignoring `runtime.agentImage`.
- `AGENT_WORKER_IMAGE` rendered hardcoded `vexaai/v012-agent-worker:v0.12.18`, ignoring `runtime.agentWorkerImage`.
- The MinIO init Job rendered `minio/mc:latest`.

Code anchors:

- `deploy/helm/charts/vexa/values.yaml:215-219` exposes the three runtime image values.
- `templates/_helpers.tpl` lets explicit browser image win, but lets the global tag override agent image and hardcodes the agent-worker repository under a global tag.
- `templates/deployment-runtime.yaml` passes helper results into runtime env.
- `templates/job-minio-init.yaml:31` hardcodes `minio/mc:latest`.
- The separate immutable-image issue for long-running components explicitly leaves spawned images to separate coverage; reuse its conventions after that helper work settles.

### Deployments to validate

- **Required:** Helm render fixture with a fake internal registry and no public fallback.
- **Required:** k3s pulling only the rendered exact references.
- **Optional additional:** OpenShift with public registry egress disabled.
- **Lite/Compose:** out of scope unless shared image-resolution code is touched.

### Docs surface

- Helm values reference: deterministic precedence for explicit full references versus global tags/digests.
- Kubernetes deployment guide: inventory every chart-owned and runtime-spawned image required in a private mirror.
- Upgrade note for the precedence change.

### The components

- [ ] **C1 — Explicit runtime workload references win deterministically**
[...]

Issue #1005, opened 2026-08-05
[DEV] Quota-controlled clusters start every runtime workload — apply resources to bot and agent-worker Pods


DmitriyG228:
### Value this issue delivers

> A meeting bot starts in a Kubernetes namespace that requires CPU and memory requests and limits.
>
> An agent worker starts under the same policy with independently configurable sizing.

### Why this matters

The chart deploys a runtime service that creates meeting-bot and agent-worker Pods dynamically. Namespace policies commonly require every container to declare CPU and memory requests and limits. Current runtime Pods omit them even when the accepted workload contract carries resource intent, so those namespaces reject both shipped workload classes.

### Where we are (honest)

Reproduced at `6a50a2320b797919f54af9a434ae8ecdf35973fe`:

- `core/runtime/src/runtime_kernel/models.py:35-52` defines optional `Resources(cpu, memoryMb, gpu)` on `WorkloadSpec`.
- `core/runtime/contracts/runtime.v1/runtime.schema.json:25-33` permits the field; the meeting-bot golden already carries it.
- `core/runtime/src/runtime_kernel/kernel.py:144-185` receives the spec but calls `backend.start(workloadId, runnable, effective_env)`, dropping `spec.resources`.
- `core/runtime/src/runtime_kernel/backend.py` and its implementations expose no resources argument.
- `core/runtime/src/runtime_kernel/k8s_backend.py:126-149` generates `kubectl run` without container resources.
- The meeting and agent producers currently emit no per-workload resources (`bot_spawn/invocation.py:204-235`; `core/agent/shared/adapters.py:300-307`).
- An offline capture test created resource-bearing `meeting-bot` and `agent` specs. Both reached a three-argument backend call, and both generated k8s commands without `resources`, `requests`, or `limits`.
- Current runtime tests remain green (`9 passed, 1 cluster-dependent skip`), demonstrating the missing behavior is not gated.

### Deployments to validate

- **Required:** offline runtime unit/contract tests at PR head.
- **Required:** Helm render with distinct meeting-bot and agent-worker resource values.
- **Required:** real k3s namespace whose ResourceQuota requires both requests and limits; spawn both profiles.
- **Optional additional:** OpenShift restricted namespace using the same quota policy.
- **Compose/Lite:** regression-test the Backend port change. Kubernetes resource-policy behavior is not claimed for those substrates unless separately implemented.

### Docs surface

[...]

PR #1097, opened 2026-08-09
[DEV] validation branch: #1005 + #1006 installable together (DO NOT MERGE)

DmitriyG228:
## What this is

**A validation branch, not a merge candidate.** It exists so one engineer can `git clone`, install, and test the fixes for [#1005](https://github.com/Vexa-ai/vexa/issues/1005) and [#1006](https://github.com/Vexa-ai/vexa/issues/1006) on a quota-controlled, mirror-only cluster. Please do not merge it — merge [#1093](https://github.com/Vexa-ai/vexa/pull/1093) and [#1096](https://github.com/Vexa-ai/vexa/pull/1096) on their own merits.

Composition, off `origin/main` @ `fde7dfb7`:

| Commit | Content |
|---|---|
| `6a5aefb7` | merge `1005-runtime-resources` (= [#1093](https://github.com/Vexa-ai/vexa/pull/1093)) — spawned Pods declare CPU/memory. **Python under `core/runtime/`, so it needs a rebuilt runtime image.** |
| `99668602` | merge `1006-mirror-images` (= [#1096](https://github.com/Vexa-ai/vexa/pull/1096)) — mirror-only image references. **Chart-only; works against published images.** |
| `28d0b330` | validation values file + README, and one new fix found while proving this (below) |

One merge conflict, in `deploy/helm/charts/vexa/README.md`: purely additive — #1005 adds *Sizing the spawned bot and agent-worker Pods*, #1006 adds *Image references*. Both kept, in that order. `values.yaml` auto-merged cleanly (disjoint keys under `runtime:`).

---

## ⚠️ The install blocker we found — it is not either of the two fixes

**The `minio-init` post-install hook Job declared no resources.** On a namespace whose `ResourceQuota` names `requests.cpu`/`limits.memory`, that Pod is rejected. Because it is a *hook* Job, it retries forever, so **`helm install --wait` hangs indefinitely while every Deployment sits happily `Running`** — a failure that looks nothing like a quota problem.

Reproduced live, before the fix, on the first install attempt:

```
Warning FailedCreate job/vexa-vexa-minio-init  Error creating: pods "vexa-vexa-minio-init-t8n9g"
is forbidden: failed quota: require-requests-and-limits: must specify limits.cpu for: minio-init;
limits.memory for: minio-init; requests.cpu for: minio-init; requests.memory for: minio-init
```

…repeated on backoff, release stuck at `STATUS: pending-install`, while all 13 other Pods were `Running`.

Neither #1005 (which sizes *runtime-spawned* Pods) nor the per-component `resources:` blocks cover it, because it is a hook Job that is neither. `28d0b330` adds a `minio.mc.resources` value and wires it into the Job template. After that fix the same install reached `STATUS: deployed` with `vexa-vexa-minio-init` `Completed`.

**This would have hit the customer before either fix got a chance to be evaluated.** It should be fixed on `main` independently of this branch — filed as [#1098](https://github.com/Vexa-ai/vexa/issues/1098).

---

## Observation bundle

Environment: `bbb`, fresh k3d cluster `oenbval`, Kubernetes **v1.35.5+k3s1**, namespace `vexa-quota`.

Quota under test — all four dimensions, which is what forces every container to declare requests *and* limits:

```yaml
spec:
  hard:
    requests.cpu: "16"      limits.cpu: "32"
    requests.memory: 32Gi   limits.memory: 64Gi
```

Runtime image built from this branch on `bbb`, **not pushed to any registry**, side-loaded via `k3d image import`:

```
vexa/v012-runtime:oenb-val-1005-1006
sha256:a83951231a51c452a94ed96eb18bb1aefa5a8c88e893a6f97e7a9e43f99a63e0   (284 MB)
```

| # | Observation | Verdict |
|---|---|---|
| **N1** | Unsized probe Pod **REJECTED** by the quota — run *before* anything else | **RED, as required** |
| **N2** | Zero `LimitRange` objects in the namespace (so nothing is defaulting the pods) | **confirmed, 0 items** |
| **N3** | Negative control **re-run after** the green result — still rejected | **RED, as required** |
| **A0** | `helm install` reaches `STATUS: deployed` under the quota (after the minio-init fix) | **GREEN** |
| **A1** | Deployed runtime is the locally-built image, digest matched | **GREEN** |
| **A2** | Sizing env reached the runtime Deployment | **GREEN** |
| **A3** | meeting-bot Pod admitted, carrying **configured** requests + limits, `Guaranteed` | **GREEN** |
| **A4** | agent-worker Pod admitted, **independently** sized, `Guaranteed` | **GREEN** |
| **A5** | Runtime unit suite on the merged branch | **177 passed, 1 skipped** |
| **A6** | `make -C deploy/helm test` (incl. the mirror-only invariant) on Linux | **PASS** |

### N1 · negative control, run first

```
$ kubectl apply -f - <<< '<pod with no resources>'
Error from server (Forbidden): error when creating "STDIN": pods "probe-unsized" is forbidden:
failed quota: require-requests-and-limits: must specify limits.cpu for: probe; limits.memory for:
probe; requests.cpu for: probe; requests.memory for: probe
```

A sized probe in the same namespace was accepted, and `kubectl get limitrange -o json` reported `0` items — so admission is genuinely driven by the declared resources, not by a namespace default. **N3** re-ran the identical unsized probe *after* the green results and it was rejected again, so the green result cannot be explained by the quota having lapsed.

### A1/A2 · what actually got deployed

```
$ kubectl -n vexa-quota get deploy vexa-vexa-runtime -o jsonpath='{...containers[0].image}'
vexa/v012-runtime:oenb-val-1005-1006

$ kubectl ... containerStatuses[0].imageID
sha256:a83951231a51c452a94ed96eb18bb1aefa5a8c88e893a6f97e7a9e43f99a63e0   ← the image built here

$ kubectl -n vexa-quota exec deploy/vexa-vexa-runtime -- env | grep RUNTIME_
RUNTIME_BACKEND=k8s        POD_NAMESPACE=vexa-quota
RUNTIME_BOT_CPU=1          RUNTIME_BOT_MEMORY_MB=512
RUNTIME_AGENT_WORKER_CPU=0.25   RUNTIME_AGENT_WORKER_MEMORY_MB=256
```

### A3/A4 · the Pods the runtime spawned

Both created through the real runtime HTTP API (`POST /workloads`) against the real `K8sBackend`, which reported `"backend":"k8s"` on both spawns.

```
=== vexa-mtg-v2 ===            (profile meeting-bot, configured cpu 1 / 512 MiB)
  resources : {"limits":   {"cpu": "1",    "memory": "512Mi"},
               "requests": {"cpu": "1",    "memory": "512Mi"}}
  qosClass  : Guaranteed
  labels    : {"runtime.managed": "true", "runtime.workload_id": "mtg-v2"}

=== vexa-agent-v2 ===          (profile agent, configured cpu 0.25 / 256 MiB)
  resources : {"limits":   {"cpu": "250m", "memory": "256Mi"},
               "requests": {"cpu": "250m", "memory": "256Mi"}}
  qosClass  : Guaranteed
  labels    : {"runtime.managed": "true", "runtime.workload_id": "agent-v2"}
```

Both **admitted** by the same quota that rejected the unsized probe seconds earlier and seconds later; both carry the **configured** values, sized **independently** per class; both `Guaranteed`.

---

## Honest limits — read these before trusting the result

1. **Placeholder images.** `BROWSER_IMAGE` and `AGENT_WORKER_IMAGE` were set to `busybox:1.36`, not the real 3.6 GB bot. What is proven is the **resource-declaration and admission path** through the real kernel, real profile registry and real API server — **not** that the real bot boots inside those limits. Sizing the real images for real meetings remains the operator's call. Consequently `vexa-mtg-v2` reached `Succeeded` and `vexa-agent-v2` reached `StartError` (`exec: "python": executable file not found` — busybox has no Python). Those terminal states are artefacts of the placeholder; **admission and the resources block are what this proves**, and both are exactly right.
2. **`imagePullPolicy=IfNotPresent`** was set for this run because the image was side-loaded rather than pushed. The chart default is `Always`; a real mirror-only install keeps `Always` and pulls from its registry.
3. **Nothing was published.** The runtime image exists only on `bbb` and inside the k3d cluster. Per founder decision we publish only once the customer names the exact tag and repository path his Harbor can reach.
4. **Not tested:** OpenShift SCC behaviour on spawned Pods, real Harbor pull-secret paths, the customer's actual quota headroom, and whether the real bot lives inside any given memory ceiling.
5. **`make -C deploy/helm test` fails on macOS** — `tests/test_mirror_only.sh` uses `sed -E 's/^\s*image:...'`, and BSD sed does not support `\s`, so the A2 assertion mis-parses. It passes on Linux. Pre-existing in #1096, unrelated to this branch, but it will bite any maintainer running the gate on a Mac.
6. **One unexplained reading, recorded rather than explained away.** During the teardown/reinstall window (between `helm uninstall` and the clean reinstall) an earlier spawn returned `"backend":"docker"` and failed with `exitCode 1`, creating no Pod, even though the Deployment carried `RUNTIME_BACKEND=k8s`. `WorkloadStatus.backend` is read from the live backend object (`self.backend.name`), so a `DockerBackend` really was in use in whichever process served that request — and `_build_backend()` has no silent fallback path. I could not subsequently identify that process (the Service may have been routing to a terminating Pod of the previous, wedged release). **It did not reproduce on the clean install**, where both spawns reported `k8s` and produced real Pods. Flagging it because a runtime that can end up on the docker backend while its env says `k8s` would be a genuine bug, and I did not root-cause it.
7. **`MARVIN-TEST-KIT-1005.md` and `OBSERVATION-BUNDLE-1005.md` sit in the repo root**, inherited from `1005-runtime-resources`. I deliberately did not delete them (conservative merge), but they should not ship on `main` — flagging for #1093.

---

## For the operator: the trap that produces a green-but-wrong result

The runtime Deployment renders:

```
image: "{{ .Values.runtime.image.repository }}:{{ .Values.global.imageTag | default .Values.runtime.image.tag }}"
```

**`global.imageTag`, when set, silently overrides `runtime.image.tag`.** Setting it to a release tag — the natural move on a mirror-only install, since it pins every other image at once — yields the **old runtime without the #1005 fix**, with no error and no warning. Every Pod goes `Running`, the spawned bots still declare nothing, the quota still rejects them, and the conclusion is "the fix does not work" — having never run it.

Either leave `global.imageTag` unset, or set it to the same tag the runtime was built with. Then verify before trusting any spawn:

```bash
kubectl -n <ns> get deploy <release>-vexa-runtime \
  -o jsonpath='{.spec.template.spec.containers[0].image}{"\n"}'
```

This hole exists because [#1001](https://github.com/Vexa-ai/vexa/issues/1001) (a digest helper for long-running Deployments) is still open. It is called out in a banner at the top of `deploy/helm/values-oenb-validation.yaml` and in the chart README.

---

## Contribution rights

- [ ] I certify that I have the right to submit this contribution

**Left unticked deliberately** — @DmitriyG228 please tick. This box is a legal certification and an agent must never make it on a human's behalf. All three commits carry DCO `Signed-off-by`.

### DIFF
--- a/.github/workflows/release-images.yml
+++ b/.github/workflows/release-images.yml
@@ -181,6 +181,9 @@ jobs:
         run: |
           bash deploy/helm/tests/test_helm_lint.sh
           bash deploy/helm/tests/test_template.sh
+          # #1006 — the mirror-only invariant: no image this chart can cause a cluster to pull may
+          # fall back to a public registry, helm hooks and runtime-spawned Pods included.
+          bash deploy/helm/tests/test_mirror_only.sh
 
       - name: Pinned render — every image ref follows global.imageTag
         env:
--- a/MARVIN-TEST-KIT-1005.md
+++ b/MARVIN-TEST-KIT-1005.md
@@ -1,0 +1,315 @@
+# Test kit — #1005 on a quota-controlled cluster
+
+For the validator running this on a real quota-enforcing namespace (OpenShift dev cluster included).
+Branch `1005-runtime-resources`, based on `616778fe`.
+
+**What you are proving:** the two Pods the runtime creates *dynamically* — the meeting bot and the
+agent worker — now declare CPU and memory requests **and** limits, so your namespace admits them
+instead of rejecting them. Nothing else about the release changes.
+
+**What has already been proven, and where** — so you can skip re-doing it:
+
+| Already green | Where |
+|---|---|
+| Both profiles admitted under a `ResourceQuota` requiring all four values, sized independently | k3d v1.35.5, namespace `vexa-quota` — see `OBSERVATION-BUNDLE-1005.md` A1/A2 |
+| Image, command, env, labels, tolerations, nodeSelector, workspace mounts all survive | same, A3 |
+| Removing one dimension reds admission; restoring it greens the same spawn | same, A4 |
+| Offline runtime suite, Helm render, Compose stack, all 34 repo gates | same, A- |
+
+**What only your cluster can prove:** OpenShift SCC behaviour on the spawned Pods, your real quota
+headroom, and whether the real bot image lives inside the memory ceiling you set. Read
+[the honest caveats](#honest-caveats) before you start — one of them is likely to bite.
+
+---
+
+## 1 · Install
+
+Two values blocks matter. The first is the new one.
+
+```yaml
+# values-quota.yaml
+runtime:
+  backend: k8s
+  # Sizing for the Pods the runtime SPAWNS. NOT runtime.resources (which sizes the runtime
+  # Deployment itself). One value per dimension sets BOTH the request and the limit.
+  workloadResources:
+    meetingBot:
+      cpu: 2                 # → requests/limits cpu: 2000m
+      memoryMb: 4096         # → requests/limits memory: 4096Mi
+    agentWorker:
+      cpu: 0.5               # → requests/limits cpu: 500m
+      memoryMb: 1024         # → requests/limits memory: 1024Mi
+```
+
+```bash
+helm upgrade --install vexa deploy/helm/charts/vexa \
+  -n <your-namespace> --create-namespace \
+  -f values-quota.yaml \
+  --set global.imageTag=<TAG> \
+  --set secrets.adminApiToken=$ADMIN_TOKEN \
+  --set secrets.internalApiSecret=$INTERNAL_API_SECRET
+```
+
+Confirm the sizing reached the runtime before you spawn anything:
+
+```bash
+kubectl -n <your-namespace> get deploy vexa-runtime -o json \
+  | jq '.spec.template.spec.containers[0].env[]
+        | select(.name|test("^RUNTIME_(BOT|AGENT_WORKER)_"))'
+```
+
+Expected — four entries, non-empty:
+
+```json
+{"name":"RUNTIME_BOT_CPU","value":"2"}
+{"name":"RUNTIME_BOT_MEMORY_MB","value":"4096"}
+{"name":"RUNTIME_AGENT_WORKER_CPU","value":"0.5"}
+{"name":"RUNTIME_AGENT_WORKER_MEMORY_MB","value":"1024"}
+```
+
+If they are empty strings, the chart values did not land — the spawned Pods will declare nothing and
+your namespace will reject them exactly as before. Fix this before going further.
+
+> **Sizing note, not a formality.** `memoryMb` is a **hard ceiling**: a workload exceeding it is
+> OOM-killed. Before this change nothing was enforced, so bots ran unbounded and we have no measured
+> figure to hand you. `4096` above is deliberately more generous than the chart default (`2048`).
+> Start high, watch actual usage, then tighten — do not start at the default and discover the ceiling
+> as a dropped meeting.
+
+---
+
+## 2 · Spawn both profiles and watch
+
+Trigger one meeting bot (your normal `POST /bots` path) and one agent dispatch. Then, for each:
+
+```bash
+kubectl -n <your-namespace> get pods -l runtime.managed=true
+kubectl -n <your-namespace> get pod <pod> -o json | jq '{
+  qos:        .status.qosClass,
+  phase:      .status.phase,
+  resources:  .spec.containers[0].resources,
+  image:      .spec.containers[0].image,
+  command:    .spec.containers[0].command,
+  envCount:   (.spec.containers[0].env|length),
+  labels:     .metadata.labels
+}'
+```
+
+### Expected observations
+
+| Pod | Name shape | Expect |
+|---|---|---|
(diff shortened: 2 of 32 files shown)


DmitriyG228:
Closing this vehicle — it is explicitly DO-NOT-MERGE and its purpose is spent. Its cargo did not all land, so the open half is named below rather than closed with it.

**Landed:** #1005 via #1093 (`12769637`) — `resource_requirements()` in `core/runtime/src/runtime_kernel/k8s_backend.py` with `RUNTIME_BOT_CPU` / `RUNTIME_BOT_MEMORY_MB` / `RUNTIME_AGENT_WORKER_*` declared in `config.v1.json`. Released in **v0.12.24**, installed in production.

**Landed separately:** the install-hang this branch found (#1098) — `f6d43242` (#1321) added `minio.initResources` and made the `mc` image a value. Released in **v0.12.25**.

**Not landed:** #1006. `vexa.agentImage` / `vexa.agentWorkerImage` in `_helpers.tpl` still discard an explicit reference whenever `global.imageTag` is set. That stays open on #1006 / #1096.

The validation values file (`deploy/helm/values-oenb-validation.yaml`) was never intended for main.

PR #1114, opened 2026-08-10
fix(gates): report the failure instead of swallowing it (#1107)

### LINKED ISSUE #1107: gate:schema swallows every failure message (empty Buffer is truthy) and blocks pre-push from a fresh worktree
### 1. The gate reports nothing when it fails

`scripts/gates.mjs`, `gateSchema()`:

```js
catch (e) { return fail([`schema ${rel(d)}:\n${(e.stdout || e.stderr || e).toString()}`]); }
```

`execSync` with `stdio: "pipe"` sets `e.stdout` to a **zero-length Buffer** when the child wrote only to stderr. An empty Buffer is an object, and **every object is truthy in JS**, so `e.stdout || e.stderr` short-circuits on the empty stdout and the real diagnostic is discarded.

Reproduction, no repo needed:

```
e.stdout is Buffer len 0 | truthy? true
(e.stdout || e.stderr).toString() => ""
correct form                     => "boom"
```

So the operator sees exactly this, and nothing else:

```
▶ gates: schema

❌ gates failed
✗ pre-push: gate:schema FAILED — push aborted
```

A gate that cannot say why it failed is worse than one that does not run: the operator's only remaining move is `--no-verify`, which is precisely the habit the hook exists to prevent.

Suggested fix, one line:

```js
catch (e) {
  const out = e.stdout?.length ? e.stdout : (e.stderr?.length ? e.stderr : e);
  return fail([`schema ${rel(d)}:\n${out.toString()}`]);
}
```

### 2. Underneath it: the gate fails on clean `main` in a fresh worktree

With the message restored, the actual error is:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'ajv'
  imported from core/agent/contracts/event.v1/validate.mjs
```

Reproduced on a **detached worktree of `origin/main` with no local changes** — `node scripts/gates.mjs schema` fails. It is an environment failure (contract validators need `ajv`, which a fresh checkout has not installed), not a schema failure.

The consequence is that **`git push` is blocked from any clean worktree** until someone happens to run an install, and the hook gives no hint that this is what it wants. Worth either failing with an actionable message ("run `npm install` — the contract validators need `ajv`") or skipping the gate with a stated reason when the dependency is absent, so the signal stays honest.

### Why CI did not catch it

CI's `gates` context is green on current PRs, so **CI and the local pre-push hook disagree about the same named gate.** Whatever CI runs either installs deps first or does not include this leg. That divergence is worth closing on its own — a local gate stricter than CI, and mute about why, trains contributors to bypass it.

Found while pushing a docs-only branch during a `/ring` revolution; the gate failure had nothing to do with the diff.

DmitriyG228:
Closes #1107.

Every gate in `gates.mjs` reported its own failures through:

```js
(e.stdout || e.stderr || e).toString()
```

Under `stdio: "pipe"`, a child that writes only to stderr leaves `e.stdout` a **zero-length Buffer**. An empty Buffer is an object, so it is **truthy**, so the `||` chain short-circuits on it and returns `""`. The operator sees the gate's name, a colon, and nothing:

```
▶ gates: schema

❌ gates failed
✗ pre-push: gate:schema FAILED — push aborted (fix it, or bypass with: git push --no-verify)
```

A gate that cannot say why it failed is worse than one that does not run: the only move left is the `--no-verify` the hook itself suggests, which is precisely the habit these hooks exist to prevent.

Found when `gate:schema` aborted a **docs-only** push. The real cause was a missing `ajv` in a fresh worktree — an actionable, one-command fix, invisible behind the empty string.

## The change

One `errText()` helper replacing **all 21 call sites**. It chooses on **length**, never truthiness: prefer stdout, fall back to stderr, then the error itself. Existing `.slice()` calls are untouched — the helper returns a string.

## Test

`scripts/gate-error-text.test.mjs` drives a **real `execSync` failure** rather than a hand-built object, because the defect lives in what Node actually attaches to the thrown error. A fixture asserting against `{stdout: Buffer.alloc(0), stderr: Buffer.from("boom")}` would pass against a wrong implementation that checked `!= null` instead of `.length`.

It also asserts the *precondition* — that an empty Buffer is truthy and the old expression returns `""` — so the test still means something if Node ever changes what it attaches, and guards that the pattern does not return anywhere in the file.

**Red-before/green-after:** with the old expression, **3 of 5 fail**; with the fix, **5 of 5 pass**.

`gate:readme`, `gate:schema` and `gate:exports` all still green after the change.

---

## Update 2026-08-21 — the second defect, and a correction

Rebased onto `main` (`6e5c344f`); the first commit is unchanged. Two additions.

### 1. The restored message still hid the commonest cause

Restoring the diagnostic was half of #1107. What a dependency-less worktree actually produces is a Node module-resolution stack trace — it says what broke, not that the fix is one command — so the tempting move is still the `--no-verify` these hooks exist to prevent.

The hint lives in `fail()`, the single point every gate's errors print through. Two reasons, both load-bearing: it is appended **after** each caller's `.slice(…)`, so the one actionable line can never be the part that gets truncated; and no individual gate has to remember to say it. All 21 call sites inherit it.

It names **`pnpm`**, verified rather than assumed — `package.json` pins `packageManager: pnpm@11.7.0`, the repo carries only `pnpm-lock.yaml`, CI runs `pnpm install --frozen-lockfile`, and `gateLicenses` already says ``run `pnpm install` first`` on its own non-JSON path. A test asserts the hint string against `package.json` instead of hard-coding it twice, so the hint cannot silently outlive a package-manager change.

### 2. Correction: 21 call sites, not 22

The earlier figure counted the comment's own mention of `` `errText(e)` ``. Changelog fragment corrected. `gateLicenses`'s `(e.stdout || "")` is deliberately **not** converted — it is a raw-capture with `stderr` set to `ignore`, not an error report.

### Validation

Fresh worktree, no `node_modules`, against `origin/main`:

```
before   ✗ schema core/agent/contracts/event.v1:
         (nothing)

after    ✗ schema core/agent/contracts/event.v1:
         Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'ajv' imported from …/event.v1/validate.mjs
         → hint: did you run `pnpm install` in this worktree? A fresh worktree has no node_modules.
```

The hook proved it on itself: pushing this branch from the dependency-less worktree aborted at `gate:schema` and printed the hint, which named the fix. Installing deps turned all 15 fast gates green and the push through.

- **8/8** tests pass. Removing the hint turns exactly the two new ones red; the pre-existing five stay green.
- `gate:schema` green **with** deps (25 contracts) — success path untouched.
- All 15 pre-push fast gates green locally.

### Overlap with #1129

[#1129](https://github.com/Vexa-ai/vexa/pull/1129) (@Msparihar) targets the same issue and reached the same root cause independently. It is narrower — `gateSchema()` only, so the other 20 sites stay silent — and its hint reads `npm install`, which is not this repo's package manager. The hint here is the same idea with the package manager verified and applied at the shared choke point. **That PR has been open since 2026-08-11 without a maintainer response; it deserves one either way, and the credit for surfacing the hint idea is theirs.**

## COMMITMENTS IN THIS DRAFT

- **Money out** — none.
- **Money in** — none.
- **Work we will do** — none. No follow-up is promised to anyone here.
- **A date** — none.
- **A right** — none.
- **Published terms** — none. The hint text is operator-facing console output, not a documented interface.
- **A precedent** — one, worth a deliberate yes/no: this puts an **advice line into `fail()`, which every gate prints through**. It is the first hint of its kind there, so the next one will be added by pointing at this. If you would rather hints stayed local to the gate that needs them, say so and it moves into `gateSchema()`.

Rights box left for the founder — agents do not sign attestations.

## Contribution rights
- [ ] **Employer/client authorization required:** an employer, client, or other entity owns or may control this contribution.
  

DmitriyG228:
### Founder's reproduction verified against this head

Run verbatim on both sides — fresh worktree, no `node_modules`.

**`origin/main` (`6e5c344f`)** — `node scripts/gates.mjs schema`:

```
▶ gates: schema


❌ gates failed
```

Nothing after the colon. The real error, visible only by running the validator directly:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'ajv' imported from
  /private/tmp/wt-test/core/agent/contracts/event.v1/validate.mjs
```

**This head (`a74ead70`)** — same command, same conditions:

```
▶ gates: schema
node:internal/modules/package_json_reader:316
  throw new ERR_MODULE_NOT_FOUND(packageName, fileURLToPath(base), null);
        ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'ajv' imported from
  /private/tmp/wt-1114/core/agent/contracts/event.v1/validate.mjs
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:316:9)
    at packageResolve (node:internal/modules/esm/resolve:768:81)
    ... 
  code: 'ERR_MODULE_NOT_FOUND'
}

Node.js v24.13.0

  → hint: did you run `pnpm install` in this worktree? A fresh worktree has no node_modules.

❌ gates failed
```

The cause is named and the fix is one command. This covers the reproduction.

**Also checked on this head**

- `e.stdout || e.stderr` in `scripts/gates.mjs`: **0 remaining**. All 21 call sites go through `errText()`, which selects on `.length`, never on truthiness — the empty-Buffer-is-truthy trap that caused this.
- No other file in the tree carries the pattern; the only remaining textual hits are the regression assertions inside `scripts/gate-error-text.test.mjs`.
- `node --test scripts/gate-error-text.test.mjs` in the fresh worktree: **8/8 pass**, with no dependencies installed.
- The hint names `pnpm`, and one of those tests asserts that against this repo's `package.json`.

**CI** — everything machine-checkable is green: `gates`, `node`, `python`, `static`, CodeQL ×3, DCO, `docs-current`, GitGuardian. The three reds are the human gates and nothing else: `contribution-rights` (box unticked — agents do not sign attestations) and `merge-card`, which fails only its **Value** row (`missing state: value-signed`); its Diff and Acceptance rows are ✅.

No changes made to this branch — verification only.

Issue #1119, opened 2026-08-11
Teams Speakers aren't seperated and attributed


danieldormann:
### What happened

Every Teams turn is labelled and "Speaker" and grouped as one. Names are never extracted and speakers never seperated. I'll provide a fix for this.

### Meeting platform (if relevant)

MS Teams

### Deployment — observed on which version / verified against which tag?

Full compose, image tag v012, repo sha 4c1612d8

### Logs, screenshots, meeting shape

Two people. German. Started via web interface. 

  <img width="809" height="291" alt="Image" src="https://github.com/user-attachments/assets/bc283b4e-2c6e-41da-878f-80ee4995814c" />

danieldormann:
I have a fix and live validation already; will open a PR shortly.

DmitriyG228:
[`#1024`](https://github.com/Vexa-ai/vexa/pull/1024) (open since 2026-08-06) also resolves Teams
speaker names, touching the same three `teams-capture` files — but from `data-tid` on the
`[data-stream-type]` wrapper, which it reports as stable across Fluent class-hash churn.
[`#1121`](https://github.com/Vexa-ai/vexa/pull/1121) reports no `data-tid` to key on and reads leaf
text instead.

You observed on sha `4c1612d8`, which does not include #1024 — so both readings may hold on
different code paths. #1024 returns `''` and falls back to legacy selectors when no stream wrapper
is found, which is the case #1121 handles.

Flagging before either merges; they overlap on all three files.

DmitriyG228:
Confirmed independently: on published v012 we measured three live Teams meetings at 0 named segments of 76 — the voice-level detection fires, but the name resolver returns empty (the class hash rotted again: `___2u340f0` → `___12zni01`) and the no-blank-name gate silently drops every hint.

Your structural fix in #1121 is the right direction and is being incorporated, with authorship, into a broader Teams-attribution overhaul now in flight (structural name paths, fail-loud diagnostics when resolution misses, caption capture as a second attribution source). This issue closes via #1121's merge; the deeper items are tracked in #797.

DmitriyG228:
Delivered in **[v0.12.22](https://github.com/Vexa-ai/vexa/releases/tag/v0.12.22)**, published today.

[...]

Issue #868, opened 2026-07-21
mixed binder publishes seg_N while a live hint names the turn — 30.7% of words unnamed on a real Zoom session (Zoom/Teams/YouTube shared)


DmitriyG228:
## Value this change must deliver

> **A turn whose speaker the platform has already named never publishes under a provisional `seg_N`: the mixed-lane binder uses the name it was handed, on every platform that feeds it (Zoom, Teams, YouTube), proven by replay on a recorded real session.**

## Where we are (honest, measured)

Surfaced under #852 while validating the Zoom layout fix (`19d84597`). Replaying the fixed live Zoom capture (botsig9, 162.8s, 4 speakers, 139 DOM hints) through the real lane:

- **Real-STT replay (prior session): 27 of 66 turns — 30.7% of words — published as `seg_N`.** All 27 unnamed turns overlapped a live `dom-active` hint (continuous coverage); 23/27 with a single unambiguous name in the hint stream. Not signal absence — a binder gate rejected an available name. (Verified twice, self + workflow.)
- **`seg_N` is a per-turn counter** (`chunked-transcriber.ts`), not a cluster id; this lane does no speaker clustering. Every unnamed turn is a user-visible unnamed row.
- **New instrument** (binder `explainMatch` + `[binder-reject]` narration, staged on `quality/0.12.17-base` + working tree): under the mock-STT arm, all 26 resolve-time rejects fire on the **confidence gate** (`MIN_MATCH_CONFIDENCE=0.6`), every one a 2-candidate contest with the best name holding 0.35–0.59 of total support — the ±2.5s match tolerance pulls the lag-shifted neighbouring speaker into every post-handover window, so a moment the raw hint stream calls unambiguous is contested inside the binder. Caveat: the mock arm reproduces the reject *reason* but not the 30.7% magnitude (3.7% of words); the real-STT arm re-run is in flight and the verdict binds to it.
- Refuted along the way: the open-turn grace (`OPEN_TURN_GRACE_MS`) as the source of the phantom competitor — closing other names' open turns on each `dom-active` assertion changed nothing (competing support comes from *closed* heartbeat slices inside the tolerance window).
- The handoff's structural suspicion stands unresolved alongside the gate finding: repaint fires only on an incoming hint — **no timer sweep** — so a name that arrived just before the hint stream pauses is never revisited.

## Components

- `core/meetings/modules/mixed-pipeline/src/cluster-name-binder.ts` — `windowMatch` gates; owner of the fix.
[...]

PR #1024, opened 2026-08-06
fix(teams): resolve speaker names from stable data-tid (attribution restored)

jbschooley:
**Delivers issue:** _(none — maintainer's own work)_

Restores MS Teams speaker attribution and fixes Teams transcription quality. Three fixes, all **live-witnessed** on today's Teams web client:

- **Name extraction (Fluent-v9 rot).** Teams' hashed class names churn every release; the old name selectors resolved to `""`, so every hint was suppressed and all speech fell back to `Speaker`. Names now resolve from Teams' **stable** `data-tid` on the `[data-stream-type]` tile wrapper, anchored on the voice-level outline. (`vdi-frame-occlusion` speaking detection was correct all along — the stale `#speaker-split` DIAG claiming otherwise is removed.)
- **Transcription doubled words.** Teams delivers the whole meeting as ONE server-side mix (track id `mainAudio`) but hands the bot a redundant second track whose audio is already in that mix; the mixed lane combined both → every word double-fed to the transcriber → repeated words. Teams now transcribes **only** the `mainAudio` mix.
- **Double-mirror.** The WebRTC hook mirrored each remote track twice (both `addEventListener('track')` and the `ontrack` setter fired `handleTrack`) — now deduped by track id.

## Contribution rights
- [x] **Independent:** I created this contribution, or otherwise have the right to submit it under Apache-2.0, and it is not owned or controlled by an employer, client, or other entity. 
- [ ] **Employer/client authorization required:** an employer, client, or other entity owns or may control this contribution. 
- [ ] **Unsure:** I need a private rights review before merge. 

## Observation bundle
- **C1 — root cause witnessed live —** inspected the live Teams web DOM (browser console) in an active meeting: display names live in `data-tid` on the `[data-stream-type]` wrapper; `vdi-frame-occlusion` lights on the speaker; and the client holds **exactly one** audio receiver (`mainAudio`) even while a specific person speaks → Teams is a single server mix, and per-track attribution is not feasible.
- **C2 — unit tests (green) —** `teams-capture.test.ts`: name resolves from `data-tid` on `[data-stream-type]`; anchors on the stream nearest the voice outline (not DOM order); returns `''` (→ legacy fallback) when absent.
[...]

PR #1121, opened 2026-08-11
fix(teams-capture): resolve speaker names structurally, not by obfusc…

danieldormann:
**Delivers issue:** #1119 

## Contribution rights

Every commit carries my DCO `Signed-off-by`.

- **C1 — the symptom, and where it is NOT.** ran: three live Teams meetings on the published
  `v012` bot; queried `transcriptions`. saw: 76 segments, **0** carrying a speaker name — every
  turn labelled `Speaker`, one distinct value. Bot logs showed
  `[TeamsSpeakers] Scanned 4 participants, observing 1 with signal` — so detection was alive —
  alongside `hint-counters bridge-crossed=0 pipeline-received=0 binder-matched=0 binder-missed=0`.
  concluded: hints are not merely rejected downstream, they are never emitted. `bridge-crossed`
  increments as the *first* statement of the hint sink, so 0 means the sink was never called.

- **C2 — the gate.** ran: read `msteams-speakers.ts` `emit()` and `extractName()`. saw:
  `if (!identity.name) return;` — every hint is gated on a resolved name. A `Removed:  (13f2…)`
  log line carried an **empty** name. concluded: `extractName()` returns `''`; everything
  downstream is blocked by that one value. The failure is silent and total.

- **C3 — ground truth, not guesswork.** ran: added a temporary dump to `getIdentity()` logging the
  first three unnamed tiles' attributes and text-carrying descendants; rebuilt the bot; ran one
  meeting. saw: the roster tile carries **no `data-tid`, no `title`, no `aria-label`**, and the
  name div's only marker is an obfuscated build class — `___12zni01`, where
  `teamsNameSelectors[0]` hardcodes `___2u340f0`. concluded: selector rot. Writing a new hardcoded
  hash would only reset the rot timer; the fix has to be structural.

- **C4 — the fix.** ran: added `plausibleNameFromLeaves()` — leaf-text scan rejecting control
  labels, timestamps and letterless fragments — mirroring the heuristic fallback the chat reader in
  this same module already uses (`teams-chat.ts:109`). Hoisted it to module scope and exported it
  so the name path is unit-testable for the first time. saw: typecheck clean, 16/16 tests green.
  concluded: the existing selector list still runs first and is unchanged; the fallback only
  engages when it misses.

- **C5 — live witness.** ran: rebuilt the bot from this branch, ran a live Teams meeting with two
  German-speaking participants. saw: **13/13 confirmed segments correctly attributed by name**,
  two distinct speakers, against 0 of 76 before. concluded: value is real at the altitude of the
  claim (D12 — speaker behaviour needs a live multi-speaker meeting).

## Acceptance floor

| Row | Evidence |
|---|---|
| **A1 — red→green on the live DOM** | RED: the pre-fix selector-only path against the captured tile returns `""`. GREEN: `✅ name: live 2026-08 roster tile resolves despite the rotted class`. Base sha `4c1612d8`. |
| **A2 — rot-resistance (the actual defect class)** | `✅ name: survives a further class rotation` — fixture uses a deliberately invented future class (`___future0`), so the next Teams build cannot regress this the way `___2u340f0` did. |
| **A3 — negative control, shown red** | `✅ name: unresolvable stays empty (no hint is better than a wrong hint)` — a tile with no name-shaped text still yields `''`, so `emit()` suppresses rather than inventing a GUID. No green-on-empty. |
| **A4 — no false positives from tile furniture** | `✅ name: skips control labels, timestamps and bare digits` and `✅ name: leaf wins over concatenated wrapper text`. |
| **A5 — no regression** | 16/16 tests pass; the 11 pre-existing chat/selector tests are untouched and green. `tsc --noEmit` clean. |
| **A6 — live, multi-speaker** | 13/13 confirmed segments named across 2 speakers on a real Teams call; 0/76 on the same stack before the change. |

## Docs diff (D6c)

No docs change. This is an internal DOM-extraction fallback inside `@vexa/teams-capture` with no
reader-facing surface: no new configuration, no API field, no changed behaviour to teach. The
user-visible effect is that documented behaviour (named speakers) starts working again. If
reviewers would rather see the rot hazard recorded for operators, the natural home is a short note
in the Teams section of the bots docs — happy to add it.

- **Dependencies:** none added, removed or upgraded. The module still has zero runtime
  dependencies (`package.json` unchanged) — nothing for a licence scan to flag (P17).
- **Secrets:** none introduced. Test fixtures use invented names (`Priya Nair`, `Tariq Baig`),
  matching the fictional names already used by this file's chat tests. No captured meeting content,
  no real participant names, no hostnames.
- **SAST/injection surface:** the new code is pure string inspection of DOM text — no `eval`, no
  `innerHTML`, no network, no filesystem, no dynamic selector construction. It reads
  `textContent` and returns a string. Runs in the browser context with no new privileges.
- **Input handling:** length-bounded (2–50 chars) and character-class filtered, so hostile page
  text cannot produce an unbounded or control-character speaker label.

## Validation request

**Who:** any competent non-author; the originating reporter preferred. Needs a Microsoft Teams
account and one other participant.

**What they'll watch:** join a Teams meeting with 2+ people who each speak, then confirm the
transcript shows **distinct participant names** rather than a single `Speaker`. Corroborating
instrument channels: `hint-counters bridge-crossed` > 0 in the bot log (it was 0 before), and
`select speaker, count(*) from transcriptions where meeting_id = <id> group by 1` showing more
than one distinct value.

**Deployment validated (D12b):** full Docker Compose stack (`deploy/compose`), published images at
tag `v012`, with the meeting bot rebuilt from this branch via `make -C deploy/compose bot`
(`vexa/vexa-bot:dev`). Long-lived install, base sha `4c1612d8`. Env deltas from stock: Microsoft
OAuth configured, MinIO host ports remapped off a portainer collision, self-hosted
`deploy/transcription` unit (`large-v3`, `int8`) reached over the compose network.

**Not validated:** Lite, k8s/helm and hosted. This change is browser-side DOM extraction with no
deployment-specific path, so I expect no divergence — but nobody ran them, so they stay honestly
unclaimed.

## Authorship

Tooling disclosure (optional, welcome, never an attribution): diagnosis, patch drafting and the
test fixtures were produced in a Claude Code session; every claim above was checked against the
running stack and the numbers come from its database.

---

## Known and deliberately out of scope

Pending drafts are persisted with their own segment ids (`turn:44:p0` beside the confirmed
`turn:44:0`), and a draft emitted before its hint binds keeps the `Speaker` fallback — so a UI
rendering both shows the same utterance twice, once mislabelled. In the validation meeting this was
1 of 24 segments. `pipeline.ts:146` says "late attribution still repaints by segment_id", but the
draft and the confirmed segment have *different* ids, so nothing supersedes the draft.

This is independent of selector rot, lives in the mixed lane's publish/persist path rather than in
`teams-capture`, and would affect any platform on that lane. It is **not** the same as #868, which
concerns valid names being *rejected at binding* by a confidence gate; this is a draft published
*before* a name exists. Happy to file it separately.

### DIFF
--- a/core/meetings/modules/teams-capture/src/index.ts
+++ b/core/meetings/modules/teams-capture/src/index.ts
@@ -13,6 +13,8 @@ export {
   teamsNameSelectors,
   teamsParticipantIdSelectors,
   teamsMeetingContainerSelectors,
+  plausibleNameFromLeaves,
+  TEAMS_NAME_FORBIDDEN,
 } from './msteams-speakers.js';
 export type { TeamsSpeakers, TeamsSpeakersOptions, TeamsSpeakerIdentity } from './msteams-speakers.js';
 export { createTeamsChat } from './teams-chat.js';
--- a/core/meetings/modules/teams-capture/src/msteams-speakers.ts
+++ b/core/meetings/modules/teams-capture/src/msteams-speakers.ts
@@ -61,6 +61,57 @@ export const teamsMeetingContainerSelectors: string[] = [
 
 const VOICE_LEVEL_SELECTOR = '[data-tid="voice-level-stream-outline"]';
 
+/** Control-affordance labels that share a tile with the display name and must never be
+ *  mistaken for it. Matched as case-insensitive substrings. */
+export const TEAMS_NAME_FORBIDDEN: string[] = [
+  'more_vert', 'mic_off', 'mic', 'videocam', 'videocam_off',
+  'present_to_all', 'devices', 'speaker', 'speakers', 'microphone',
+  'camera', 'camera_off', 'share', 'chat', 'participant', 'user',
+];
+
+/** First leaf-text fragment in `element` that can plausibly be a display name.
+ *  
+ *  Teams renders the display name in a div whose only distinguishing
+ *  attribute is a build-specific obfuscated class like '___2u340f0'with no data-tid, title or 
+ *  aria-label to key on.
+ *  
+ *  Leaves only ('childElementCount === 0'), so a wrapper's concatenated text can never win.
+ *  De-duplicated, because Teams nests the same name twice. Rejects control labels, bare
+ *  timestamps, and any fragment without a letter in it. 
+ */
+export function plausibleNameFromLeaves(
+  element: { querySelectorAll(sel: string): ArrayLike<{ childElementCount: number; textContent: string | null }> },
+  forbidden: string[] = TEAMS_NAME_FORBIDDEN,
+): string {
+  const seen = new Set<string>();
+  const leaves = element.querySelectorAll('*');
+
+  for (let i = 0; i < leaves.length; i++) {
+    const el = leaves[i];
+
+    // skip non-leaf elements
+    if (el.childElementCount !== 0) continue; 
+
+    const text = (el.textContent || '').trim();
+
+    // skip duplicate text
+    if (!text || seen.has(text)) continue; 
+
+    seen.add(text);
+
+    // skip text that probably isn't a name
+    if (text.length < 2 || text.length > 50) continue;
+    if (!/\p{L}/u.test(text)) continue;                       // digits/glyphs only → not a name
+    if (/^\d{1,2}:\d{2}/.test(text)) continue;                // "10:42 AM"
+
+    // skip text that contains forbidden substrings
+    const lower = text.toLowerCase();
+    if (forbidden.some((sub) => lower.includes(sub.toLowerCase()))) continue;
+    return text;
+  }
+  return '';   // name not resolvable yet — emit NO hint rather than a meaningless GUID
+}
+
 export interface TeamsSpeakerIdentity {
   id: string;
   name: string;
@@ -120,11 +171,7 @@ export function createTeamsSpeakers(opts: TeamsSpeakersOptions): TeamsSpeakers {
   }
 
   function extractName(element: HTMLElement): string {
-    const forbidden = [
-      'more_vert', 'mic_off', 'mic', 'videocam', 'videocam_off',
-      'present_to_all', 'devices', 'speaker', 'speakers', 'microphone',
-      'camera', 'camera_off', 'share', 'chat', 'participant', 'user',
-    ];
+    const forbidden = TEAMS_NAME_FORBIDDEN;
     for (const selector of teamsNameSelectors) {
       const nameElement = element.querySelector(selector) as HTMLElement | null;
       if (!nameElement) continue;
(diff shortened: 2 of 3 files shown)


DmitriyG228:
Excellent diagnosis and an exemplary observation bundle — C1's `bridge-crossed=0` and C3's live DOM dump match what we measured independently: detection alive, `extractName()` empty, everything suppressed by the no-blank-name gate.

We're preparing a larger Teams-attribution branch (structural name resolution incl. a stable `data-tid` path, fail-loud diagnostics, caption capture). Your commits are being cherry-picked into it with authorship and DCO intact. Two things we're adopting directly:

- **A2, the rot-resistance fixture** (`___future0`) — it now guards our resolution path too.
- Your guard checks, folded into our shared name-candidate filter, credited to this PR.

Review for main to follow shortly.

DmitriyG228:
Absorption done — your commit rides our branch with authorship and sign-off intact. Two things it did for us:

Your A4 fixture caught a real defect: our guard only rejected a bare `10:42`, so `10:42 AM` in a roster tile could still have become a speaker name. Your check is now in the shared guard — the caption reader inherits it too. And your A2 fixture (the invented `___future0` class) is now the test guarding our resolver through the next Fluent rotation — the piece we didn't have.

On the code: the branch already carries a leaf-scan fallback under a richer chain (stable `data-tid` on the stream wrapper → ARIA → leaves), so `plausibleNameFromLeaves` became the exported entry point to that scan — your five tests run verbatim against it, all green. One deliberate change: control labels match as whole tokens, not substrings — `mic` is inside *Michael* and *Mica*, `camera` inside *Camera*.

DmitriyG228:
Two things blocking this, one of them one click.

**`docs-current` wants a docs decision, not docs.** The gate says:

> *"touches a product surface but changes no `docs/**` and lacks the `docs: none` label — update the
> docs, or add `docs: none` (with a reason) to confirm none is needed (D6c)"*

Speaker-name resolution is internal behaviour with no user-facing surface change, so `docs: none`
with that as the reason is very likely the right answer — but the label is a maintainer's call, not
something an agent should apply to your PR.

**`merge-card` wants `state: value-signed`**, which is the maintainer value sign-off. Also theirs.

**One thing worth knowing before it is reviewed.** [`#1024`](https://github.com/Vexa-ai/vexa/pull/1024)
touches the same three `teams-capture` files and resolves the same problem from the opposite premise —
it reads the name from `data-tid` on the `[data-stream-type]` wrapper and calls that stable across
Fluent class churn, where this PR reports no `data-tid` to key on. You observed on sha `4c1612d8`,
which does not include #1024, so both readings may hold on different code paths. Flagged on
[`#1119`](https://github.com/Vexa-ai/vexa/issues/1119) so whoever merges sees it first.

For what it is worth to that ordering: this PR is **3 files and current**; #1024 is **61 files and
behind main**.

DmitriyG228:
Withdrawing my previous comment — it is wrong and I should not have posted it.

Two comments above it, from earlier today, already say **"Absorption done — your commit rides our
branch with authorship and sign-off intact."** So this PR's work is taken, and the `docs: none` and
`state: value-signed` I described as blockers are not things anyone needs to chase here. The
[`#1024`](https://github.com/Vexa-ai/vexa/pull/1024) comparison I offered as an open merge-order
question was settled before I wrote it, and by absorption rather than by ordering.

I read the thread only after posting. Apologies for the noise, @danieldormann — the substance above
mine stands and mine adds nothing.

DmitriyG228:
Shipped in v0.12.22 via #1125 — cherry-picked with your authorship and sign-off intact, credit in the release notes. Closing as delivered. Danke!

PR #1123, opened 2026-08-11
fix(teams,transcript): speaker-producer names, transcript retract/promote, mainAudio-only transcription [1/4 v0.12.22]

DmitriyG228:
A superseded draft is now retracted instead of left standing, a confirmed turn's dangling tail is promoted instead of dropped, and Teams transcribes only the `mainAudio` mix with double-mirrored tracks deduped.

68 assertions green across `mixed-pipeline`, `teams-capture` and `mixed-capture-core`; 5 commits, 24 files.

Commits authored by @jbschooley.

Stack position 1 of 6 — base `main`. v0.12.22-rc.1 staging candidate.
Next: #1124

Origin: #1024 (reference only).

## Contribution rights

- [x] **Independent:** I created this contribution, or otherwise have the right to submit it
  under Apache-2.0, and it is not owned or controlled by an employer, client, or other entity.
  
- [ ] **Employer/client authorization required:** an employer, client, or other entity owns or
  may control this contribution. I am requesting Vexa's private corporate-authorization process.
  
- [ ] **Unsure:** I need a private rights review before merge.

PR #1124, opened 2026-08-11
fix(teams): speaker producer — stable data-tid names, every tile accounted for, level-held indicators [2/6 v0.12.22-rc.1]

DmitriyG228:
Teams speaker producer: display names resolved from the stable `data-tid`, every tile accounted for with the firing indicator named, edge indicators held as levels, and an unnamed turn claimed by its hint record rather than by whoever fired next. Producer diagnostics and health are surfaced on the bot.

172 assertions green across `mixed-pipeline`, `teams-capture` and `mixed-capture-core`; 14 commits (9 new), 42 files.

Stack position 2 of 6 — base `rc-1-transcript-fixes` (#1123). v0.12.22-rc.1 staging candidate.
Next: #1125

### DIFF
--- a/core/meetings/modules/mixed-pipeline/package.json
+++ b/core/meetings/modules/mixed-pipeline/package.json
@@ -11,7 +11,7 @@
   "files": ["dist"],
   "scripts": {
     "build": "tsc",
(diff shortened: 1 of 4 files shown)


DmitriyG228:
**This PR was marked merged in error. Nothing from it reached `main` through this record.**

While rebuilding the v0.12.22 stack onto current `main`, the six branch heads were force-pushed *before* their bases were retargeted. That briefly left this PR's head (`rc-3-teams-producer`, commit 7 of 54) as an ancestor of its own base branch (`rc-1-transcript-fixes`, commit 14 of 54) — and GitHub marks a PR merged when its commits are already contained in its base. The same thing happened to #1127.

A merged PR cannot be reopened, so this record is closed for good. **No commits were lost.**

**Where this work lives now:** every commit that was here is part of #1123, which now covers commits 1–14 of the release (speaker producer *and* the transcript fixes). It ships in `v0.12.22`.

The correct order for a stacked force-push is bases first, then heads. That has been written back into our release lessons.

PR #1126, opened 2026-08-11
fix(mixed): hallucination gate, gap-reclaim, and the v0.12.22 release freeze [4/4 v0.12.22]

DmitriyG228:
Invented media-artifact text is suppressed and every suppression reported; the segmenter's silence verdict no longer deletes speech that was reclaimed from the gap.

213 assertions green across `mixed-pipeline`, `teams-capture` and `mixed-capture-core`; 20 commits (4 new), 50 files.

Stack position 4 of 6 — base `rc-2-teams-name-resolver` (#1125). v0.12.22-rc.1 staging candidate.
Next: #1127

## Contribution rights

- [x] **Independent:** I created this contribution, or otherwise have the right to submit it
  under Apache-2.0, and it is not owned or controlled by an employer, client, or other entity.
  
- [ ] **Employer/client authorization required:** an employer, client, or other entity owns or
  may control this contribution. I am requesting Vexa's private corporate-authorization process.
  
- [ ] **Unsure:** I need a private rights review before merge.

PR #1125, opened 2026-08-11
feat(capture-signal): bounded signal tapes, and the structural Teams name resolver [2/4 v0.12.22]

DmitriyG228:
Teams speaker names resolve structurally from the DOM leaf rather than from an obfuscated class name, with the timestamp guard applied.

177 assertions green across `mixed-pipeline`, `teams-capture` and `mixed-capture-core`; 16 commits (2 new), 42 files.

Commits authored by @danieldormann.

Stack position 3 of 6 — base `rc-3-teams-producer` (#1124). v0.12.22-rc.1 staging candidate.
Next: #1126

Origin: #1121 (reference only).

## Contribution rights

- [x] **Independent:** I created this contribution, or otherwise have the right to submit it
  under Apache-2.0, and it is not owned or controlled by an employer, client, or other entity.
  
- [ ] **Employer/client authorization required:** an employer, client, or other entity owns or
  may control this contribution. I am requesting Vexa's private corporate-authorization process.
  
- [ ] **Unsure:** I need a private rights review before merge.

PR #1127, opened 2026-08-11
feat(capture-signal): tape the transport and the DOM, store it bounded, ship it at teardown [5/6 v0.12.22-rc.1]

DmitriyG228:
Captured-signal tapes: the transport's own account of who is audible is read and taped, the tape is capped and carries every stream, it ships at teardown, and meeting-api stores and bounds what is kept. Enabled by default with a kill switch on bot context; the 0.10 closed-caption lane reopens as a second attribution source; the build is stamped into the image so a tape can name its own bot.

273 assertions green across `mixed-pipeline`, `teams-capture` and `mixed-capture-core`; 29 commits (9 new), 88 files.

Stack position 5 of 6 — base `rc-4-pipeline-quality` (#1126). v0.12.22-rc.1 staging candidate.
Next: #1128

### DIFF
--- a/core/meetings/modules/mixed-pipeline/src/tape-replay.ts
+++ b/core/meetings/modules/mixed-pipeline/src/tape-replay.ts
@@ -219,6 +219,15 @@ async function main(): Promise<void> {
   // produced the text render above. STT output varies between runs, so the two artefacts have
   // to come from one run or they describe different transcripts. `Speaker` is written out here
   // exactly as the render shows it — a provisional seg_N is not a display name.
+  // Every publish call in order, drafts included — the ONLY way to ask what the retract leg
+  // deleted. `durable` shows what survived; this shows what was ever shown to a viewer, so a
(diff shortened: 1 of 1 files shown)


DmitriyG228:
**This PR was marked merged in error. Nothing from it reached `main` through this record.**

While rebuilding the v0.12.22 stack onto current `main`, the six branch heads were force-pushed *before* their bases were retargeted. That briefly left this PR's head (`rc-5-capture-signal`, commit 22 of 54) as an ancestor of its own base branch (`rc-4-pipeline-quality`, commit 54 of 54) — and GitHub marks a PR merged when its commits are already contained in its base. The same thing happened to #1124.

A merged PR cannot be reopened, so this record is closed for good. **No commits were lost.**

**Where this work lives now:** every commit that was here is part of #1125, which now covers commits 15–24 of the release (capture-signal *and* the structural name resolver). It ships in `v0.12.22`.

The correct order for a stacked force-push is bases first, then heads. That has been written back into our release lessons.

PR #1128, opened 2026-08-11
feat(mixed): transport-drawn turns, word-level attribution, roster elimination, source→name correlator [3/4 v0.12.22]

DmitriyG228:
The transport draws the turn and the track is named once; attribution moves to word granularity; the roster panel names the last speaker by elimination and says so when nothing can name anyone; a bot is not a speaker and a placeholder is not a name; the source→name correlator is extracted platform-agnostic; audio that already produced a final is never re-transcribed. Includes the tape-replay harness on the tape's own clock and its equivalence bound.

389 assertions green across `mixed-pipeline`, `teams-capture` and `mixed-capture-core`; 47 commits (18 new), 99 files. Tip tree is identical to `teams-hot` at `dbac24ac`.

Stack position 6 of 6 — base `rc-5-capture-signal` (#1127). v0.12.22-rc.1 staging candidate.

## Contribution rights

- [x] **Independent:** I created this contribution, or otherwise have the right to submit it
  under Apache-2.0, and it is not owned or controlled by an employer, client, or other entity.
  
- [ ] **Employer/client authorization required:** an employer, client, or other entity owns or
  may control this contribution. I am requesting Vexa's private corporate-authorization process.
  
- [ ] **Unsure:** I need a private rights review before merge.

PR #1136, opened 2026-08-12
feat(gateway): GET /version — the deployment says what it is running

DmitriyG228:
`dashboard.staging.vexa.ai` showed **v0.12.18** while the cluster served **v0.12.22-rc.3**, during a deploy verification. The number was a build-time constant inside the dashboard image, describing the backend that image had been baked against. A frontend cannot honestly report a backend version it compiled in.

So the backend answers for itself.

- **`GET /version`** on the gateway — public and keyless like `/health`, reads `VEXA_VERSION` / `VEXA_REVISION` from the environment **on every request**, returns `"unknown"` when unset. No fallback constant: a stale number is worse than none, because a reader acts on it.
- The chart renders `VEXA_VERSION` through the **same helper** as the gateway container's `image:` tag, so what a deployment reports and what it runs are one value with one editable source. `deploy/helm/tests/test_template.sh` fails the render if they diverge.
- Compose interpolates it from the same `IMAGE_TAG` that selects the image.
- `/version` joins `/health` in the edge-guard exclusions: one server-side proxy asks it on every page load, and a single pod IP would otherwise spend its whole per-IP budget on it.

Additive and off the RC path — it adds a route and an env, changes no existing behaviour. Stacks alongside the v0.12.22-rc.1 set (#1123–#1128); ships as a follow-up deploy **after** 0.12.22, not part of the promotion.

**Validation**
- gateway suite `115 passed, 1 xfailed` (4 new tests: env-read-per-request, `"unknown"` on unset, keyless)
- `gate:helm PASS`, `gate:config-contract` green (5 services · 208 keys)
- image built on bbb and run twice from the **same image**: stamped env → `{"version":"0.12.22-rc.3","revision":"5f314bbd"}`; unstamped → `{"version":"unknown","revision":"unknown"}`

The two frontends that consume this are `Vexa-ai/vexa` branch `dashboard-version-truth` and `Vexa-ai/vexa-platform` branch `version-truth`. Until the gateway carries the route, both chips honestly read *version unknown*.

### DIFF
--- a/core/gateway/services/gateway/src/gateway/app.py
+++ b/core/gateway/services/gateway/src/gateway/app.py
@@ -13,7 +13,9 @@ behavior is the v0.12 carve of the deployed ``services/api-gateway/main.py``:
     unknown_action / invalid_subscribe_payload / invalid_unsubscribe_payload / missing_api_key
     error vocabulary; raw redis payloads forwarded over ``tc:…:mutable`` / ``bm:…:status`` /
     ``va:…:chat`` (main.py:2165-2340),
-  * ``/health`` — liveness ``{status:"ok", service:"gateway"}`` (gate:health discovers it).
+  * ``/health`` — liveness ``{status:"ok", service:"gateway"}`` (gate:health discovers it),
+  * ``/version`` — the running release identity read from the environment at request time, so a
+    UI can report what the deployment IS rather than what its own build was baked against.
 
 The collaborators (admin-api, downstream services, redis) are injected as PORTS (``ports.py``)
 so the same app runs with real adapters in prod (``adapters.py``) and in-process fakes in the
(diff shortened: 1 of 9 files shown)


DmitriyG228:
**Follow-up before this is consumed — `VEXA_VERSION` is currently sourced from the wrong layer.**

As written here, the chart renders `VEXA_VERSION` from the gateway's own image tag. That is component-level: on a release where the gateway does not change, the tag stays at the previous release (prod today pins `v0.12.18-260723.stage2` while the release is 0.12.22). The frontends that consume this route treat the version they display as the **release** — what helm, the ownership lock and the GitHub release all name — so a gateway answer sourced from the image tag would put a true fact about the wrong layer into a public headline.

Consequence for the consumers, which are already built: both the webapp and the dashboard resolve gateway → release-pin → unknown, and both will keep **preferring their release pin over this route** until `VEXA_VERSION` names the release. Nothing breaks; the live answer is simply not preferred yet.

The fix belongs in the ceremony, not in this PR — render `VEXA_VERSION` from the release identity the deploy is naming (as the platform chart does for `webapp.env.RELEASE_VERSION` / `dashboard.platformVersion`) rather than from the image tag, and keep the image tag as a separate `image_tag` field so operator-level precision is not lost. The helm assertion in `deploy/helm/tests/test_template.sh` should then assert the release value instead of the image tag.

Leaving the route and its tests as they are — the shape is right, only the source needs repointing when this is deployed.

PR #1139, opened 2026-08-12
fix(release): bound the gh buffer so the release gates survive a large batch

DmitriyG228:
## What broke

`release-value-gate.mjs` and `release-witness-script.mjs` both call `gh api` through an `execSync` helper with no `maxBuffer`, so Node's **1 MB default** applied.

A release compare carries every file patch, so the payload scales with the size of the batch — not with the handful of fields these scripts read from it. `compare/v0.12.18...v0.12.22` returns **1,581,885 bytes** (77 commits, 242 files).

Both scripts died with `spawnSync /bin/sh ENOBUFS`:

- the **witness receipt could not be generated at all** for that release, and
- the **value gate reported "could not evaluate"** (exit 3) instead of a verdict.

The failure mode is the wrong way round: the gates go blind on exactly the largest releases, which are the ones most worth gating.

## The fix

One bound, both scripts. **No verdict logic changes** — same requests, same parsing, same exit codes; only the buffer moves.

## Verified against v0.12.22

| Script | Before | After |
|---|---|---|
| `release-value-gate.mjs` | `ENOBUFS`, exit 3 "could not evaluate" | real verdict — `23 of 23 PR(s) ACCEPTED`, 54 unaccounted commits correctly reported |
| `release-witness-script.mjs` | `ENOBUFS`, no output | emits its 23 values (`2 user-visible, 21 by-proxy`) |

## Not fixed here

Three sibling scripts carry the identical helper and the same latent defect:

- `scripts/docs-current-gate.mjs`
- `scripts/merge-card-gate.mjs`
- `scripts/release-witness-template.mjs`

Left alone to keep this change to the two scripts currently blocking a release. They want the same one-line fix before their payloads grow.

## Contribution rights

### DIFF
--- a/scripts/release-value-gate.mjs
+++ b/scripts/release-value-gate.mjs
@@ -22,6 +22,11 @@
 
 import { execSync } from "node:child_process";
 
+// A release compare carries every file patch, so the payload scales with the batch, not with
+// what these scripts read from it. v0.12.18...v0.12.22 returned 1.51 MB and blew execSync's 1 MB
+// default, taking the gate out on exactly the largest releases. Bound it well above any batch.
+const MAX_GH_BYTES = 256 * 1024 * 1024;
+
 const REPO = process.env.GITHUB_REPOSITORY;
 const VERSION = process.env.RELEASE_VERSION;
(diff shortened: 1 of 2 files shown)


DmitriyG228:
Rights selection (Independent) and value-signed label executed on the founder'\''s explicit instruction ("go sign approved", session 2026-08-12). The certification is the founder'\''s; this comment records the delegation.

Issue #875, opened 2026-07-21
Discord as a first-party platform lane: native voice capture, all code in-tree like every other platform


DmitriyG228:
## Value this change must deliver

> A user sends a Vexa bot to a Discord voice channel exactly like any other platform — same API call, same lifecycle, same dashboard/API/MCP surface — with speaker-tagged transcripts produced by Vexa's own pipeline. All code lives in the Vexa tree, first-party, like the gmeet/teams/zoom lanes: no external bridge, no third-party dependency for the capture path, no direct-to-Postgres writes.

## Where we are (honest)

- Platform plumbing landed in #452: `Platform.DISCORD`, `construct_meeting_url` → None, concurrency exclusion, dashboard icon.
- #463 proved feasibility: the DAVE E2EE receive path (voice-gateway v8, MLS group join per RFC 9420, per-SSRC key ratchets, rtpsize transport decrypt, per-sender frame decrypt → Opus → PCM) is implementable and works. That existed as an external process writing rows straight into Postgres — bypassing the meeting lifecycle, the collector contract, and the API surface.
- Direction decision (maintainer, 2026-07-21): first-party, **in-tree**. The DAVE receive implementation is written/owned inside Vexa as a platform module. (If any code is derived from MIT sources, the license header and attribution ride with the file — Category A, FINOS-clean — but the module is Vexa's, maintained in-tree like every other lane.)

## Components (waypoints)

1. **Discord capture module** (in-tree, sibling to the other platform lanes): gateway connect + voice join with operator-supplied `DISCORD_TOKEN`, DAVE receive path, per-sender 48 kHz→16 kHz PCM with sender identity as the speaker hint source. Discord delivers per-sender streams — this is the **gmeet-shaped** (per-channel) lane, not the mixed lane.
2. **Lifecycle integration** — bot request → spawn → active → leave/teardown through the normal meeting FSM; `left_alone` for an empty voice channel.
3. **Pipeline wiring** — per-sender streams into the per-speaker pipeline (`@vexa/gmeet-pipeline`-class), segments through the standard collector/store path.
4. **Surface** — transcripts/participants on api.v1 like any platform; Discord lane docs (folds into #537's integrations surface).

## Forks to decide early

- DAVE core provenance: clean-room in-tree implementation vs derived-with-attribution from the proven MIT code — decide by maintenance cost against Discord protocol updates; either way the code lands and lives in the Vexa tree.
[...]

PR #1147, opened 2026-08-13
feat(discord): first-party Discord platform lane (#875)

rennf93:
Implements #1335 (the prepared issue written from this code). Built against the #875 sketch, Option A, claimed there on 2026-07-26.

Discord becomes an in-tree platform lane like gmeet/teams/zoom: a bot is sent to a voice channel through the normal API, and speaker-tagged transcripts flow through the standard collector path. No external bridge, no direct-to-Postgres writes. Discord delivers per-sender audio streams, so this is the gmeet-shaped (per-channel) lane, not the mixed one. The bot is not a browser bot, so it runs under its own runtime profile rather than riding the Playwright image.

## What landed

**Contracts and API.** `discord` added to `invocation.v1`'s Platform enum (back-compatible widening, re-sealed per the procedure in `scripts/gates.mjs`), with a new golden, the TS Platform mirror widened, and `discord` added to `stop_router`'s supported set. A Discord voice channel has no join-by-URL concept, so `meeting_url` is optional for discord (`NO_MEETING_URL_PLATFORMS`): `native_meeting_id` is the channel id and is sufficient, and the sealed `meetingUrl` key survives as an explicit JSON null. The two tests that asserted `discord` is rejected now use `webex` as the negative control, and new tests cover the id-only, URL-supplied, profile-resolution and stop-addressable paths.

**Runtime.** A `discord-bot` profile pointed at `DISCORD_BOT_IMAGE`, with `DISCORD_TOKEN` forwarded through `base_env` and declared secret. One operator-supplied bot token per deployment; Discord has no per-meeting auth.

**Service.** `core/meetings/services/discord-bot`, Python 3.11 (the receive stack uses stdlib `audioop`, removed in 3.13). The DAVE/E2EE receive stack is ported from the proven external bridge: voice gateway v8, the DAVE binary op-flow, rtpsize transport decrypt, MLS group join and per-sender ratchets, per-SSRC Opus decode to 48 kHz. On top of it, an adapter layer speaks the sealed wire contracts: `invocation.v1` in via `VEXA_BOT_CONFIG`, `transcript.v1` out via XADD to `transcription_segments`, `lifecycle.v1` out to the callback URL, `acts.v1` in over Redis, plus `left_alone` on an empty channel.

**Offline proof.** 106 tests, no network and no live Discord, following the mock-bot pattern. The protocol tests run against the real `dave.py` wheel with only MLS internals faked.

## Acceptance status

| # | Observation | Status |
|---|---|---|
| A5 | Offline leg: recorded fixture drives the module without live Discord, in CI | done, 106 tests |
| A4 | Lifecycle: leave-on-empty-channel both directions; graceful acts.v1 stop; duplicate-meeting exclusion for the same channel holds | witnessed live 2026-08-14 (865 s hold, then `left_alone`; meeting 1 `stopped` via acts.v1) plus regression tests |
| A- | Negative control: existing platform lanes' suites stay green | done, see gates below |
| A1 | `POST /bots` with `platform=discord` joins the voice channel | witnessed live 2026-08-14, spawn to active ~4.9s (evidence comment below) |
| A2 | Two humans speaking produce speaker-tagged rows with Discord display names | witnessed live 2026-08-14, 68 rows / 2 speakers through the standard collector path |
| A3 | DAVE proven against a current E2EE voice channel | witnessed live in-tree 2026-08-14 |

The live witness ran on 2026-08-14 (two humans, real E2EE channel): full evidence in the comment below, including two lane bugs it caught that the offline suite could not reach (left_alone firing off the empty member cache, and an incomplete redis teardown), both fixed on this branch.

## Known limitation of the spawn model

A Discord bot holds exactly one voice connection per guild per bot token (py-cord keys `_voice_clients` by guild id, `discord/state.py:291`, and raises `ClientException("Already connected to a voice channel.")` on a second connect in the same guild, `abc.py:2076`). Container-per-meeting spawns an independent process per meeting, so two concurrent meetings in the same guild under the same token cannot detect each other: the second join moves the bot and the first capture ends. Observed in production on 2026-08-14 when two deployments sharing one token stole the guild's voice session from each other. Supported shape today: one active meeting per guild per token; the duplicate-meeting exclusion still refuses a second bot for the same channel. Lifting it needs a persistent per-deployment Discord service that `POST /bots` addresses as a command rather than a spawn, which is a maintainer design decision recorded on the issue, not something this PR attempts.

## Gates

`gate:python` (13 packages), readme, isolation, isolation-py, exports, graph, graph-py, schema, contract-version, config-contract, dataflow, db-schema, db-budget, docs-version, licenses: all green. Re-run after merging current main (2026-08-26): meeting-api 1195 passed / 5 skipped, runtime 164 passed / 1 skipped (the docker-backend lifecycle test ran against a real container; the k8s-backend one needs a live cluster), discord-bot 106 passed, TS bot suite green, the pre-push hook's 14 static gates (readme, dataflow, isolation, isolation-py, exports, graph, graph-py, schema, contract-version, config-contract, licenses, execution-env, test-isolation, contract-conformance) green; zero file overlap between this branch and main's changes since the fork point.

## Provenance and licensing

The receive stack under `dave_voice/` is derived from https://github.com/rennf93/discord-vexa-bridge, which is AGPL-3.0 standalone. As its sole copyright holder I contribute it here under Apache-2.0 to match this tree, per the written grant on #875. Every ported file carries the header and a provenance line.

One correction to the record, because it matters for #875's "no third-party dependency for the capture path" framing. The MLS/RFC-9420 group state machine, epoch ratchet, and AES128-GCM frame decrypt are **not** implemented here or in the bridge. They are delegated to [`dave.py`](https://github.com/DisnakeDev/dave.py), a prebuilt MIT-licensed PyPI wheel published by Disnake Development that binds Discord's own libdave. That is a third-party runtime dependency in the capture path. I described it inaccurately in my earlier comment on #875 as my own pybind11 binding compiled in my image; it is neither. What is mine is the pure-Python glue around it, which is the whole of `dave_voice/`: gateway v8, the opcode framing and 21-31 MLS op-flow, group-join orchestration, transport decrypt, SSRC to user to decryptor routing, and Opus decode.

Keeping the dependency is deliberate: libdave is the audited reference implementation and the only thing that tracks Discord's protocol bumps for free. Reimplementing MLS to avoid it would be a large step down in correctness on a protocol with no room for approximation. If a third-party wheel in the capture path is unacceptable, say so and we can discuss vendoring or a build-from-source binding, but it should be an explicit decision rather than an accident.

## Two upstream defects this surfaced

Both are fixed inside this lane, but they are not Discord-specific and are worth their own issues.

1. **`acts.v1` leave commands violate their own sealed schema.** `meeting-api/lifecycle/stop.py` sends `{"action": "leave", "meeting_id": <id>}`, while `acts.v1`'s `Leave` variant sets `additionalProperties: false` with no `meeting_id` property. Strict validation drops every real stop command. The TS reference adapter survives only because `parseAct` is structurally lenient rather than schema-strict. This lane matches that leniency and documents why; the producer and the sealed schema are untouched here.
2. **The `Transcription` envelope is missing schema-required `session_uid`** and carries an undeclared `native_meeting_id`. The TS bot's `transcript-redis.ts` emits the identical shape, so this is pre-existing drift rather than something this lane introduces. Left as-is deliberately, to match the reference producer.

## Related

The CPU-deployment deadlock in the transcription worker that this work ran into (`MAX_ACTIVE_REQUESTS` defaults to 20 from RTX 4090 benchmarks, which permanently wedges a CPU worker: 503 on every POST while `/health` stays 200) is #1158, with the fix open as PR #1156.

## Contribution rights


rennf93:
# Live-witness evidence bundle: Vexa PR #1147 (Discord platform lane, #875 acceptance)

This witness was run by the lane author (rennf93) on his own local deployment, with a consenting second participant (name withheld here for privacy; display-name attribution verifiable by a maintainer on request), against a real, live, E2EE Discord voice channel. Nobody outside the two participants was recorded. The full transcript rows, including personal conversation not reproduced here, live in the local witness Postgres DB (`vexa-witness-postgres-1`, database `vexa`) and are available on request. No conversational content is reproduced in this document. The tables below show turn structure only (timestamps, speaker attribution, word counts); the words themselves stay in the local witness DB.

All times below are 2026-08-14, UTC (as stored). Wall-clock spans: meeting 1 ~16:40:28-16:42:10, meeting 2 ~16:43:19-16:45:31, meeting 3 ~16:55:25-17:09:55.

## 1. Acceptance-observation map

| #875 observation | What it claims | Evidence | Timestamp(s) |
|---|---|---|---|
| **A1** - channel-ref join, no synthesizable URL | Bot resolves the `discord-bot` runtime profile (not the Playwright browser-bot default) and joins a voice channel by bare channel snowflake, no `meeting_url` required | Meeting 1 `status_transition`: `joining` -> `active` in 0.86s from a cold container spawn; meetings 2 and 3 show the same ~4-5s spawn-to-active pattern. Root-caused and fixed in commit `99ba04c4` (`build_workload_spec` was hardcoding `profile="meeting-bot"` for every platform before this fix - a discord spawn would have launched the browser bot and navigated a blank URL) | Meeting 1: created_at 16:40:28.660 -> active 16:40:33.583 (4.92s). Meeting 3: created_at 16:55:25.246 -> active 16:55:30.080 (4.83s) |
| **A2** - transcript delivery through the standard pipeline | Speaker audio reaches Postgres via the sealed path: bot -> `XADD transcription_segments` -> collector -> `transcriptions` table, not a direct DB write | Meeting 2: 18 rows in `transcriptions` for `meeting_id=2`, both `Renzo` and `Participant 2` present. Redis `transcription_segments` stream and per-meeting collector cursor keys `tc:meeting:1/2/3` still live in `vexa-witness-redis-1`, confirming the stream path was exercised, not a shortcut | Meeting 2 transcript span 3.53s-125.80s (session-relative) |
| **A3** - per-sender speaker attribution | Discord's per-sender audio streams resolve to correctly-named, per-speaker rows (not a single mixed-audio blob) | Meeting 2 and meeting 3 both show exactly 2 distinct `speaker` values (`Renzo`, `Participant 2`), interleaved and alternating in the correct turn order across dozens of rows each | Meeting 2: 2 speakers / 18 rows. Meeting 3: 2 speakers / 68 rows |
| **A4** - `left_alone` correctness, both directions | The alone-timer must NOT fire while humans are present, and MUST fire ~120s after they genuinely leave | **Negative direction failed live, then fixed**: meeting 2 hit false-positive `left_alone` at ~128s with both humans still mid-conversation (root cause: `channel.members` reads the empty guild-member cache without the privileged members intent). **Positive direction confirmed after the fix**: meeting 3 held the channel 13 minutes (well past the old 128s failure point) then completed with `left_alone` ~120s after both participants actually signed off. Fix in commit `443fa47b` | Meeting 2 completed 16:45:31.355 (completion_reason `left_alone`, ~127.4s active). Meeting 3 completed 17:09:55.412 (completion_reason `left_alone`, last transcript row at t+725.70s, ~865s active total) |
| **negative control** | The platform allow-list still rejects genuinely unsupported platforms after `discord` was added as a spawnable platform | `test_post_bots_invalid_platform_should_be_422` was repointed from `discord` to `webex` as the stand-in unsupported platform in commit `4bfe41c4`, paired with new `test_post_bots_discord_meeting_url_accepted` / `test_post_bots_discord_without_meeting_url_is_422`. Offline/code evidence, not a live DB row | `core/meetings/services/meeting-api/tests/test_api_agility.py`, commit `4bfe41c4` |
| **graceful stop (acts.v1)** | An operator-issued `DELETE /bots/discord/{id}` cleanly stops an active bot via the acts.v1 command bus, distinct from `left_alone` | Meeting 1's row: `"stop_requested": true`, `"completion_reason": "stopped"` (not `left_alone`, not `failed`). Bot container exits 0 with no error other than the (separately-tracked) redis-teardown race. This exercised the acts.v1 leave-command fix in `4bfe41c4` (meeting-api's producer sends `{"action":"leave","meeting_id":...}`, which the sealed schema's strict `additionalProperties:false` had been silently rejecting before that fix) | Meeting 1 completed 16:42:10.340, `completion_reason: stopped` |

## 2. Per-meeting detail

### Meeting 1 - first spawn, transcription URL misconfigured, graceful stop

- Row: `id=1`, `platform=discord`, `status=completed`, `completion_reason="stopped"`, `stop_requested=true`, `segments_captured=0`, `transcription_outcome="no_output"`.
- `status_transition`: `joining` @16:40:32.721 -> `active` @16:40:33.583 -> `completed` @16:42:10.340 (`source: bot_callback` throughout, `completion_reason: stopped`).
- Row counts: 0 transcription rows (the transcription-worker URL was misconfigured stack-side at this point, so the segmenter had audio but no working sink).
- Log excerpt (`docker logs vexa-mtg-1-2fe7c9ec`, scrubbed):
  ```
  [discord-bot] [discord-bot] transcribe HTTP 405
  ... (28 consecutive lines, same message, spanning the whole ~97s active window)
  [discord-bot] gateway connect task ended with an exception: RuntimeError('Session is closed')
  redis.exceptions.ConnectionError: Connection closed by server.
  ```
  The 405s are the segmenter repeatedly POSTing to a transcription endpoint that wasn't wired correctly yet (fixed before meeting 2 - meeting-api's own log buffer restarts cleanly at 16:43:19, right at meeting 2's join, because the fix required a `meeting-api` container reload to pick up the corrected `TRANSCRIPTION_SERVICE_URL`; that's also why meeting-api's own logs don't cover meeting 1's DELETE call directly - the DB's `stop_requested`/`completion_reason` fields are the primary record for that path).
- Timing: spawn-to-active 4.92s (A1 evidence); active window 96.76s; operator-issued stop via `DELETE /bots/discord/1517719641449955361`.

### Meeting 2 - fixed URL, speaker-tagged rows land, false-positive `left_alone` bug

- Row: `id=2`, `status=completed`, `completion_reason="left_alone"`, `segments_captured=17` (data field; 18 rows are actually present in `transcriptions` - a one-row discrepancy between the in-flight counter snapshotted into `meetings.data` and the final DB count, not investigated further here).
- `status_transition`: `joining` @16:43:23.125 -> `active` @16:43:23.990 -> `completed` @16:45:31.355 (`completion_reason: left_alone`).
- Row counts: 18 transcription rows, 2 speakers (`Renzo` 9 rows, `Participant 2` 9 rows), transcript span 3.53s-125.80s session-relative.
- meeting-api lifecycle log (scrubbed, real excerpt):
  ```
  {"event":"bot_join_requested", ...,"fields":{"platform":"discord","status":"requested","session_count":1}}
  {"event":"meeting_lifecycle_advanced", "fields":{"meeting_status":"joining"}}   16:43:23.219820Z
  {"event":"meeting_lifecycle_advanced", "fields":{"meeting_status":"active"}}    16:43:24.001749Z
  {"event":"meeting_copilot_reap_signalled", "fields":{"meeting_status":"completed"}}  16:45:31.376487Z
  {"event":"meeting_lifecycle_advanced", "fields":{"meeting_status":"completed"}}      16:45:31.376512Z
  ```
- Sample rows (both speakers, coherent lines; full 18-row transcript remains in the witness DB):

  | t (s) | speaker | words |
  |---|---|---|
  | 3.53 | Participant 2 | 4 |
  | 45.43 | Renzo | 14 |
  | 52.75 | Renzo | 33 |
  | 58.85 | Participant 2 | 1 |
  | 65.83 | Participant 2 | 7 |
  | 80.04 | Renzo | 60 |
  | 119.58 | Participant 2 | 21 |

- Bug: the bot exited at **~127.4s active** (16:45:31.355 minus 16:43:23.990) while both humans were still present and mid-conversation - a real false-positive `left_alone`. Root cause and fix documented in section 3.
- Container log tail is identical in shape to meeting 1's (same redis `ConnectionError` at teardown - see section 3).

### Meeting 3 - fix applied, image rebuilt, full 13-minute hold, legitimate `left_alone`

- Image: `vexa/discord-bot:witness` was rebuilt between meeting 2 and meeting 3. `vexa-mtg-1-2fe7c9ec` and `vexa-mtg-2-d16f369b` both ran image `sha256:24fd33052460...` (pre-fix); `vexa-mtg-3-d689623e` ran the rebuilt `sha256:810cef620eb9...` (post-fix, commit `443fa47b`).
- Row: `id=3`, `status=completed`, `completion_reason="left_alone"`, `segments_captured=68` (matches the DB row count exactly).
- `status_transition`: `joining` @16:55:29.255 -> `active` @16:55:30.080 -> `completed` @17:09:55.412.
- Row counts: 68 transcription rows, 2 speakers (`Renzo` and `Participant 2`), transcript span 1.24s-725.70s session-relative (~12 min of continuous conversation), active window 865.33s total (~14.4 min) - i.e. the bot held the channel well past the old 128s failure point, then legitimately timed out ~120s after the last real activity.
- meeting-api lifecycle log (scrubbed, real excerpt):
  ```
  {"event":"bot_join_requested", "fields":{"platform":"discord","status":"requested"}}
  {"event":"meeting_lifecycle_advanced", "fields":{"meeting_status":"joining"}}   16:55:29.351092Z
  {"event":"meeting_lifecycle_advanced", "fields":{"meeting_status":"active"}}    16:55:30.090678Z
  {"event":"meeting_copilot_reap_signalled", "fields":{"meeting_status":"completed"}}  17:09:55.452380Z
  {"event":"meeting_lifecycle_advanced", "fields":{"meeting_status":"completed"}}      17:09:55.452450Z
  ```
- Turn-structure sample (both speakers, spanning start to the natural sign-off; the full 68-row transcript remains in the witness DB, no content reproduced here):

  | t (s) | speaker | words |
  |---|---|---|
  | 1.85 | Participant 2 | 4 |
  | 52.14 | Participant 2 | 2 |
  | 374.47 | Participant 2 | 5 |
  | 406.55 | Participant 2 | 9 |
  | 671.63 | Renzo | 7 |
  | 687.21 | Participant 2 | 11 |
  | 698.22 | Renzo | 9 |
  | 723.32 | Renzo | 5 |

  The last row above (t=723.32s) is essentially the moment both participants sign off; `left_alone` then fires at the standard ~120s `everyoneLeftTimeout` after that, landing the `completed` callback at 17:09:55, consistent with the alone-poll's periodic granularity.
- Container log tail: identical redis `ConnectionError` shape to meetings 1 and 2, **despite running the post-fix image** - see Finding 2 below.

## 3. Findings

### Finding 1 - `left_alone` false-positive from an empty guild-member cache (fixed, `443fa47b`)

`_non_bot_member_count` originally read `channel.members`, which py-cord only populates from the privileged `members` intent - an intent this bot deliberately does not request. Every occupant already in the channel when the bot joins is invisible to `channel.members` until their next voice-state change, so the count read 0 from the moment of joining and the alone-timer fired on schedule at `everyoneLeftTimeout`, killing meeting 2's capture at ~127s with two humans still talking. This was unreachable by the offline test suite because it mocks the member count directly.

Fix: count occupants from `channel.voice_states` instead (py-cord's documented intent-independent replacement), still consulting a cached `Member` where available to exclude other bots, and treating an unclassifiable occupant as human (overcounting keeps the bot in the call; undercounting is the bug being fixed). New regression tests: `test_count_uses_voice_states_not_the_empty_member_cache`, `test_count_excludes_only_ourselves_when_cache_is_empty`, `test_count_excludes_a_cached_bot_member`, `test_uncached_occupant_counts_as_human` (`core/meetings/services/discord-bot/tests/test_bot.py`).

Live-witness proof of the fix: meeting 3 held the same channel for 865s (past the old ~128s failure point) and only completed `left_alone` after both humans genuinely left.

### Finding 2 - redis teardown race (partially fixed, still reproducing after `443fa47b`)

All three witnessed runs - meetings 1, 2, **and 3 (post-fix image)** - exit with the identical unhandled exception:

```
[discord-bot] gateway connect task ended with an exception: RuntimeError('Session is closed')
Task exception was never retrieved
future: <Task finished name='Task-20' coro=<RedisPubSubClient.subscribe.<locals>._listen() ...
redis.exceptions.ConnectionError: Connection closed by server.
```

Commit `443fa47b`'s teardown reorder (extracted `_teardown()`, cancels `background_tasks` - the flush loop and the alone-poll loop - before closing redis) fixed the race for those two loops, and added a passing regression test (`test_teardown_order_is_stop_capture_then_drain_then_close`). But `background_tasks` does not include the acts.v1 Redis pub/sub listener: `RedisPubSubClient.subscribe()` (`adapters/acts_redis.py`) starts its own untracked `_listen()` task and exposes a `close()` method that cancels it - and `_teardown()` never calls it. That listener task is still blocked on `self._pubsub.listen()` when `redis_client.aclose()` runs, so it dies with the same `ConnectionError` every time, exit code 0 notwithstanding. This is corroborated directly: it appears in meeting 3's log even though meeting 3 ran the rebuilt image containing the `443fa47b` fix. Not a live-lane-blocking bug (exit code stays 0, no data loss - it fires after the final flush and lifecycle event in every witnessed run), but it's a real, currently-open gap: the `RedisPubSubClient` returned by `acts.subscribe()` needs to be threaded into `_teardown()` and closed before `redis_client.aclose()`, the same way `background_tasks` now are.

### Finding 3 - same-token voice-hijack between the external bridge and the witness bots (production, same evening)

Reported by the operator from the same evening's NAS deployment (not independently re-pulled into this bundle - NAS log access was out of scope for this collection pass; included here as the operator's own contemporaneous observation, consistent in timing with the 16:40-17:10 UTC witness window above). The operator's existing external bridge (`discord-vexa-bridge`, NAS deployment, `renzof93/discord-vexa-bridge:latest`) and this witness stack's bots share the same Discord bot token and therefore the same bot identity. Discord only allows one voice session per bot identity per guild; the bridge and the witness bots silently stole the guild's single voice session from each other over the course of the evening. When the operator typed `/leave` against the bridge, it closed its own half-captured meeting with 56 queued segments still pending - a real symptom of two independent services racing for one voice slot under one identity. This is direct, live evidence for the per-guild voice-exclusivity design question raised on #875 (container-per-meeting spawn model vs. a persistent bridge-style service): whatever ships needs either a single voice-session owner per guild or an explicit hand-off protocol between deployments sharing a token, because Discord's session-of-one enforces it whether or not the software expects it.

## 4. Environment

- Compose project `vexa-witness` (control plane: `deploy/compose/docker-compose.yml` + a one-file override `docker-compose.witness.yml` adding `DISCORD_BOT_IMAGE`/`DISCORD_TOKEN` to `runtime` and pointing `meeting-api` at the CPU transcription worker over `host.docker.internal`), plus a separate compose project `vexa-witness-tx` for `deploy/transcription/docker-compose.cpu.yml`. Both isolated from the NAS's `vexa` project and from any local `vexa-v012` project (different `-p` names).
- Services: `postgres:17-alpine` (127.0.0.1:5458), `valkey/valkey:8-alpine` (internal), `minio` (127.0.0.1:9000/9001), `vexaai/v012-admin-api:witness`, `vexaai/v012-runtime:witness`, `vexaai/v012-meeting-api:witness` (later `:dev` after the meeting-api reload noted in section 2), all locally built from the `vexa-discord-lane` checkout; `vexaai/v012-transcription-cpu:dev` behind an nginx LB, `MODEL_SIZE=small`, CPU-only.
- Bot image: `vexa/discord-bot:witness`, locally built, two distinct digests witnessed (`sha256:24fd33052460...` for meetings 1-2, `sha256:810cef620eb9...` for meeting 3 after the `443fa47b` rebuild).
- `MODEL_SIZE=small` caveat: several very short interjection rows (1-4 words, visible as low word counts repeated at unrelated timestamps in the tables above) match Whisper's small-model hallucination pattern on short, low-information utterances - a known artifact of the model tier used for this witness run, not a defect in the Discord capture lane itself. Production would typically run a larger model.
- What differs from production: `meeting-api` trusts `x-user-id` directly with no `x-api-key` gate (that check lives in the `gateway` service, intentionally not started for this witness - the runbook documents this as a deliberate shortcut, not a security gap in the lane); no gateway/agent-api/terminal/mcp/dashboard were started; admin token is the compose dev-default `changeme`; single test user (`witness@vexa.ai`, numeric id `1`). None of these affect the discord-bot code path itself, which is what's under witness here.
- Source files for this environment: `witness/runbook.md`, `witness/docker-compose.witness.yml`, `witness/stack-state.txt`, `witness/INCIDENT.md` (a note from an earlier session in this same scratchpad describing a prior, aborted attempt - superseded by the successful runs documented above).

Issue #1148, opened 2026-08-14
mixed pipeline drops text-only custom STT responses as zero-length segments


rene-schwabe:
### Value this issue delivers
Custom OpenAI-compatible STT endpoints that return only `{ "text": "..." }` produce visible, durable transcript segments instead of a blank transcript.

### Why this matters
The custom-STT contract explicitly permits a text-only JSON response, but the mixed pipeline currently discards that valid response shape. Operators can therefore see successful STT requests while the corresponding transcript remains blank.

### Where we are (honest)
- Era: the production observation was made on v0.12.22. Current `main` at `e0b356d6de3f8322db45d3cb9d66282ae108bebf` still contains the same mapping behavior.
- `docs/docs/how-to/custom-stt.mdx` states that `{ "text": "transcribed speech" }` is the minimum accepted response and that segments/timestamps are optional.
- `core/meetings/modules/whisper/src/transcription-client.ts` preserves non-empty `text` when the provider returns no segments.
- `core/meetings/modules/mixed-pipeline/src/chunked-transcriber.ts` converts that response to a fallback segment with `start: 0, end: 0`, then adds `spanStart` to both values. Because the resulting `rawEndMs` is an absolute, truthy timestamp, the fallback to `publishEnd` is not taken. The zero-length segment is then rejected by `s.endMs <= s.startMs`.
- The gap is in the mixed-pipeline brick: it has no fixture for a valid STT result whose text is non-empty and whose segments array is empty.

### Deployments to validate (D12b — a behavior proven on one setup is proven on that one)
- Required offline leg: `@vexa/mixed-pipeline` with its real `ChunkedTranscriber`, injected boundary source, voiced PCM, and a stub STT response containing text but no segments.
- Required live leg: a full Compose v0.12.22 deployment using an OpenAI-compatible custom STT endpoint whose successful response contains non-empty text and no timestamped segments. Infrastructure identifiers, endpoint details, and credentials stay private; the attestation should publish only the Vexa version, response shape, and before/after transcript outcome.
- Lite Compose, k8s/Helm, and hosted are out of scope and remain unclaimed. Lite does not need a separate run because the defect is inside the shared bot image's in-process mapping, not deployment wiring; the offline fixture exercises that exact branch.

### Docs surface (D6c — the change isn't delivered until the docs tell it)
[...]

PR #1149, opened 2026-08-14
fix(mixed): preserve text-only custom STT results

rene-schwabe:
**Delivers issue:** #1148

## Contribution rights

Every commit carries the contributor's matching DCO `Signed-off-by` line.

- **C1 — Preserve text-only STT output through the mixed transcript lane**
  - Expected: a valid `{ text, segments: [] }` result publishes one confirmed transcript segment
    with a positive span matching the committed speech window.
  - Negative control: ran `tsx src/text-only-stt.smoke.test.ts` with the fixture present and the
    implementation at base `e0b356d6de3f8322db45d3cb9d66282ae108bebf`; saw `published=[]`,
    `text-only STT response was not published with a positive speech span`, exit 1.
  - Head: ran the same fixture with `db98a571aad5f095411e60c2a713c8a201ea4d35`; saw one segment
    containing the expected text at `startMs: 10000`, `endMs: 12000`, exit 0.
  - Package: built `@vexa/mixed-pipeline` and its workspace dependencies, then ran the complete
    mixed-pipeline test script; all fixtures passed.
  - Repository: ran `node scripts/gates.mjs all`; all gates passed. Compose reported
    `docker not available → skip (green-or-skip)`; stress and chaos remained their documented
    opt-in skips.

## Acceptance floor

| Row | Evidence |
|---|---|
| A1 | Base `e0b356d6`: fixture RED with `published=[]`; head `db98a571`: fixture GREEN with the expected `10000..12000 ms` confirmed segment. |
| A2 | Sanitized full-Compose operator observation on v0.12.22: the text-only response shape produced no transcript before the mapper change and visible, durable output with the patched bot. Infrastructure identifiers, endpoint details, and credentials are intentionally omitted. A non-author witness is still requested below. |
| A3 | Full mixed-pipeline test script passed; `node scripts/gates.mjs all` completed with `gates green` at the committed tree. |

## Docs diff (D6c)

- `core/meetings/modules/mixed-pipeline/README.md` names the new offline fixture and the contract it
  protects.
- `docs/changelog.d/1148-text-only-stt.md` records the user-visible behavior and links the existing
  custom-STT how-to.
- `docs/docs/how-to/custom-stt.mdx` already states the correct minimum response contract, so this PR
  does not weaken or duplicate it; the implementation now matches that page.

- No dependencies, contracts, deployment settings, endpoints, or credentials changed.
- `gate:licenses`: 460 resolved dependencies OSS-clean; the existing two Cat-B exceptions remain
  unchanged.
- `gate:image-licenses`: six pinned images and one bundled component audited; no image diff.
- TypeScript build, isolation, graph, config-contract, schema, contract-conformance, access,
  tracing, and the complete repository gate suite passed locally.
- GitHub secret scanning and repository SAST checks remain authoritative in CI.

## Validation request

Please have one non-author maintainer or custom-STT operator witness the A1 red→green fixture and,
where a suitable endpoint is available, confirm that a successful text-only response becomes
visible/durable transcript output on full Compose. The public attestation should name only the Vexa
version, Compose setup class, response shape, and before/after outcome; infrastructure identifiers
and secrets must stay private. The witness should also confirm that the existing custom-STT how-to
accurately describes the resulting behavior. Lite, k8s/Helm, and hosted remain unclaimed.

## Authorship


DmitriyG228 (approved):
Reviewed against the mixed-pipeline mapping and the OpenAI-compatible minimum response. The bug is real: a text-only result (no provider timestamps) computed endMs == startMs and the segment died silently — custom-STT users lost the transcript. The fix's interpretation is the honest one: text with no span answers for the submitted speech window. Verified the timestamped path is byte-equivalent to the old arithmetic. One note, not a blocker: a degenerate provider segment with start == end > 0 now stretches to the window's end rather than being dropped — acceptable, since discarding text is the worse failure; if a provider ever emits those legitimately we can clamp then. Test is a real smoke against the actual transcriber and is wired into the suite. Thank you — and apologies for the fifteen-day wait; that's on us, not the PR.

DmitriyG228 (approved):
Re-approval on the current head: the only delta since my validated-train approval is the mixed-pipeline test-chain union (main's #1300 landed first) — content verified identical to the station-validated consist resolution.

PR #1156, opened 2026-08-14
fix(transcription): stop a CPU worker admitting a GPU-sized concurrency default

rennf93:
Implements #1158.

A CPU-only deployment of the transcription worker can permanently deadlock: it answers `GET /health` 200 while returning 503 to every `POST /v1/audio/transcriptions`, forever, until the container is restarted. Observed live on a 12-core CPU deployment on 2026-08-13: a 51-minute meeting was force-completed as `stt_degraded: timeout x393, unavailable x250` with zero transcript rows while the worker reported healthy throughout.

Mechanism: `MAX_ACTIVE_REQUESTS` defaults to 20, a figure the code comments source from RTX 4090 benchmarks. On a CPU device each of the 20 admitted jobs gets 1/20th of the cores, so none finish in bounded time, the semaphore never releases, and with `FAIL_FAST_WHEN_BUSY` (default true) every later request 503s. The worker never self-recovers and never reports unhealthy.

## What this changes

- The default becomes device-aware: 20 only for an explicit `cuda` device, 2 otherwise. Inverted deliberately so faster-whisper's `auto` (which can resolve to CPU at runtime) does not inherit the GPU default. An operator-set `MAX_ACTIVE_REQUESTS`/`MAX_CONCURRENT_TRANSCRIPTIONS` still wins unconditionally.
- `DEVICE` is normalized once at read time (strip + lowercase), so every comparison in the file tolerates casing and stray whitespace.
- A whole-request hang backstop, `TRANSCRIPTION_TIMEOUT_S` (default 120s): a call that never returns cannot hold its semaphore permit forever. The bound covers all temperature-fallback attempts together, not each one. 120s was chosen against a measurement, not a guess: on the reference CPU box (12 cores, int8, large-v3-turbo, 2 concurrent, 6 threads) a 30-second chunk, the gmeet lane's hard cap and the largest input this service is ever handed, takes 28.5 to 29.0s of wall clock single-request, so the backstop sits at ~4x legitimate worst-case work and is expected to fire rarely or never.
- The executor is sized one thread larger than the semaphore: `asyncio.wait_for` can only reclaim the permit, not the thread, so without headroom the next admitted request just queues behind the zombie and the reclaimed permit buys nothing.
- A startup warning when the medium or large model family is paired with a CPU device. `deploy/transcription/docker-compose.cpu.yml` already documents medium as shedding load on CPU, and large-v3-turbo measures ~0.95x realtime there, so neither sustains a live meeting at any concurrency. Prefix match, so `distil-*` models are not flagged alongside the family they were distilled from.
- `http_504` added to the transcription LB's `proxy_next_upstream`: the app can now return 504, and a wedged worker is the worst case to keep sending traffic to, so it fails over like the 503s already did. Config validated with `nginx -t`.

## What it deliberately does not fix

A native call that truly never returns still burns one executor thread forever; the pool never recycles it, so N+1 cumulative hangs over a process lifetime still exhaust it. Closing that needs subprocess isolation, out of scope here and noted in the code.

## Verification

Service suite: 26 passed (5 at baseline). Device-aware default, the inversion, the executor headroom and the timeout are each mutation-tested: reverting any one of them fails exactly the tests written for it. Full repo gates green (`gate:python` 12 packages, schema, contract-version, config-contract, dataflow, readme, isolation, exports, graph, db-schema, db-budget, docs-version, licenses), `pnpm typecheck` 9/9, `pnpm test` 34/34. mypy and ruff findings byte-identical to baseline.

## Contribution rights


DmitriyG228 (approved):
Validated at train scope for 0.12.25 (station record: DmitriyG228/vexa-stations channels/vexa-internal/stations/t25-01225 — 7-lens agent battery + adversarial verify, module suites, bytes-verified station, live founder meeting through the assembled candidate). Three edge-warns recorded in the station record (MAX_ACTIVE_REQUESTS=0 boots into silent-503; double timeout-wave wedge; http_504 failover window mismatch) — riding as release notes; the fix closes a worse live incident class. Thank you.

PR #1157, opened 2026-08-14
fix(gateway): move the guard pin to fastapi-guard 7.8.2 / guard-core 3.15.0 and validate the guard env vars at the boundary

rennf93:
Implements #1159.

Four operator env vars flow raw into fastapi-guard's `SecurityConfig` in the gateway: `GUARD_BLOCK_CLOUD_PROVIDERS`, `GUARD_IP_WHITELIST`, `GUARD_IP_BLACKLIST`, `GUARD_TRUSTED_PROXIES`. None were validated or documented on Vexa's side, and what happens to a bad entry depends entirely on which guard-core version the resolver picked, because fastapi-guard declares guard-core unconstrained.

On the guard-core this repo's own `uv.lock` pins (3.4.0), `block_cloud_providers` is a silent, case-sensitive filter: the natural spelling `aws` is dropped without an error or a log line, so an operator who set it got cloud blocking silently OFF. On guard-core >= 3.12.0 (what a fresh, unpinned resolve already installs) the same entry instead raises `ValueError` deep inside `SecurityConfig` construction, a library stack trace that never names the env var. The IP-list fields already raise on 3.4.0.

Full disclosure: I author fastapi-guard and guard-core. The library raising on bad input is correct and stays; this PR fixes Vexa's side of the boundary, which was handing unvalidated operator input to a closed-set field.

**Revised after maintainer review (2026-08-29/30).** The first revision gated the enforcement behind an opt-in, as asked. That was treating the symptom: the reason an unrecognized entry was inert is that this branch pinned guard-core 3.4.0, where `block_cloud_providers` is a silent, case-sensitive filter. Documenting a fail-open on a security control ships the vulnerability with a manual attached, so this PR now moves the pin instead.

## What this changes

- **The gateway moves to fastapi-guard 7.8.2 / guard-core 3.15.0** (was 7.2.2 / 3.4.0). On this pin an unrecognized or wrong-case `block_cloud_providers` entry raises during `SecurityConfig` construction, in the library, whatever Vexa does.
- All four vars are validated at Vexa's own boundary, before `SecurityConfig` construction, with errors naming the env var, the offending entry and the accepted values, riding the existing `ConfigError` boot-refusal path. For `GUARD_BLOCK_CLOUD_PROVIDERS` this is not new enforcement: the boot fails either way on this pin, and the only difference is whether the operator gets a legible error or a pydantic stack trace from library internals. That is the same treatment the three IP-list vars already had.
- `GUARD_BLOCK_CLOUD_PROVIDERS_STRICT` (new, default `false`) adds case-insensitivity only. It is the one genuinely behavior-changing piece, since normalizing `aws` to `AWS` turns a failing boot into live cloud blocking, so it stays opt-in.
- The accepted provider set is guard-core's six canonical names (`AWS`, `Azure`, `DigitalOcean`, `GCP`, `Linode`, `Vultr`). The previous three-name list would have refused a `DigitalOcean` that guard-core accepts.
- The `:!region` carve-out is functional on this pin (`CloudManager._refresh_providers` strips the selector before fetching ranges, which was the old bug), so the region half is validated for lowercase (guard-core compares it case-sensitively) and the "do not use it yet" warning is gone from `.env.example`.
- IP-list entries are pre-parsed with stdlib `ipaddress`, so a bad entry names the env var instead of surfacing as a pydantic `ValidationError`.
- `.env.example`, the helm values and `docs/configuration.mdx` document accepted values, casing, the opt-in and the carve-out.

## Verification

Gateway suite 337 passed, 1 xfailed. Repo gates green.

Verified in real containers, not only unit tests: the gateway image was built from the new lockfile (fastapi-guard 7.8.2 / guard-core 3.15.0 confirmed inside the image), then booted across a 10 case matrix. Valid configs come up healthy, including `DigitalOcean` and `AWS:!us-east-1`. A wrong-case or unrecognized name exits 1 with `ConfigError: GUARD_BLOCK_CLOUD_PROVIDERS entry 'aws' is not a recognized cloud provider. Accepted values: AWS, Azure, DigitalOcean, GCP, Linode, Vultr.` and `GUARD_BLOCK_CLOUD_PROVIDERS_STRICT=true` boots that same value normalized. A bad `GUARD_IP_WHITELIST` entry fails the same legible way.

One library behavior change surfaced in the bump: guard-core now sets `Retry-After` on every 429, so the test distinguishing the per-token limiter from the guard limiter by that header's absence now discriminates on its value instead.

## Contribution rights


DmitriyG228:
Thank you — the boundary validation is the right instinct, and three of your four land. Ejecting this one from the current train on integrated-state findings: cloud blocking flips from inert to live on upgrade, so AWS/GCP-hosted API clients 403 the day this deploys (`edge_guard.py:119`); a config that booted yesterday can crash-loop today (`:120`); and the error text recommends a carve-out that silently disables blocking on the shipped guard-core (`:123`). Gate the new enforcement behind an explicit opt-in and it rides the next train.

rennf93:
Gated, in fa6d95af.

- `GUARD_BLOCK_CLOUD_PROVIDERS_STRICT` (new, default `false`). Off, `block_cloud_providers` is built exactly as before this PR (`set(_guard_csv(...))`, passed to guard-core as written), so nothing flips live and nothing that booted yesterday refuses to boot. The only addition on that path is one boot WARNING per entry the pinned 3.4.0 will silently drop, naming the entry and the opt-in.
- On, the normalize-and-reject behavior from the original diff applies, unchanged.
- The `:!region` recommendation is out of the error text. The suffix is still parsed (so a config does not have to change once the pin moves past 3.13.0), and `.env.example` keeps its "do not use yet" warning.
- The three IP-list validations are untouched.

Gateway suite 337 passed / 1 xfailed, repo gates green. PR body updated to match.

rennf93:
Correction on my last reply: gating the enforcement was treating the symptom.

The reason an unrecognized entry was inert is that this branch pinned guard-core 3.4.0, where `block_cloud_providers` is a silent, case-sensitive filter. That is a fail-open on a security control: an operator sets cloud blocking, gets nothing, and there is no error and no log. Writing a boot warning to explain that behavior, which is what my previous commit did, ships the vulnerability with a manual attached. The pin was the bug.

So this PR now moves the gateway to fastapi-guard 7.8.2 / guard-core 3.15.0, and the three findings resolve differently:

- **Cloud blocking flipping from inert to live on upgrade.** On 3.15.0 an unrecognized or wrong-case name raises during `SecurityConfig` construction, in the library, whatever Vexa does. So Vexa validating it at the boundary is not new enforcement: the boot fails either way, and the only difference is whether the operator gets an error naming the var and the entry, or a bare pydantic stack trace. That is the same treatment you accepted for the three IP-list vars.
- **A config that booted yesterday crash-looping today.** Still real, and now caused by the version bump rather than by the validator, which is the honest place for it. The one genuinely behavior-changing piece, case-normalizing `aws` to `AWS` so a previously-failing entry starts blocking, stays behind `GUARD_BLOCK_CLOUD_PROVIDERS_STRICT` (default false).
- **The error text recommending `:!region`.** Dropped from the message. Separately, the carve-out is functional on 3.15.0 (`CloudManager._refresh_providers` strips the selector before fetching ranges, which was the bug), so the "do not use it yet" warning in `.env.example` is gone too.

Also corrected: the accepted provider set is six names on this pin (`AWS`, `Azure`, `DigitalOcean`, `GCP`, `Linode`, `Vultr`), not three. The old three-name list would have refused a `DigitalOcean` that guard-core accepts.

Happy to split the dependency bump into its own PR if you would rather review it separately.

Verified in real containers, not just unit tests: the gateway image built from the new lockfile (fastapi-guard 7.8.2 / guard-core 3.15.0 confirmed inside the image), then a 10 case boot matrix. Valid configs boot healthy, including `DigitalOcean` and the `AWS:!us-east-1` carve-out, both of which the old three name list would have refused. A wrong case or unrecognized name exits 1 with:

```
gateway.config_preflight.ConfigError: GUARD_BLOCK_CLOUD_PROVIDERS entry 'aws' is not a recognized
cloud provider. Accepted values: AWS, Azure, DigitalOcean, GCP, Linode, Vultr. Fix or remove the
entry and restart.
```

`GUARD_BLOCK_CLOUD_PROVIDERS_STRICT=true` boots that same `aws` value normalized. A bad `GUARD_IP_WHITELIST` entry fails the same legible way. Gateway suite 337 passed, 1 xfailed; repo gates green.

One library behavior change surfaced in the bump and is worth flagging rather than burying: guard-core now sets `Retry-After` on every 429, so the gateway test that told the per-token limiter apart from the guard limiter by that header's absence now discriminates on its value instead (`1` for the per-user limiter, the rate window for guard).

DmitriyG228:
Superseded by #1324 — same author, strict superset. #1324 carries the identical `fastapi-guard>=7.8.1,<8.0` / guard-core 3.15.0 pin and the same `GUARD_BLOCK_CLOUD_PROVIDERS_STRICT` boundary validation, plus the WS consolidation onto upstream `guard_websocket` and the `GUARD_AUTO_BAN_RATE_LIMIT` knob. Every file this PR touches, #1324 touches too.

Neither has landed: main still pins `fastapi-guard>=7.2,<8.0`. Closing this so the guard bump has one lane — #1324 is the one we'll take.

Thanks @rennf93.

Issue #1109, opened 2026-08-10
[DEV] Bot dispatched at a nonexistent meeting ID reports "stopped by the user" — false cause shown to developers in their first API call


DmitriyG228:
Four bots this morning recorded `last_error` = **"stopped by the user at requested
(never admitted)"** with a total lifetime of **0.8–1.0 seconds** between `created_at`
and `updated_at`:

| meeting | user | created | lifetime | recorded reason |
|---|---|---|---|---|
| 25819 | 3636 | 04:43:17 | **0.9 s** | stopped by the user at requested (never admitted) |
| 25820 | 3636 | 05:19:31 | **0.9 s** | stopped by the user at requested (never admitted) |
| 25821 | 3636 | 05:25:10 | **1.0 s** | stopped by the user at requested (never admitted) |
| 25823 | 3758 | 05:31:38 | **0.8 s** | stopped by the user at requested (never admitted) |

A human does not click start and then stop within nine tenths of a second, three
separate times, forty minutes apart. Whatever ends these bots, "the user stopped it"
is not it.

Contrast with a genuine user withdrawal in the same window — meeting 25824, same user
3758, lifetime **44.5 s**, reason *"stopped while awaiting admission (withdrew the
join…)"*. That one is plainly real. The sub-second class is not.

## Why it matters

The failure reason is the primary input to bot-failure triage. If a code path that is
not a user action writes "stopped by the user", those failures are **silently
classified as user intent and excluded from investigation** — they look like customers
changing their minds rather than a defect. This is the same shape as the
silent-failure class that dominated the 0.12.21 retro: the system reports something
plausible instead of failing loudly.

## Scope — real but not a regression

Sub-2-second failures over the last 14 days:

| day | count | distinct users |
|---|---|---|
| 2026-07-27 | 4 | 1 |
| 2026-07-29 | 2 | 1 |
| 2026-07-30 | 1 | 1 |
| 2026-08-06 | 1 | 1 |
| **2026-08-10** | **5** | **2** |

**13 of 430 failures (3%)** across 14 days. Today is the highest single day but Jul 27
had 4, so this is a **standing low-rate defect, not something the 0.12.21 cutover
introduced.** Filing it as scope intake, not an incident.

## Billing impact: none — verified

All four were recorded `billing_disposition = not_billed`,
`not_billed_reason = bot_never_joined`, value `0`. Each did receive an `admit`
service-authority decision, so the admission path worked; the bot died after it. No
[...]

PR #1151, opened 2026-08-14
feat(calendar,teams): stage multi-feed and CSRC transcription

DmitriyG228:
### Calendar #1150

- support up to 10 independent ICS calendar connections per hosted user
- create, list, update, sync, and disconnect each calendar through plural Calendar APIs
- deduplicate shared meetings while retaining every calendar source
- store a redacted, JSON-safe VCALENDAR/VEVENT metadata snapshot on each meeting
- carry attendees, recurrence context, source identity, auto-join preference, and per-calendar bot name
- reconcile edits, postponements, cancellations, and disconnected-source cleanup

### Microsoft Teams #499

- remove Pyannote from the Teams path and route raw CSRC activity into persistent virtual channels
- feed each virtual channel through the shared GMeet-compatible window with prompt feedback,
  LocalAgreement, full-text fallback, timeout persistence, and terminal flush
- smooth short CSRC flicker while preserving real overlap as concurrent virtual channels
- keep unknown identities as stable Speaker A/B labels and reject uncanonical media/topic labels
- mark exact post-confirm duplicated phrases in API text as
  `⟦words⟧{CSRC A↔CSRC B}` using shared routed PCM, same text, and nearby Whisper word times
- retain both contested rows without choosing, deleting, or reassigning a speaker
- render contested API markers as ordinary unresolved transcript text in Terminal

## Exact source and release bundle

- Calendar/Teams implementation head: `756d91952114f70b7344da108f7632a2b7f43f95`
- DCO remediation head: `50e7efbb289b305c99ece2b1e4d056793ada83c8`
  - this is a tree-identical, metadata-only remediation commit by the original author
- frozen combined release-candidate head: `272b5117629666f14148f20aa4a0f8469d1f831d`
- frozen manifest: `releases/v0.12.23/candidate-images.json`
- manifest SHA-256: `5ad5bd86f5eb8e8a9fe48ef921189998b81fd09af621375b2bf7539d00c37562`
- final validation: [GitHub Actions run 31951789671](https://github.com/Vexa-ai/vexa/actions/runs/31951789671) — success on the frozen combined candidate

These three heads have different jobs: `756d9195…` identifies the Calendar/Teams implementation,
`50e7efbb…` adds only the original author's DCO attestation, and `272b5117…` is the frozen
cross-PR release candidate whose immutable images are being carried through the release train.

## Evidence

- Calendar admin API: 10 passed
- Calendar meeting API focused/edge/metadata/auto-join: 69 passed
[...]

PR #1160, opened 2026-08-14
fix(meetings): prevent stale stop attribution on continued runs

yaraPB:
> 👋 **Highly recommended (not required): say hi on [Discord](https://discord.gg/Ga9duGkVz9)** and
> tell us, in a sentence or two, **what** you’re changing and **why** — the story behind the diff.
> It makes reviewing your value bundle faster and connects you to the reporter and the maintainers.
> Your PR is judged on its evidence, not on whether you show up — but showing up helps a lot.

**Delivers issue:** #1109

## Contribution rights

- **C1 — Python attribution source:** traced `synthesize_terminal_for_dead_workload` in `reconcile.py`; saw the exact `stopped by the user at <status> (never admitted)` string selected from persisted `stop_requested`; concluded the issue’s wording is generated by the Python fallback, not API formatting.
- **C2 — Stop-state writers:** traced the user stop route and service-authority adapter; saw the current checkout has no fresh-invalid-meeting writer that sets `stop_requested`; concluded invalid Google Meet detection alone cannot explain the exact user-stop wording.
- **C3 — Reuse reproduction:** seeded a terminal meeting with `stop_requested=True`, continued it, and drove a replacement workload failure; before the patch the replacement could inherit the marker, while the patched path cleared it; concluded stale continuation state is a real reproducible cause.
- **C4 — Correlation audit:** traced workload construction and SQL lookup; saw workload IDs use a fresh connection UUID and lookup uses exact `bot_container_id`, while the latest-session lookup is not explicitly tied to a workload; concluded placeholder native IDs do not directly collide, but session correlation remains a hardening concern.
- **C5 — Browser/runtime boundary:** traced Google Meet’s `Meeting not found` selector and the runtime callback; saw the browser-side diagnosis can be collapsed into a generic rejection and runtime `exitCode`/`stopReason` are not passed into synthetic attribution; concluded those are separate follow-up improvements, not proven triggers for all four issue rows.

## Acceptance floor

The issue does not publish a formal acceptance table, so the rows below map the evidence-backed scope of this PR.

| Row | Evidence |
|---|---|
| A1 — continued runs do not inherit a previous stop marker | Base `e0b356d6`; head `09392db15362c0b49ff07d7a87f781b1ddd7fa01`; `tests/test_continue_meeting.py`: **5 passed**. The regression test drives a replacement workload to terminal while `awaiting_admission` and verifies a non-user reason. |
| A2 — genuine user-stop behavior remains distinct | Existing stop-route/lifecycle coverage remained green in the previously run narrow suite: **43 passed**. The patch only resets state when starting a new attempt; it does not change the user-stop writer or classifier. |
| A3 — fresh invalid-meeting behavior is not falsely attributed by this checkout | Source-level negative control: Google Meet rejection paths do not write `stop_requested`; the Python fallback without that marker chooses `join_failure`/`awaiting_admission_timeout`, not `stopped`. A production end-to-end claim is intentionally not made because the issue provides no row history or runtime/bot trace. |
| A4 — exact production trigger for all four reported rows | **Not claimed.** The issue’s public data lacks `meetings.data`, transition history, runtime event payloads, bot callbacks/logs, and deployed image provenance. The investigation document lists the evidence needed to resolve this. |

## Docs diff (D6c)

Added `ISSUE-1109-TECHNICAL-INVESTIGATION.md` as the PR’s technical evidence artifact. No public quickstart, how-to, reference, or concept page needed an update because this change is lifecycle-state hardening rather than a user-facing API/documentation change.

- Dependency changes: none; no dependency or lockfile files were modified.
- Focused secrets scan over all four committed files: passed — no high-confidence private-key, GitHub-token, AWS-key, or bearer-token patterns found.
- Diff integrity: `git diff-tree --check --no-commit-id -r HEAD` passed.
- Repository license gate: attempted with `node scripts/gates.mjs licenses`, but the existing gate crashed before evaluating the diff with `TypeError: pkgs.map is not a function` in `scripts/gates.mjs:393`. This is reported as a gate defect, not marked green.

## Validation request

Please have a maintainer or the issue reporter review the continuation scenario and, if possible, compare one fresh synthetic/nonexistent meeting against its database transition trail, runtime event, and bot callback. Deployment validated here: local meeting-api Python test harness only, on head `71b123878547a24cf358acd4c693a78195d28c70`; no Lite, full compose, Kubernetes, or hosted deployment claim is made.

## Authorship
Tooling disclosure (optional, welcome, never an attribution): local Python/pytest harness, repository static gates, and focused source/data-flow inspection.

### DIFF
--- a/ISSUE-1109-TECHNICAL-INVESTIGATION.md
+++ b/ISSUE-1109-TECHNICAL-INVESTIGATION.md
@@ -1,0 +1,456 @@
+# Issue #1109 — Technical Investigation
+
+## Scope
+
+This document records the deep technical investigation of [Vexa Issue #1109](https://github.com/Vexa-ai/vexa/issues/1109): a bot dispatched to a nonexistent meeting can be reported as **“stopped by the user”**, even though the user never issued a stop request.
+
+The investigation was performed on branch `codex/issue-1109`, based on `origin/main`. It covers the Python/FastAPI meeting service, the meeting lifecycle reconciler, the bot-spawn adapters, the Google Meet join module, and the runtime callback boundary.
+
+This is an engineering investigation, not a claim that the current branch completely fixes the issue.
+
+## Executive summary
+
+The misleading message is generated in the Python meeting lifecycle reconciler. When a workload disappears before it is admitted to a meeting, the reconciler creates a synthetic terminal event. If the meeting row contains `stop_requested=True`, it converts that flag into:
+
+```text
+stopped by the user at <status> (never admitted)
+```
+
+That behavior is correct for a real user-initiated stop, but the investigation found two separate problems and one important limit on what can be proven from this checkout:
+
+1. **Stale stop state can survive when a terminal meeting is continued.** The old meeting row is reused, and its row-level `stop_requested` flag can be carried into the replacement workload. This can make a later workload look like a user-stopped workload. The current branch fixes this state-reset problem and adds regression coverage.
+
+2. **A clean fresh invalid-meeting failure does not set `stop_requested` in the current code.** The exact `requested` + “stopped by the user” combination therefore requires either a user-stop write, stale state from a reused row, a race involving a stop request, or behavior from a deployed version/configuration not represented by this checkout. The public issue data does not include the affected rows’ `data`, transition history, runtime event, or bot log, so the exact production trigger cannot be proven from the issue alone.
+
+3. **There is a separate, confirmed error-classification bug.** The Google Meet join code detects pages such as “Meeting not found,” but maps the generic `error_page` result to **“Bot admission was rejected by meeting admin.”** This is misleading, but it is not the same string as #1109’s “stopped by the user” message and should not be presented as the proven trigger for #1109.
+
+The strongest complete solution therefore needs the Python hardening already added, better attribution of stop intent, and better propagation/classification of the actual join failure. The Python service should never infer a user stop solely from an ambiguous, stale, or system-generated boolean flag.
+
+## Issue behavior
+
+The issue reports four short-lived workloads attached to meeting IDs `25819`, `25820`, `25821`, and `25823`. They lived for roughly one second and were reported with a message similar to:
+
+```text
+stopped by the user at requested (never admitted)
+```
+
+The issue contrasts those failures with a genuine user withdrawal, which lasted much longer and was reported as a stop while awaiting admission. The issue also states that the failed nonexistent-meeting cases were not billed because the bot never joined.
+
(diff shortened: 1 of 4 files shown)


DmitriyG228:
On the rights declaration — you selected **Unsure**, so here is the determination, which is usually two questions:

1. Was this contribution assigned work for an employer or client?
2. Was it prepared using assets an employer or client controls?

If **no to both**: edit the PR description and select **Independent** — that plus DCO sign-offs on your commits completes the gate.

If **yes to either, or still unsure**: email [dmitry@vexa.ai](mailto:dmitry@vexa.ai) with this PR's URL and we'll run the private review — nothing personal goes on the public PR. Technical review continues either way.

One mechanical item: the branch is behind main — a rebase will refresh CI.

yaraPB:
> On the rights declaration — you selected **Unsure**, so here is the determination, which is usually two questions:
> 
> 1. Was this contribution assigned work for an employer or client?
> 2. Was it prepared using assets an employer or client controls?
> 
> If **no to both**: edit the PR description and select **Independent** — that plus DCO sign-offs on your commits completes the gate.
> 
> If **yes to either, or still unsure**: email [dmitry@vexa.ai](mailto:dmitry@vexa.ai) with this PR's URL and we'll run the private review — nothing personal goes on the public PR. Technical review continues either way.
> 
> One mechanical item: the branch is behind main — a rebase will refresh CI.

1. No it was my own but I am under an internship contract right now that's why I put unsure
2. No

DmitriyG228:
Delivered — differently from the approach here, and shipped. Closing with credit.

The planned-row half is **literally the line you wrote**: `planned.pop("stop_requested", None)` is in main at `bot_spawn/adapters.py:548` and mirrored in `bot_spawn/fakes.py:165`.

The `reopen_meeting` half went stricter than clearing the flag. A row the user stopped is now refused for reuse outright, under the row lock (`bot_spawn/adapters.py:185-186`, gated at `service.py:547`), answering `409` with *"meeting {id} was stopped by the user and cannot be continued — POST /bots again (without continue_meeting) to start a new bot"*. A stale marker cannot be inherited by a continued run because the continued run no longer happens.

Carried by `d0216ece` (#1151). Rung: **released in v0.12.24, installed in production.**

#1109 stays open for the residual: as your own investigation notes, the reopen path was never proven to be the production trigger for those four sub-second failures.

Thanks @yaraPB — and the Google Meet `error_page` misclassification your investigation surfaced (mapping to "admission rejected" instead of "meeting not found") is a distinct defect that deserves its own issue. Please file it.

PR #1164, opened 2026-08-16
feat(meeting-api): record-keyed transcript access, explicit empty state, honest list filters

DmitriyG228:
**Delivers issue:** none yet in this repo — the work originates from a read-only production sweep tracked in the private business workspace. That is a gap, not an omission to ignore: if this direction is wanted, it should get an issue here to hang acceptance on. **Draft: not ready to merge**, and the "What is NOT done" section below says why.

## Problem — measured, not inferred

A read-only sweep of one production account on 2026-08-16 (`GET /meetings` ×8 pages, `GET /transcripts/{platform}/{native}` ×286; nothing created, updated or deleted) asked what an archive consumer can actually retrieve through this API.

| | |
|---|---|
| Meetings enumerated | 747 (2025-08-26 → 2026-08-13) |
| Records with **no** `native_meeting_id` | **461 — 61.7%**, addressable by no route |
| 286 native fetches resolve to distinct records | 199 |
| …of which carry any segment | **166 — 22.2% of the archive is net retrievable** |
| Links carrying more than one record | **46 collapse sets** over 132 meetings; largest **8 records → 1 retrievable** |
| Fetches answering `200` with `segments: []` | **60 of 284** |

Three distinct defects produce that number, and none of them is a data problem.

**1. Records are not addressable.** The only transcript route is keyed by `(platform, native_meeting_id)` — but a native id is a **join link**, not a record key, and the two are not one-to-one. 461 records carry no native id at all. Two carry a native id that is itself a URL (`https://teams.live.com/meet/9366473044740?p=…`); interpolated into the path, its `/` and `?` reshape the request, so it 404s permanently — the gateway stores a value it cannot key on. And 46 links carry several records, of which only the newest resolves (45 of 46 unambiguously; the 46th is the URL-shaped pair, which resolves to nothing). One recurring link hides **seven** transcripts behind the eighth.

`GET /meetings/{id}/transcript` answered **405**, and the mechanism is worth naming: `/meetings/{a}/{b}` matched the native-keyed `PATCH`/`DELETE` pair with `platform="15496"`, `native_meeting_id="transcript"`, and that pair registers no `GET`. The by-row read already existed — as `GET /transcripts/by-id/{id}` — it was simply not where anyone looks for it.

**2. Empty is silent.** 60 meetings answer `200` with `segments: []`. They are not errors; they are empty. A consumer that trusts the status code writes empty rows into its archive and never learns why.

[...]

PR #1168, opened 2026-08-16
fix(meeting-api): match meeting hosts exactly, not by substring

DmitriyG228:
**Delivers issue:** none — found while dogfooding the link parser.

## Contribution rights

## What was wrong

`collector/meeting_link.py::parse_meeting_url` classified the meeting platform with **substring
tests on the hostname**:

```python
if "meet.google.com" in host:      # also true for meet.google.com.<anything-else>
if "zoom" in host:                 # also true for any host with "zoom" in it
if "teams.microsoft.com" in host:  # also true for teams.microsoft.com.<anything-else>
```

A hostname *contains* the platform's domain whenever the platform's domain is a prefix of it, and
the registrable part of a hostname is on the right, not the left. So a host that belongs to
whoever submitted the URL satisfied these tests and was accepted as a genuine platform link.

This is not cosmetic. `parse_meeting_url` is what decides whether a submitted URL becomes a
meeting; the URL rides along as `meeting_url`, and a bot later opens it in a real browser. An
accepted host is therefore a host we navigate to. On a hosted deployment the submitter is any API
caller, so the host reached was theirs to choose rather than ours — including hosts that resolve
somewhere inside the deployment's own network. Matching the host exactly is the fix for all of it.

The same class was flagged by CodeQL on the mailroom branch, which is where the fix below comes
from.

## The fix

`_host_matches(host, *domains)` — equality, or a dotted-suffix subdomain. Ported from the
mailroom's vendored copy (`vexa_mailroom.meeting_link`), not reinvented, so **the two parsers now
agree**; that matters because the mailroom hands its extracted `meeting_url` to `POST /meetings`,
which parses it a second time with this copy.

Zoom's domains are now named (`zoom.us`, `zoomgov.com`) instead of inferred from the substring
`"zoom"`. That is the rule the join brick already applies
(`modules/join/src/index.ts::resolvePlatform` — `host === "zoom.us" || host.endsWith(".zoom.us")`);
this parser was the outlier.

`_is_platform_lookalike` closes the same gap in the self-hosted jitsi fallback. A host of the form
`meet.google.com.<other-domain>` carries a `meet` *label*, so after the exact-match rule refused
it, the jitsi naming heuristic would have adopted it as a self-hosted room and handed the bot the
same host by another route. Hosts an operator has explicitly declared in `VEXA_JITSI_HOSTS` are
deliberately exempt — naming your own deployment is a decision, not a guess.

### Preserved

| Shape | Still parses |
|---|---|
| `meet.google.com/abc-defg-hij` | google_meet |
| `zoom.us`, `us02web.zoom.us`, `us05web.zoom.us`, `company.zoom.us`, `zoomgov.com` | zoom |
| `teams.live.com/meet/<id>`, `teams.microsoft.com/meet/<id>`, `foo.teams.microsoft.com` | teams |
| `teams.microsoft.com/l/meetup-join/19:meeting_…@thread.v2` | teams |
| `meet.jit.si/<room>` | jitsi |
| `jitsi.example.org/<room>` (jitsi-named host, pasted links) | jitsi |
| `meet.example.org`, `eu.meet.example.org` (the `meet` LABEL rule, pasted links) | jitsi |
| any host in `VEXA_JITSI_HOSTS`, in strict mode and in the ICS scan | jitsi |
| bare `abc-defg-hij` / bare 9–11 digits | google_meet / zoom |

### Now rejected

Hosts that merely contain a platform's domain, in both the pasted-link and ICS-scan modes:
`meet.google.com.<other>`, `zoom.us.<other>`, `teams.microsoft.com.<other>`,
`teams.live.com.<other>`, and hosts that merely have `zoom` somewhere in the name
(`zoom.<other>`, `notzoom.com`). `meetings.example.org` stays rejected as before — `meet` must be
a whole label.

The change only ever *narrows* what is accepted, so nothing the mailroom accepts is rejected here.

- **C1 — the defect is real on `main`.** ran: the new `TestHostMatchingIsExact` cases against the
  unmodified parser (`git stash` on the source file only) · saw: **8 failed, 14 passed** — every
  lookalike host in the table above parsed as a genuine platform link, e.g.
  `meet.google.com.<other>/abc-defg-hij` → `("google_meet", "abc-defg-hij")` · concluded: the
  substring tests, not an edge case in the callers, are what accepts these.
- **C2 — the fix closes it without collateral.** ran: `uv run pytest -q` in
  `core/meetings/services/meeting-api` · saw: **936 passed, 5 skipped** (was 924 before the 12 new
  cases; every pre-existing jitsi, zoom, teams and calendar-sync assertion unchanged) · concluded:
  the narrowing costs none of the legitimate shapes the suite already pins.
- **C3 — the two parsers agree.** ran: read `vexa_mailroom.meeting_link` and its
  `tests/test_meeting_link.py` goldens before writing anything · saw: the same `_host_matches`
  helper and the same lookalike expectations · concluded: ported rather than reinvented, so the
  mailroom's second-parse contract holds.

## Acceptance floor

| Row | Evidence |
|---|---|
| Lookalike platform hosts resolve to no platform | `TestHostMatchingIsExact::test_lookalike_host_resolves_to_no_platform` — 7 hosts × (default mode, `generic_hosts=True`, free-text scan). Negative control in C1: all 7 fail red on the unfixed parser. |
| Every legitimate host shape still parses | `test_real_platform_hosts_still_parse` (10 cases) + `test_self_hosted_jitsi_conventions_survive` (4 cases) + the untouched `TestParseJitsi` / `TestConfiguredJitsiHosts` / `test_parse_meeting_url_formats`. |
| `VEXA_JITSI_HOSTS` still wins | `test_declared_jitsi_host_is_honoured_even_if_it_looks_like_a_platform`. |
| Full suite green | `uv run pytest -q` → 936 passed, 5 skipped. |
| Repo gates green | The pre-push hook's 14 static gates all green: readme, dataflow, isolation, isolation-py, exports, graph, graph-py, schema, contract-version, config-contract, licenses, execution-env, test-isolation, contract-conformance. Full suite rides CI. |

## Docs diff (D6c)

Changelog fragment in `docs/changelog.d/`. No reference page changes: the accepted host shapes are
the ones the docs already document — this removes shapes that were never documented as supported.

No dependency, licence or generated-artifact surface is touched — the diff is one pure-string
module plus its tests. Static gates listed above ran green locally. CodeQL and the full `gates`
suite ride CI on this PR.

## Validation request

Any competent non-author reviewer, on the diff itself: the parse table above is the whole contract,
and the test file is written to be read as that table. What to watch is the **preserved** column —
the risk in this change is over-narrowing, not under-narrowing, so a reviewer who runs a real
customer's Zoom or Teams link through `parse_meeting_url` and gets the same answer as before has
checked the thing that matters. No deployment was exercised: pure string logic, no I/O, no
framework imports.

## Notes for the maintainer

- **Contribution-rights boxes are deliberately left unticked** — that selection is a legal
  certification and is the submitter's to make, not an agent's.
- **The commit carries no `Signed-off-by`** for the same reason; `.github/dco.yml` allows
  individual remediation.
- **Overlap with #1164 (`mk-stage0-connector`):** none at file level. #1164 touches
  `collector/{app,adapters,fakes,ports}.py` and `tests/test_connector_access.py`; this PR touches
  `collector/meeting_link.py` and `tests/test_meeting_link.py`. Same package, disjoint files, both
  cut from current `origin/main`. `app.py` calls `parse_meeting_url`, but its signature is
  unchanged and the behaviour only narrows, so the two should merge in either order.

### DIFF
--- a/core/meetings/services/meeting-api/src/meeting_api/collector/meeting_link.py
+++ b/core/meetings/services/meeting-api/src/meeting_api/collector/meeting_link.py
@@ -7,6 +7,10 @@ DESCRIPTION). Pure string logic — no I/O, no framework imports; the one config
 ``VEXA_JITSI_HOSTS`` (P14, declared in config.v1.json), consulted per call so tests and
 reloads see the live env.
 
+Hosts are matched EXACTLY or as a dotted subdomain (``_host_matches``) — never by substring,
+which would read ``meet.google.com.attacker.example`` as Google Meet. ``vexa_mailroom``'s
+vendored copy carries the same helper; the two parsers are meant to agree.
+
 Id formats (mirrors the dashboard join-form):
   * google_meet → ``abc-defg-hij``
   * zoom        → 9–11 digits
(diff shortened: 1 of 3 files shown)


DmitriyG228:
Draft — technically complete, blocked only on the submitter.

All build checks green: `gates` (static · python · node), `lite-smoke`, `value-fsm`, `docs-current`, CodeQL. Three checks are red on purpose: `DCO`, `contribution-rights` and `evaluate` all need the rights selection and the `Signed-off-by` line, which are the submitter's certification to make.

One note for whoever reviews the run history: the first `static` attempt failed in `scripts/sbom.test.mjs` ("fixture setup did not alter Dockerfile.lite"). That file is not in this diff, the two tests pass locally on this branch, and a plain re-run went green — transient, not a signal about this change.

PR #1177, opened 2026-08-17
terminal: a meeting you can link to — /meetings/<id>

DmitriyG228:
## What

The Terminal's meeting view is now addressable: **`/meetings/<meeting id>`**. Opening a meeting puts its
id in the URL; loading that URL directly — hard reload, pasted link, fresh session — restores the same
meeting view; an unknown or foreign id renders a clean not-found state.

## Today's behaviour (what this changes)

The terminal is a **single route** (`/`). Which meeting is open lives only in the dockview layout in
`localStorage` (`vexa.terminal.dock.v3`), so a hard reload restores the tab *in that browser* while the
URL references nothing — a meeting cannot be pasted to a colleague, bookmarked, or reopened in a fresh
session. The only URL that ever reached a meeting was a `?tshare=` token, and that flow deliberately
strips itself back to `/` after redeeming (it stashes the id in `localStorage` for the first-view
resolver). An id absent from the meetings list also rendered a permanent "Connecting…" header.

## Design

- **The reference is the meetings-domain ROW id** — the id the client already keys the tab, the
  row-keyed live streams (`tc:meeting:{id}`) and `GET /api/transcripts/by-id/{id}` by. **No API change
  is implied**: the public transcript API still keys by `platform/native_meeting_id`; the route only
  carries the id the existing client data paths already resolve.
- **One shell, two routes.** `src/app/meetings/[meetingId]/page.tsx` renders the same `<App/>` as `/`.
- **The URL follows the active tab; it does not drive it.** In-app navigation (row click, preview slot,
  restored dock) is untouched and stays the interactive path. A `history.replaceState` effect mirrors the
  active meeting tab into the address bar — `replaceState`, not push, since the workbench keeps its own
  history under Escape / Alt+Left — and only ever rewrites the two shapes it owns (`/`, `/meetings/<id>`).
- **Landing goes through the existing first-view resolver**, extended with `routeMeetingId`: ranked below
  a just-redeemed share (that stash belongs to *this* navigation), above everything else, and applying to
  a returning user, because opening the address is the explicit act. A passively-mounted shared workspace
  does not ride along; a just-accepted invite does.
- **An unknown id is an answer, not a spinner** — gated on the meetings list having answered at least
  once (`useLiveMeetingsLoaded`), so an offline list can never make a live meeting look deleted.
- OAuth already round-trips `pathname + search`, so a meeting URL survives sign-in.

## Files

| File | |
|---|---|
| `clients/terminal/src/app/meetingRoute.ts` | new — pure path ⇄ id contract (bounded id charset, no extra segments) |
| `clients/terminal/src/app/meetings/[meetingId]/page.tsx` | new — the route, rendering the same `<App/>` |
| `clients/terminal/src/workbench/firstView.ts` | `routeMeetingId` input + its priority |
| `clients/terminal/src/workbench/Workbench.tsx` | reads the route id at first render; URL-sync effect |
| `clients/terminal/src/surfaces/liveMeetings.ts` | `useLiveMeetingsLoaded()` — has the list answered yet |
| `clients/terminal/src/surfaces/meeting.tsx` | `meetingResolution()` + `MeetingNotFound` |
| `clients/terminal/DECISIONS.md` | D7 |

## Tests

New: `src/app/__tests__/meetingRoute.test.ts` (round-trip, encoding, trailing slash, rejected ids /
extra segments / malformed escapes, which paths may be rewritten) · `src/surfaces/__tests__/meetingNotFound.test.tsx`
(resolving vs not-found vs resolved, and the rendered dead end). Extended: `src/workbench/__tests__/firstView.test.ts`
(7 routed-landing cases).

```
vitest run   → 56 files, 424 passed
tsc --noEmit → clean
next build   → clean; ƒ /meetings/[meetingId] present
gate:readme  → green
```

Not merged. No changes outside `clients/terminal/`.

### DIFF
--- a/clients/terminal/DECISIONS.md
+++ b/clients/terminal/DECISIONS.md
@@ -92,3 +92,28 @@ A pass to make the first-run + workspace experience real (no mocks, no dead UI):
   folds the meeting's live transcript (redis `tc:meeting:{native}`) into the prompt server-side, fresh each
   turn. The client adds no prompt preamble and points at no notes file. (Replaced the earlier notes-file
   workaround, itself a stand-in for the unwired `meeting.read_transcript` MCP tool that hung the worker.)
+
+## D7 — the open meeting is addressable: `/meetings/<id>`
+Until now the terminal was a **single route** (`/`). Which meeting was open lived only in the dockview
+layout in `localStorage` (`vexa.terminal.dock.v3`), so a hard reload restored it *in that browser* while
+the URL referenced nothing: a meeting could not be pasted to a colleague, bookmarked, or reopened in a
+fresh session. The only URL that ever reached a meeting was a `?tshare=` token, and that flow deliberately
+strips itself back to `/` after redeeming.
+
+- **The reference is the meetings-domain ROW id** — the same `id` the client already keys the tab, the
+  live streams (`tc:meeting:{id}`) and `GET /api/transcripts/by-id/{id}` by. **No API change is implied.**
+  The public transcript API still keys by `platform/native_meeting_id`; the route carries only the id the
+  existing client data paths already resolve, so nothing about fetching moved.
+- **One shell, two routes.** `src/app/meetings/[meetingId]/page.tsx` renders the same `<App/>` as `/`.
+  A second app would drift; a route that only *deep-links* would fork the layout logic.
+- **The URL follows the active tab, it does not drive it.** In-app navigation (row click, preview slot,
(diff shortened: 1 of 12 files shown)


DmitriyG228:
Absorbed into the 0.12.23 train candidate (codex/1150-calendar-multi) by founder ruling — one train, no parallel lanes. The commits are in the candidate via merge 0ffbdae1 and ship in rc.12; this PR closes as integrated, not rejected.

PR #1178, opened 2026-08-17
fix(terminal): a refused Join says why, instead of "your key doesn't have access"

DmitriyG228:
## What changed

The Terminal client reached parity with the hosted dashboard on paywall and transparent join-denial errors.

`meeting-api` refuses a bot with `403 {"detail":{"code":"service_not_allowed","reason":…,"decision_id":…}}` and reports its own outage as `503 service_authority_unavailable` ([`bot_spawn/router.py`](https://github.com/Vexa-ai/vexa/blob/main/core/meetings/services/meeting-api/src/meeting_api/bot_spawn/router.py#L425-L441)). The terminal read neither.

### What a 403 said before this PR

| Surface | What the user saw on a paywall |
|---|---|
| Sidebar **Add bot** (`meeting.tsx` `MeetingsList`) | "Your key doesn't have access to this." — `presentError`'s flat 403 line |
| Row **Send now** (`actionsFor` → `presentMeetingActionFailure`) | same flat line, auto-cleared after 6s |
| Onboarding **drop-a-bot card** (`meetingsOnboarding.tsx`) | `Couldn't send (403): {"code":"service_not_allowed","reason":"insufficient_balance"}` — the raw payload |
| Prep tab **Send now** (`meetingPrep.tsx`) | "Something went wrong — details are in the browser console." (body read as text, rethrown as a bare `Error`, so the JSON-shaped message failed `presentError`'s `isProse` check) |
| Any of them, on the 503 | "The request failed (503)." |

An empty prepaid balance was indistinguishable from an outage and from a permissions fault.

### What it says now

`clients/terminal/src/surfaces/serviceDenial.ts` maps the seven-reason vocabulary — `allowed`, `billing_setup_required`, `insufficient_balance`, `payment_past_due`, `spend_cap_reached`, `concurrency_limit_reached`, `billing_unavailable` — to copy kept **byte-aligned** with `services/dashboard/src/lib/service-denial.ts` and with `Vexa-ai/vexa-platform` `services/webapp/apps/webapp/lib/service-denial-view.ts`, and classifies each into what the customer can DO: `paywall` / `setup` / `limit` / `retryable` / `unknown`.

- An **unmapped** reason renders VERBATIM (`service not allowed: <code>`) with a `console.error` **in production too** — an unmapped reason is a prod-only event by definition (it means the authority shipped a reason ahead of this copy), so a dev-only log has no signal exactly where the drift happens.
- A **401**, and a **403 without the denial code**, keep their access-error rendering. Routing a real permission fault to the paywall would tell someone to pay for a problem money cannot fix.

`clients/terminal/src/surfaces/ServiceDenialPanel.tsx` renders it **in-flow, not as a transient line** — a paywall is a thing the user must act on, and the sidebar's error line cleared itself after 5s (the meeting row's after 6s); both are now sticky when the failure is a denial. Styling is the terminal's own idiom (CSS-var semantic tones, ui-kit `Icon`), not the dashboard's Tailwind: `--warn` carries `paywall`/`setup` because the account is fine and needs money, and only an unmapped reason — a bug — gets `--danger`.

### The plumbing that made it possible

`ApiError` gains `body`: the parsed failure body as it came off the wire. It was already being flattened to a string for the operator channel, and **flattening is exactly what destroyed the distinction**. `getJson`'s error path moved into an exported `readApiFailure()` so the four join edges that call `fetch()` directly produce the same error object. The denial reader also parses a JSON-shaped `detail` string, so a call site predating the structured field still renders transparently rather than silently regressing to "access denied".

## The pin now has THREE twins — please reconcile at merge time

`clients/terminal/src/surfaces/__tests__/serviceDenialVocabulary.test.ts` is the twin of the dashboard's `services/dashboard/tests/service-denial-vocabulary.test.ts`. Its failure message names all **six** places a reason must be added, because there are now **three** copy modules across two repos, not two:

1. `Vexa-ai/vexa` `clients/terminal/src/surfaces/serviceDenial.ts` (new)
2. `Vexa-ai/vexa` `services/dashboard/src/lib/service-denial.ts`
3. `Vexa-ai/vexa-platform` `services/webapp/apps/webapp/lib/service-denial-view.ts`

**The two older pins' failure messages still say FOUR places**, because they predate this module. Updating them was deliberately left out of this PR to keep it to one surface — the three message lists should be reconciled when this merges, alongside #1176 and `Vexa-ai/vexa-platform#292`.

The deeper fix remains narrowing `reason` to an enum in `core/meetings/contracts/service-authority.v1`, which is SEALED and types `Decision.reason` as an opaque string. That is a breaking change to a frozen `.vN` and needs a `service-authority.v2` on a `lane:contract` PR, not an edit to v1.

## Verification

- `npx vitest run` (`clients/terminal`) — **55 files, 411 tests, all pass**, including 12 new vocabulary/access-boundary assertions and one new row-action denial test.
- `npx tsc --noEmit` — clean.
- `npx next build` — clean.
- `pre-push` fast gate suite (14 static gates) — green.

One pre-existing test was updated: `meetingActions.test.tsx` pins the exact `MeetingActionFailure` shape, which gains the additive `denial` field.

### Note on `gate:schema`

The first push attempt aborted on `gate:schema`, which reproduces identically on a clean `origin/main` checkout with zero changes. The cause is environmental, not a schema violation: in a fresh worktree without root `node_modules`, every contract's `validate.mjs` dies with `Cannot find package 'ajv'`, and the gate reports it as an empty-bodied failure. Worth noting because the empty body makes it read as a real conformance failure — the gate swallows `ERR_MODULE_NOT_FOUND` into `""`. With root deps present the gate passes on all 25 contracts.

Refs `Vexa-ai/vexa-platform#291`
Refs #1176 (the pattern source — reason→copy mapping, kind system, in-flow panel, vocabulary pin)

### DIFF
--- a/clients/terminal/src/surfaces/ServiceDenialPanel.tsx
+++ b/clients/terminal/src/surfaces/ServiceDenialPanel.tsx
@@ -1,0 +1,86 @@
+"use client";
+/** ServiceDenialPanel — the in-flow rendering of a refused Join.
+ *
+ *  A panel, not a transient error line: a paywall is a thing the customer has to ACT on, and the
+ *  one-line red `⚠ …` the join surfaces use for "bad link" both loses the fix and (on the sidebar)
+ *  clears itself after 5s. Styling follows the terminal's existing inline-notice idiom — CSS-var
+ *  semantic colours, tinted wash + border, `Icon` from the ui-kit (cf. `workbench/OpsNotice.tsx`
+ *  and `canvas/MeetingHealthBanner.tsx`).
+ *
+ *  Twin of `services/dashboard/src/components/join/service-denial-panel.tsx`; the WORDS live in
+ *  `surfaces/serviceDenial.ts` and are shared with it, only the skin differs.
+ */
+import { Icon } from "../ui-kit";
+import { denialActionUrl, type ServiceDenialKind, type ServiceDenialPresentation } from "./serviceDenial";
+
+/** `--danger` is destructive/errors ONLY and `--warn` is attention-not-error (globals.css §semantic
+ *  set), so a paywall — the account is fine, it needs money — is warn, and only a reason this build
+ *  has never heard of (a bug) is danger. */
+const TONE: Record<ServiceDenialKind, { fg: string; bg: string; icon: string }> = {
+  paywall: { fg: "var(--warn)", bg: "var(--warnbg)", icon: "zap" },
+  setup: { fg: "var(--warn)", bg: "var(--warnbg)", icon: "gear" },
+  limit: { fg: "var(--accent)", bg: "var(--accentbg)", icon: "info" },
+  retryable: { fg: "var(--t2)", bg: "var(--panel2)", icon: "refresh" },
+  unknown: { fg: "var(--danger)", bg: "var(--dangerbg)", icon: "alert" },
+};
+
+export function ServiceDenialPanel({
+  presentation,
+  onRetry,
+}: {
+  presentation: ServiceDenialPresentation;
+  onRetry?: () => void;
+}) {
+  const { kind, title, body, action, retryable, reason } = presentation;
+  const tone = TONE[kind];
+
+  return (
+    <div
+      role="alert"
+      data-testid="service-denial-panel"
+      data-denial-kind={kind}
+      data-denial-reason={reason}
+      style={{
+        display: "flex", flexDirection: "column", gap: 5,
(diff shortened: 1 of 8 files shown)


DmitriyG228:
Absorbed into the 0.12.23 train candidate (codex/1150-calendar-multi) by founder ruling — one train, no parallel lanes. The commits are in the candidate via merge 0ffbdae1 and ship in rc.12; this PR closes as integrated, not rejected.

Issue #1197, opened 2026-08-17
[DEV] Completed meetings cannot be erased from the terminal UI — only via the API


DmitriyG228:
## Where we are (honest)

`vexa.ai/legal/terms` says transcripts are *"kept until you delete via API/UI"*; `/legal/security`
repeats it. #1188 delivers the **API** half — `DELETE /meetings/{id}` erases a completed meeting's
transcript rows and recording objects. There is no **UI** half.

In the terminal client, `actionsFor` in
[`clients/terminal/src/surfaces/meeting.tsx`](clients/terminal/src/surfaces/meeting.tsx) offers
`Delete` only on `idle`/`scheduled` rows — a plan-delete, on a meeting that has no data yet. Terminal
rows (`completed`/`failed`/`stopped`) offer exactly one action, `Re-send`:

```
case "completed": case "failed": case "stopped": default:
  return [{ id: "resend", label: "Re-send", tone: "accent", run: send }];
```

Asserted by
[`meetingActions.test.tsx`](clients/terminal/src/surfaces/__tests__/meetingActions.test.tsx):
`it("completed/failed/stopped → Re-send", …)`. So a user with a finished confidential meeting has no
way to erase it from the product they are looking at.

One good consequence of the current state, worth preserving: because no delete control exists on a
terminal row, #1188 cannot turn an existing button into an accidental irreversible erase. Any UI that
adds one must therefore carry its own confirmation.

## Acceptance

| Row | Observation |
|---|---|
| A1 | A `completed`/`failed`/`stopped` row offers a delete-artifacts action; a plan row's existing `Delete` is unchanged |
| A2 | The action is confirm-gated, and **declining issues no request at all** (a red test: remove the gate ⇒ this fails) |
| A3 | The confirmation names what is destroyed, says it cannot be undone, and does not imply backups can recover it |
| A4 | Accepting fires exactly `DELETE /api/meetings/{id}` — the same owner-scoped route #1188 implements |
| A5 | The destructive action is visually distinct from the non-destructive ones |

## Depends on

**#1188.** Until that merges, `DELETE` on a completed meeting answers `409` and this UI would surface
an error. This must land after it, never before.

PR #1188, opened 2026-08-17
feat: delete completed meeting artifacts

DmitriyG228:
Refs #116.

## Value

Owners can erase completed meeting transcripts and recording artifacts from Vexa's live primary stores while lifecycle evidence remains tombstoned. Unknown and foreign-owner resources are indistinguishable, active meetings are rejected, and failed storage deletion remains retryable.

This does not promise immediate erasure from snapshots, object-version history, or backups; those remain subject to deployment retention policy.

## Independent witness

- commit: `3ae320562534561d62c999486144aeddcb8e5874`
- tree: `20dbd9f234340da1020004e8fb41f1f0cbec61fc`
- meeting-api: 922 passed, 5 skipped
- gateway: 112 passed, 1 xfailed
- conformance: 86 passed
- focused oracle: 11/11
- isolated Compose/Postgres/Redis/S3-compatible black box: owner/non-owner/active/idempotency/storage-failure-retry/transcript+recording deletion/terminal tombstone all passed

## Contributor rights

- [x] Independent — standing founder declaration applies to Dmitriy's own Vexa contributions.
- [ ] Employer/client authorization required
- [ ] Unsure

Every commit carries Dmitriy's matching DCO sign-off.

PR #1198, opened 2026-08-17
feat(terminal): delete a completed meeting's artifacts from the UI (#1197)

DmitriyG228:
**Delivers issue:** #1197
**Stacked on #1188 — base is `codex/116-completed-artifact-delete`, not `main`.** Until that merges,
`DELETE` on a completed meeting answers `409` and this button would surface an error. Land it after,
never before. Retarget to `main` once #1188 is in.

Our legal pages say transcripts are kept *"until you delete via API/UI"*. #1188 delivers the API
half. Terminal rows offered only `Re-send`, so a finished confidential meeting could not be erased
from the product the user is actually looking at. This is the UI half.

## Contribution rights

The commit carries `Signed-off-by: DmitriyG228`, matching repository-local identity. **No box is
ticked — a human must select one.**

## What this changes

`completed`/`failed`/`stopped` rows gain a **Delete data** action firing the same owner-scoped
`DELETE /api/meetings/{id}` that #1188 implements. Three details are the actual design:

1. **The confirm resolves before `fetch` is constructed.** `fetch` is eager — building it inside the
   action and then deciding would fire the request regardless of the answer. A declined confirm must
   issue *no request at all*, and that is the property most likely to rot silently, so it has a test
   that fails if the guard is removed.
2. **The prompt does not overpromise recovery.** It names what is destroyed (recordings only when the
   row has one), says it cannot be undone, and states backups are not a recovery path — matching the
   boundary #1188's docs draw. A confirmation that implies a backup could restore this would be worse
   than none.
3. **A `danger` tone, rather than reusing `live`.** Both render in `--danger`, but `live` means "acts
   on a running bot" — reusing it for an irreversible erase on a terminal row would make the call
   site lie. The plan-row `Delete` deliberately keeps its `muted` tone: it removes a row nobody has
   data in yet, which is not the same act.

## Observation bundle

- **C1 — the offered set and the endpoint.** ran: `meetingActions.test.tsx` · saw: `23 passed` ·
  concluded: terminal rows now offer `["resend", "erase"]`, erase fires exactly
  `DELETE /api/meetings/{id}` with no body, and the plan `Delete` is untouched.
- **C2 — the guard is real, not decorative.** ran: deleted the `if (!ok) return;` line and re-ran ·
  saw: `× declining the confirm issues NO request at all` → `1 failed | 22 passed` · restored ⇒
  `23 passed`. concluded: the gate is load-bearing; a future refactor that drops it goes red rather
  than silently shipping a one-click irreversible delete.
- **C3 — no collateral.** ran: full terminal suite, then `gate:node` · saw: `54 files, 403 passed`
  and `18 package(s) · build + test green` · concluded: the added tone and action break nothing.

## Acceptance floor

| Row | Evidence |
|---|---|
| A1 | C1 — `["resend","erase"]` on completed/failed/stopped; plan rows unchanged |
| A2 | **C2** — the red→green pair on the guard itself |
| A3 | C1 — prompt asserted to name recordings conditionally, to say "cannot be undone", and to state backups are not a recovery path |
| A4 | C1 — `DELETE /api/meetings/{id}`, no body |
| A5 | C1 — erase is `danger`, plan `Delete` is `muted` |

## Docs diff (D6c)

`docs/changelog.d/` fragment. No reference page changes: #1188 already documents the endpoint's two
meanings and the backup boundary, and this PR adds no API surface — it puts an existing route behind
a button.

No dependency added. The destructive path is owner-scoped **server-side** by #1188 — this button
carries no authorization of its own and must not be read as adding any; it can only ask for what the
caller already owns, and an unowned id answers `404`.

The review question worth asking here is not "is the request right" but **"can this fire without a
human meaning it"** — that is what C2 pins.

## What is undone

1. **No live UI witness.** The behavior is proven at the unit layer; I have not run the terminal in a
   browser and clicked the button against a live stack. That is the honest gap in this bundle.
2. **`window.confirm`, not an in-app dialog.** It matches the surrounding code (`schedule` uses
   `window.prompt`) and is unbypassable, which is the property that matters for an irreversible act.
   A styled dialog is a real improvement, and a separate one.

## Validation request

**Preferred validator:** any competent non-author. **Deployment:** hosted, after #1188.

Watch two things: the confirm cannot be skipped, and **Cancel leaves the data intact** — check the
transcript still reads back afterwards, not just that the row looks unchanged.

## Rung

**PR open**, stacked on an unmerged PR. Not merged, not released, not in any image, not deployed,
not witnessed live.

### DIFF
--- a/clients/terminal/src/surfaces/__tests__/meetingActions.test.tsx
+++ b/clients/terminal/src/surfaces/__tests__/meetingActions.test.tsx
@@ -57,8 +57,48 @@ describe("actionsFor — offered action sets per status", () => {
       expect(ids(s)).toEqual(["stop"]);
     }
   });
-  it("completed/failed/stopped → Re-send", () => {
-    for (const s of ["completed", "failed", "stopped"]) expect(ids(s)).toEqual(["resend"]);
+  it("completed/failed/stopped → Re-send + Delete data", () => {
+    for (const s of ["completed", "failed", "stopped"]) expect(ids(s)).toEqual(["resend", "erase"]);
+  });
+  it("only the terminal erase carries the danger tone (the plan Delete must not)", () => {
+    expect(actionsFor(row("completed")).find((a) => a.id === "erase")!.tone).toBe("danger");
+    expect(actionsFor(row("idle")).find((a) => a.id === "delete")!.tone).toBe("muted");
+  });
+});
+
+describe("erase — a confirm gate stands in front of an irreversible delete (#116)", () => {
+  const eraseOn = (s: string) => actionsFor(row(s)).find((a) => a.id === "erase")!;
+
+  it("declining the confirm issues NO request at all", () => {
+    const confirmSpy = vi.spyOn(window, "confirm").mockReturnValue(false);
+    eraseOn("completed").run();
+    expect(confirmSpy).toHaveBeenCalledOnce();
+    expect(fetchMock).not.toHaveBeenCalled();
+  });
+
+  it("accepting the confirm DELETEs the row-id meeting route", () => {
+    vi.spyOn(window, "confirm").mockReturnValue(true);
+    eraseOn("completed").run();
+    const { url, init } = lastFetch();
+    expect(url).toBe(`/api/meetings/${NATIVE}`);
+    expect(init.method).toBe("DELETE");
+    expect(init.body).toBeUndefined();
+  });
+
+  it("the prompt names recordings only when the row has one", () => {
+    const confirmSpy = vi.spyOn(window, "confirm").mockReturnValue(false);
+    actionsFor({ ...row("completed"), has_recording: true }).find((a) => a.id === "erase")!.run();
+    expect(confirmSpy.mock.calls[0][0]).toContain("transcript and recordings");
+    actionsFor({ ...row("completed"), has_recording: false }).find((a) => a.id === "erase")!.run();
+    expect(confirmSpy.mock.calls[1][0]).not.toContain("recordings");
+  });
+
+  it("the prompt states that the loss is permanent and backups will not recover it", () => {
+    const confirmSpy = vi.spyOn(window, "confirm").mockReturnValue(false);
+    eraseOn("completed").run();
+    const prompt = String(confirmSpy.mock.calls[0][0]);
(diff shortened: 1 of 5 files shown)


DmitriyG228:
## A witness rig is standing by — the UI leg needs ~60 seconds of human hands

I built the fixture for the "no live UI witness" gap in the description and got it to the last step,
where it stops for a reason I can't clear myself.

**Standing on our Docker host** (`vexauiwit` project, isolated ports, other stacks untouched): the
full compose stack — postgres · redis · minio · admin-api · meeting-api · gateway · terminal — with
the **terminal image built from this branch**, so the `Delete data` action in it is this PR's code.

**Seeded**, as a real completed meeting owned by a real user, not a mock:

| | |
|---|---|
| meeting | `google_meet / therapy-session-demo`, `status=completed` |
| transcript | **2 rows** in `transcriptions` |
| recording | **2 objects** in minio under `recordings/2/…` (chunk + finalized master) |
| terminal auth | wired to that user's token, `GET /api/meetings` returns the row |

**Where it stops:** the terminal's `AuthGate` polls `/api/auth/me`, and on a fresh instance the card
becomes the one-time *"Set up your instance — first sign-in becomes the admin"* claim screen. Getting
past it means creating an account, which I don't do. So the API leg is proven and the pixels are not.

**To finish it** (the port is already tunnelled, or re-tunnel with
`ssh -fNL 13400:localhost:13400 <host>`):

1. Open `http://localhost:13400`, sign in with any email containing `test` (test mode, no password).
2. Open the `therapy-session-demo` row → **Delete data** in red.
3. **Press Cancel first.** Then confirm the transcript still reads back — the negative half is the
   one worth witnessing, and it is the half a screenshot of a working delete never shows.
4. Then accept, and check the *storage* rather than the UI:

```
docker exec vexauiwit-postgres-1 psql -U postgres -d vexa -tAq -c "SELECT count(*) FROM transcriptions WHERE meeting_id=2;"
docker exec vexauiwit-minio-1 mc ls --recursive local/vexa/recordings/2/ | wc -l
```

Both read `2` right now; both must read `0` after. A `404` in the UI alone does not distinguish
erasure from the tombstone hiding the read.

Tear down with `docker compose -p vexauiwit down -v`.

PR #1201, opened 2026-08-17
fix(teams,api): carry the separate passcode into the real Teams join URL, and refuse password aliases

### LINKED ISSUE #892: [DEV] Teams bots join with a separate passcode — carry it through the real join path and reject unsupported aliases
## Gap (adjacent to #843/#855, not covered by it)
The #855 intake validation guards `native_meeting_id` against length (>255) and NUL/control bytes → typed 422. But a value carrying **URL-reserved characters** (e.g. a Teams passcode accidentally left on the id: `397421056486982?p=X8hcQVTnGNpGelJLSv`) passes both guards (short, no control bytes) and is accepted. Then three things break downstream, none typed:

1. **Malformed join URL.** `construct_meeting_url` string-interpolates the raw id into the template (`teams: https://teams.microsoft.com/l/meetup-join/{native_meeting_id}`), yielding `…/l/meetup-join/397421056486982?p=X8hcQVTnGNpGelJLSv` — the bot fails to join (`join_failure`).
2. **Unfindable record.** `platform_specific_id` is stored as the full `397421056486982?p=…` string, so the dashboard (which looks up by the clean numeric id `397421056486982`) returns 'Meeting not found for platform teams and ID 397421056486982'.
3. The user sees a bot that 'fails to join' with no indication that the input was malformed.

## Reproduced (staging, 2026-07-21)
`POST /bots {"platform":"teams","native_meeting_id":"397421056486982?p=X8hcQVTnGNpGelJLSv"}` → meeting 13563 created, `constructed_meeting_url = https://teams.microsoft.com/l/meetup-join/397421056486982?p=X8hcQVTnGNpGelJLSv`, status → `failed/join_failure`; the same URL passed with a clean `native_meeting_id: 397421056486982` + full `meeting_url` joined fine (meeting 13564, active).

## Fix (extend #855 intake validation, at the point of introduction)
At bot-spawn intake, for platforms where `native_meeting_id` is interpolated as a URL **path segment**, reject (typed 422) — or parse/normalize — an id containing URL-reserved characters (`? & = / # space`). Ideally: if the id looks like it carries a query string (`?p=…`), either extract the passcode into the passcode field or 422 with a message naming the problem ('native_meeting_id must be the bare meeting id; pass the passcode separately or supply meeting_url'). Mirror the #855 pattern: refuse at the door, name the problem, before any record is written.

## Acceptance
- `native_meeting_id` containing `?`/`=`/`&`/space → typed 422 at intake (negative control), not a 500/silent join_failure. **Delivered** — row **A2** of the live table in [the 2026-08-17 20:42Z comment](https://github.com/Vexa-ai/vexa/issues/892#issuecomment-5329890942): `meeting_password` → **422** naming `passcode`, against the pre-fix control **on the same stack** returning **201** and silently dropping it (the reported bug, reproduced). The #910 intake guard at `bot_spawn/router.py:343` covers `? # & = /` and whitespace. ✅
- A bare id + separate passcode (or a full meeting_url) still works (positive control — meeting 13564 pattern). **Delivered** — rows **A1** and **A3**, same live table: the bot receives `…/meet/397421056486982?p=X8hcQVTnGNpGelJLSv` where the pre-fix control gave `…/l/meetup-join/555666777888` with no passcode at all; a full `meeting_url` carrying `?p=` is green both ways in. ✅

Not an acceptance leg — a scope note, moved out of the list so it is read as one: the dashboard's own
parser handles this correctly today, which is why the defect only bites raw API and MCP callers.

Found while spawning a Teams witness via the raw API — the dashboard's own parser avoids this, so it only bites API/MCP integrators.

DmitriyG228:
Closes #892.

## Why this was still broken after two "delivered"s

Both prior receipts were honest; read together they name the defect.

The #910 guard **shipped and works** — a `native_meeting_id` carrying `? # & = /` is a typed 422
([`router.py:343`](core/meetings/services/meeting-api/src/meeting_api/bot_spawn/router.py)). That
guard tells the caller *"pass the passcode in `passcode` instead"*. **The path it recommends had
never been proven past its 201.** The delivered positive control asserted `status_code == 201` and
nothing else, so it could not see the URL the bot was handed.

That URL was wrong in two ways at once. `construct_meeting_url` had one Teams template:

```python
"teams": "https://teams.microsoft.com/l/meetup-join/{native_meeting_id}",
```

`/l/meetup-join/` is the **thread-id** deep link (`19:meeting_…@thread.v2`). The id a caller holds
when they also hold a separate passcode is the **short numeric meeting id** — the one every parser
in this repo extracts from `teams.microsoft.com/meet/<id>?p=<passcode>`
([`meeting_link.py:29`](core/meetings/services/meeting-api/src/meeting_api/collector/meeting_link.py),
[`link_parser.py:111`](core/meetings/services/mcp/src/vexa_mcp/link_parser.py),
[`meetingId.ts:66`](clients/terminal/src/surfaces/meetingId.ts),
[`meeting.ts:61`](clients/extension/src/meeting.ts)). And `construct_meeting_url` took no passcode
argument at all, so there was no seam for the credential to arrive through.

Two more silent drops on the same seam, found while tracing it:

- **`teams_base_host`** is a declared api.v1 field that
  [`mcp/app.py:225`](core/meetings/services/mcp/src/vexa_mcp/app.py) actively sends, and nothing
  read it. Together with the above, **the whole MCP Teams short-link flow was broken**: the parser
  extracts id + passcode + host, hands all three to `POST /bots`, and two were dropped.
- [`msteams/join.ts:167`](core/meetings/modules/join/src/msteams/join.ts) logged the meeting URL
  verbatim — harmless only while no constructed URL had a query.

## The change

**Teams URL construction is chosen by id shape**, at the point of introduction: a 10–15 digit id
joins at `/meet/<id>`, everything else keeps `/l/meetup-join/`. **The passcode is applied to the
invocation URL only** (`apply_join_passcode`), so the bot gets a joinable address while
`data.constructed_meeting_url` — persisted, and read back on every `MeetingResponse` — stays
passcode-free. `teams_base_host` is honoured against an allowlist of Teams web-client hosts (the
bot navigates it, so it is not a passthrough), and the MCP parser now reports it for
`teams.live.com` too.

**Password aliases carrying a value are refused with a typed 422** naming `passcode`, before any
DB/runtime work — the reporter's own request: reject, don't swallow.

**Design call named, as the issue asks.** A **named alias family**, not a blanket unknown-key
guard. The api.v1 request body is open by contract (no `additionalProperties: false`;
`continue_meeting` and `teams_base_host` both ride on it), so refusing every undeclared key would
be a contract break that 422s working integrations — including the integrator this fix is for. The
class closed is the one where **the dropped field is a credential**.

## Observation bundle

Live stack on `bbb` built from this branch (compose project `vexa-892`), calls through the gateway.
Every claim below is either an **observed output change** or is labelled as a weaker rung.

| # | Observation | Evidence | Rung |
|---|---|---|---|
| A1 | bare id + separate `passcode` → the bot navigates the passcode-bearing URL | live `POST /bots` → 201; spawned container `BOT_CONFIG.meetingUrl` = `https://teams.microsoft.com/meet/397421056486982?p=X8hcQVTnGNpGelJLSv` | **output changed** |
| A1 | **negative control** — same call on the pre-fix build, same stack | 201, `BOT_CONFIG.meetingUrl` = `https://teams.microsoft.com/l/meetup-join/555666777888`, **no passcode** | **output changed** |
| A1 | `teams_base_host` selects the cloud | gov: readback `https://gov.teams.microsoft.us/meet/913345678901`, bot gets `…?p=GovCode123`; `teams.microsoft.com.evil.example.com` → 422 **before** the quota check | **output changed** |
| A2 | `meeting_password` / `password` → typed 422, no row written | live: both 422 naming `passcode`. **Negative control**: pre-fix build returned **201** and dropped the code — the reported bug, reproduced live | **output changed** |
| A3 | full `meeting_url` with `?p=` stays green | URL-only derivation and URL + explicit-id companion: 201, bot gets the URL verbatim; a URL's own `?p=` wins over a separate `passcode` | **output changed** |
| A4 | passcode off readback and out of logs | live readback `https://teams.microsoft.com/meet/397421056486982` (no passcode) while the bot got it; `join-passcode.test.ts` red control printed the passcode in the log pre-fix | **output changed** |
| A- | thread id, Google Meet, Zoom, Jitsi unchanged | live: thread id keeps `/l/meetup-join/`, Meet keeps its URL; zoom/jitsi URLs not rewritten | **output changed** |
| A- | suites | meeting-api 923 passed · MCP 73 passed · `@vexa/join` all green · `node scripts/gates.mjs all` **34/34 green** | green |

**Red control on the pre-fix tree** (source reverted, tests kept): 4 of the new tests fail, the 201
body showing `constructed_meeting_url: https://teams.microsoft.com/l/meetup-join/397421056486982`.
The A3/A- regression rows pass both before and after, which is what they are for.

## What is NOT proven — the validator's row

**No bot has joined a real Teams meeting on this branch.** I have no Teams meeting with a real ID +
passcode to join, so the one claim resting on Microsoft's behavior — that Teams admits an anonymous
bot at `/meet/<id>?p=<passcode>` — is **not** witnessed. What supports it is internal: four
independent parsers in this repo encode that shape as the real Teams short link, and the MCP prompt
tells users to paste exactly it.

I probed the two URL shapes in a browser; **the probe was inconclusive** — that browser carries a
signed-in Microsoft session, so both shapes landed on "Classic Teams is no longer available"
instead of an anonymous pre-join. It reproduces neither the bot's conditions nor a verdict, so I am
reporting it as no evidence rather than reading it either way.

That single row is the human bar: **one solo Teams meeting, joined by ID + passcode.**

Also honest about a boundary: A4 covers the **constructed** URL. A caller who supplies
`meeting_url` with a passcode in its query still gets that URL stored verbatim — unchanged
behavior, and the dashboard's send-bot round-trip depends on it.

## Housekeeping

The `vexa-892` compose stack is **still running on `bbb`** — I did not tear it down, per the
standing rule not to stop containers on that host. Reap with:

```bash
ssh bbb 'cd ~/vexa-892/deploy/compose && docker compose -p vexa-892 down -v'
```

Docs move with the change: [send-a-bot](docs/docs/how-to/send-a-bot.mdx),
[api/meetings](docs/docs/api/meetings.mdx), and a changelog fragment at
`docs/changelog.d/892-teams-separate-passcode.md`.

## Contribution rights


DmitriyG228:
**Conflict resolution — merged `main` (`4647c96e`) in; `DIRTY` → `MERGEABLE`.** Merge commit, no force-push. Two conflicts, both trivial collisions rather than disagreements.

| File | Resolution |
|---|---|
| `bot_spawn/service.py` — `__all__` | Union. This PR exports `apply_join_passcode` and `resolve_teams_base_host`; `main` added `MeetingStopped`, `DEFAULT_LOBBY_BUDGET_S` and `lobby_budget_ms`. All seven kept. |
| `docs/docs/api/meetings.mdx` | Both sides appended a paragraph at the same anchor. `main`'s Jitsi footnote now sits directly under the platform table it glosses (it explains the `*` on the Jitsi row), then this PR's Teams addressing paragraph, then the existing Jitsi paragraph. |

No claim in either file contradicted the other — this was two additions landing on one line, not a fact in dispute.

**Validation:** `meeting-api` suite **1172 passed, 3 skipped**; `mcp` link-parser suite **42 passed**. All 14 pre-push static gates green. The TypeScript leg (`msteams/join-passcode.test.ts`) was **not** run here — it needs the join module's toolchain; CI owns it.

Issue #1219, opened 2026-08-18
transcript: the endpoint returns the whole transcript every poll, and silently ignores since/after/limit/offset


DmitriyG228:
## Value this issue delivers

An agent following a live meeting fetches **only what it has not already seen**, and any
unsupported query parameter is **refused with a typed 4xx** instead of being silently dropped.

## Evidence — measured on a live Google Meet, 2026-08-18

`GET /transcripts/{platform}/{native_meeting_id}` returns the **entire** transcript on every
request. Two consecutive polls, same live meeting:

```
poll 1: 10,423 bytes, 19 segments
poll 2: 11,118 bytes, 21 segments      → ~348 bytes/segment
```

There is no incremental fetch — and the parameters that would provide one are **accepted and
ignored**:

| Query | Response |
|---|---|
| `?since=2026-08-18T08:20:00Z` | **200, full payload** |
| `?after=…` | **200, full payload** |
| `?start_time=…` | **200, full payload** |
| `?limit=2` | **200, full payload (11,095 bytes)** |
| `?offset=2` | **200, full payload** |

No `400`, no typed refusal. **The caller is told the request succeeded and handed everything
anyway** — the same silent-drop class as the passcode in #892, which a customer reported on
2026-07-30 and phrased better than we can: *"a validation error instead of silent drop would save
a lot of debugging."*

### Cost

At the measured **6 segments/minute** and a 5-second poll:

| Meeting | Final payload | Polls | Downloaded |
|---|---|---|---|
| 15 min | 31 KB | 180 | **2.7 MB** |
| 60 min | 122 KB | 720 | **42.9 MB** |
| 120 min | 244 KB | 1,440 | **171.8 MB** |

**Growth is quadratic in meeting length** — the payload grows linearly while the poll count grows
linearly too. Following one two-hour call costs ~172 MB to learn perhaps 250 KB of new text.

## Why it matters beyond bandwidth

This is the difference between "an agent can watch a meeting" and "an agent can watch meetings".
Latency is already fine — **measured 2–12s** between a sentence ending and the segment being
readable, median under 10s — so polling is the right transport. It is the payload, not the
transport, that does not scale.

## Deliverable

1. Honour `since` (timestamp) **or** `after_segment_id` (cursor) on the transcript endpoint.
2. **Reject unsupported query parameters with a typed 4xx.** A parameter that changes nothing must
   not return 200.

## Acceptance

[...]

PR #1220, opened 2026-08-18
feat(transcript): incremental `?since=` reads, and a typed refusal for parameters we don't honour

DmitriyG228:
**Delivers issue:** #1219

## Contribution rights

Every commit carries `Signed-off-by: DmitriyG228 <2280905@gmail.com>` (the repo-local identity,
matching this repo's existing history).

## What this delivers

`GET /transcripts/{platform}/{native_meeting_id}` returned the **entire** transcript on every poll
and accepted-then-dropped every parameter that would have narrowed it. Both are fixed:

1. **`?since=` — an incremental read.** Returns only segments **created or changed** at/after the
   cursor. Every response carries **`next_since`** (the watermark for the next poll) and, on a
   cursor read, **`retracted_segment_ids`** (what to drop) and **`resynced`**.
2. **Unsupported query parameters return `400` naming them**, listing what the endpoint accepts.

### Why a change watermark and not `after_segment_id`

A segment is **not append-only**. A live draft is rewritten in place as its confirmation, a speaker
can be repainted retroactively across segments already served, and a superseded draft can be
retracted. An append-pointer cursor would hand a follower the draft and never the confirmation —
silently stale, which is worse than no cursor. (`segment_id` is also unsortable: three
producer-local mint formats, none of them ordered.)

So the filter is a per-segment change-stamp, which is **monotone across both homes a segment lives
in**: redis `updated_at` (stamped by ingest on every write) then postgres `created_at` (which the
db-writer's upsert re-stamps on every re-flush). A segment only moves live→durable after 30s of
immutability, so the stamp never moves backwards and the watermark can never step over a segment.

Deletion is the one change a change-cursor cannot express as data, so retraction now writes a
tombstone and a cursor read reports it. Past the tombstone TTL the cursor is refused in favour of a
full re-read (`resynced: true`) — a cursor we can no longer honour completely degrades to a full
read, never to a quietly wrong one.

`limit`/`offset` are **deliberately refused rather than implemented**: `since` is the bounding
mechanism, and a page-limited transcript would truncate a follower's feed while `next_since` claimed
it was complete.

## Observation bundle

⏳ **The live run is in flight — this PR is opened early so the work is findable.** The bundle below
is filled in as each leg lands; see the comments for the final numbers.

- **C1 — parameter refusal, live through the gateway.** ✅ ran: every parameter the issue measured as
  "200 + full payload", against a stack built from this branch on `bbb`. saw: `?limit=2`, `?after=`,
  `?offset=`, `?start_time=` → **400** naming the parameter; `?since=garbage` → 400; a valid
  `?since=` → 200. concluded: deliverable 2 is done at the observable-output rung.
- **C2 — incremental read + union control on a live meeting.** ⏳ in flight.

## Acceptance floor

| Row | Evidence |
|---|---|
| A1 — following a 2h meeting costs ~the transcript's own size | ⏳ in flight |
| A2 — `?limit=2` limits or errors | ✅ **400**, `{"detail":"Unsupported query parameter(s): limit. This endpoint accepts: since."}` |
| A3 — same meeting with/without cursor → same union of segments | ⏳ in flight |

## Docs diff (D6c)

- [`docs/docs/api/meetings.mdx`](docs/docs/api/meetings.mdx) — new **"Following a live meeting — `since`"**
  section under *Get the transcript*: the cursor, `next_since`, `updated_at`, `retracted_segment_ids`,
  `resynced`, the merge-by-`segment_id` rule, and why `limit`/`offset` are refused.
- [`docs/docs/api/errors.mdx`](docs/docs/api/errors.mdx) — the `400` on `/transcripts/*` under the
  Meetings API surface.
- [`docs/changelog.d/1219-transcript-since-cursor.md`](docs/changelog.d/1219-transcript-since-cursor.md) — fragment (the `docs-current` touch).

## Security checks

- `node scripts/gates.mjs all` — ✅ green (includes `gate:licenses` 460 deps Cat-A clean,
  `gate:image-licenses`, `gate:access` default-deny, `gate:config-contract`).
- No new dependency. No new secret. The cursor is a **read filter only** — it is applied after the
  existing ownership check, never as an authorization path (pinned by
  `test_cursor_does_not_bypass_ownership`).

## Validation request

Preferred validator: **@DmitriyG228** (the reporter). What to watch: send a bot to a real meeting,
poll `?since=` in a loop alongside a plain poll, and confirm the two end up with the same transcript
while the cursor's traffic stays flat. Deployment validated here: **full compose**, built from this
branch (`vexa/meeting-api:1219`, all other services the published `:v012` images).


DmitriyG228:
**Conflict resolution — merged `main` (`4647c96e`) in; `DIRTY` → `MERGEABLE`.** Merge commit, no force-push. Two files conflicted, both in `collector/`.

`main` landed three things under this branch: viewer-aware response projection ([#1243](https://github.com/Vexa-ai/vexa/pull/1243)/[#1244](https://github.com/Vexa-ai/vexa/pull/1244)), artifact-deletion gating ([#1293](https://github.com/Vexa-ai/vexa/pull/1293)), and `speaker_key` on segments. All three touch the same read paths this PR rewrote for `?since=`.

**The load-bearing decision: `viewer_is_owner` is threaded through the new cursor path rather than re-derived.**

`main` made `_merge_live_segments(..., *, viewer_is_owner: bool)` a **required keyword** on purpose — both callers already know the answer (the native-keyed read constrains `Meeting.user_id == user_id` in SQL; the by-id read evaluates an explicit owner branch), and re-deriving it inside the helper would be a second place for the decision to live. This PR inserts `_cursor_response` between those callers and that helper, so the requirement had to survive one more hop:

- `_cursor_response(..., *, viewer_is_owner: bool)` — also required, also passed straight through. It shapes cursor fields only; it never re-decides who sees which tier of the meeting blob.
- Native-keyed read → `viewer_is_owner=True` (owner by construction of the SELECT).
- By-id read → `viewer_is_owner=is_owner`, the same variable that authorized the read.

Same treatment in `fakes.py`, so the fake and the real store still agree on what a share recipient receives.

Other hunks: `_segment_to_api` takes the union of `main`'s `speaker_key` and this PR's `updated_at`; `main`'s artifact-deletion gate is kept ahead of this PR's `_transcript_pg_part(db, meeting, effective_since)`.

**Validation:** `meeting-api` suite **1182 passed, 3 skipped** — this PR's `test_transcript_cursor.py` / `test_transcript_merge.py` and `main`'s viewer-projection and deletion tests together. All 14 pre-push static gates green. No container workloads were run on this machine; CI owns the full suite.

DmitriyG228:
**Re-merged `main` (`42651e10`); `DIRTY` → `MERGEABLE`.** This branch had gone dirty again after [#1242](https://github.com/Vexa-ai/vexa/pull/1242) merged. Merge commit, no force-push.

One conflict, `docs/docs/api/meetings.mdx`: #1242 added a paragraph on `GET /transcripts/by-id/{meeting_id}` at the same anchor where this PR adds its `### Following a live meeting — since` section. Genuinely independent additions, so both are kept — main's paragraph first (it continues the preceding read description), then this PR's section.

The code resolution from the previous pass — `viewer_is_owner` threaded through `_cursor_response` — was untouched by this merge and re-verified green.

**⚠️ Heads-up for whoever merges second: this PR and [#1164](https://github.com/Vexa-ai/vexa/pull/1164) both edit the same region of `docs/docs/api/meetings.mdx`.** #1164 adds a *Two keys* section that documents the by-id read at length and treats `/transcripts/by-id/{id}` as an older spelling; this PR adds the `?since=` section just after it. Neither conflicts with `main` now, but they will conflict with **each other**. Nothing to do yet — just expect one textual conflict in that file, and resolve it as a union of the two sections rather than a choice between them.

**Validation:** `meeting-api` **1203 passed, 3 skipped**. All 14 pre-push static gates green. No container workloads run.

Issue #1222, opened 2026-08-18
[DEV] The meeting happening now leads the meetings list — order by activity, pin live rows


DmitriyG228:
### Value this issue delivers

The meeting happening right now is the first thing the user sees in the meetings list — never buried under yesterday's rows.

### Problem (founder witness, production, 2026-08-18)

A live calendar auto-joined meeting (row 26298, bot admitted, transcribing) was invisible in the dashboard meetings list while it was happening. Verified server-side: `GET /meetings` for the owning user returns the row — the defect is ordering, not filtering. The list sorts by `created_at DESC`, and a calendar-managed row is created at *import* time (when the calendar syncs, possibly a day or more before the meeting), so the row sinks below every meeting row created since the import. The founder concluded the meeting was missing entirely; the bot in the room plus an empty-looking list reads as a product failure.

Aggravating: the list does not live-refresh (vexa-platform#298), so rows created since page-load are absent too.

### Solution sketch

Order the list by activity time — e.g. `GREATEST(scheduled_at, start_time, created_at)` — and/or pin non-terminal (scheduled/joining/awaiting_admission/active/stopping) rows above terminal ones. `scheduled_at` is already projected in list rows (`project_list_data`), so this is an ORDER BY change in `list_meetings` plus the sort in the dashboard store.

### Validation contract

Human act: while a calendar-scheduled meeting is live, open the meetings list — it appears at the top. Falsifier: any live meeting below a terminal row. Machine oracle: list test with an imported-yesterday/live-now row asserting position 0.

DmitriyG228:
Fix staged in two PRs, one per branch that carries the code — #1226 (server: list_meetings ordering + expression indexes + MIGRATION-0005, based on codex/1150-calendar-multi) and #1227 (dashboard client sort, based on codex/prod-dashboard-source-accounting where the retained source lives). Both frozen out of rc.17: next rc or next train, founder's call. Rung: PRs open, not merged.

DmitriyG228:
Delivered server-side.

[...]

Issue #800, opened 2026-07-18
meeting-api: shared-meetings list query (OR + JSONB @>) defeats indexing — 100s+ scans saturated prod DB pool, forced 0.12.12 rollback


DmitriyG228:
## Where we are (honest)

The meetings list query built at `core/meetings/services/meeting-api/src/meeting_api/collector/adapters.py:412-433` (load-bearing arm at line 424, introduced in `1e9608d8`, present v0.12.0→v0.12.13):

```sql
SELECT meetings ... WHERE user_id = $1
   OR (data->'transcript_viewers') @> to_jsonb($1)
  [OR data->>'workspace_id' IN (...)]
ORDER BY created_at DESC LIMIT n
```

During the 2026-07-19 v0.12.12 production cutover (vexa-platform), EXPLAIN showed `Index Scan Backward using ix_meetings_created_at` with the OR as a **Filter**: for users with few/old meetings the planner walks the created_at index backwards over the whole table. Cheap plan estimate, catastrophic runtime — dozens of concurrent 100s+ executions saturated the DB pool in the post-cutover retry storm, took down bot dispatch + auth, and **forced the rollback of prod to 0.10**. An existing GIN index on `data` does not help: the OR + backward-scan plan defeats it.

This is the gating blocker for re-deploying 0.12.x to production.

## Prepared solution + forks

Split the OR so each branch is independently index-scannable:

- **Preferred:** rewrite `list_meetings` as a UNION of two (three, with workspace) branches — `user_id = $1` (btree `(user_id, created_at DESC)` — `idx_meetings_user_created` may already exist) and `data->'transcript_viewers' @> to_jsonb($1)` (jsonb_path_ops GIN on the expression) — each with its own `ORDER BY created_at DESC LIMIT n`, merged and re-limited.
- **Fork:** expression index `((data->'transcript_viewers'))` + predicate rewrite the planner can BitmapOr.

## Acceptance

- [x] EXPLAIN ANALYZE on a prod-scale dataset: worst-case user (0 recent meetings, non-empty transcript_viewers membership) < 50ms; no `Index Scan Backward ... Filter: (OR ...)` plan shape.
- [x] Negative control: query results identical to current shape (same rows, same order, has_more semantics preserved incl. the `effective_limit + 1` over-fetch).
- [x] Timed load run (concurrent list calls across many users) shows no pool saturation.

Confidence note from the incident session: this analysis was ~70% understood — **measure before trusting any fix**.

[...]

PR #1226, opened 2026-08-18
meetings: list orders by event time, non-terminal rows pinned (#1222)

DmitriyG228:
Fixes the ordering half of #1222. **Do not merge — the 0.12.23 train candidate is frozen at rc.17; lands in the next rc or next train, founder's call.**

**Witnessed evidence (prod, 2026-08-18):** row 26298, a live calendar auto-joined meeting (created at calendar-import time Aug 16 22:40, scheduled_at Aug 18 09:00), sat at list position 19 — under 18 rows created since the import — while the founder was in the meeting.

## Change

`list_meetings` (list_view path only — `GET /meetings`, `GET /bots`) now orders by:

1. non-terminal pin: `status IN (scheduled, requested, joining, awaiting_admission, active, stopping)` DESC
2. event time: `COALESCE(data.scheduled_at, start_time, created_at)` DESC
3. `id` DESC tiebreak

Internal enumeration (get-by-id filter, `/bots/status`, calendar sync, the native-id resolver's documented "newest owned row") keeps `created_at DESC`.

**#800 union-branch design preserved.** The branches must pre-limit by the same key as the outer sort (a created_at top-N would drop exactly the rows this fix surfaces), so two expression indexes carry the new key: `ix_meeting_user_event_order`, `ix_meeting_workspace_event_order`, over an IMMUTABLE `meeting_event_time()` (plpgsql, exception-guarded so a malformed `scheduled_at` can never fail a row write). `ensure_schema` creates the function before any index DDL; prod rollout is out-of-band per [MIGRATION-0005](https://github.com/Vexa-ai/vexa/blob/1222-list-order-event-time/core/identity/services/admin-api/src/admin_api/schema/MIGRATION-0005-meeting-event-order.md) — note the deploy-order hazard there (the query calls the function directly).

**EXPLAIN, 200k-row dataset, heavy user 20k rows:** each branch plans as `Index Scan Backward` on its event-order index, LIMIT 51, no sort of the caller's history (0.18 ms owner branch; 2.4 ms full union incl. GIN viewer branch). `status IN` normalizes to the same `= ANY` as the index expression, confirmed in the plan.

`needs_human_help` is deliberately not pinned (the issue's ruling lists six statuses); flag if it should be.

## Tests

- `test_list_event_order.py` — imported-days-ago/live-now row at position 0 (the 26298 shape), terminal-vs-scheduled pinning, event-time coalesce + offset timestamps + malformed fallback, internal path unchanged.
- PG-stack eval (`test_stack_postgres.py`) — function + both indexes converge via `ensure_schema` on real Postgres, malformed `scheduled_at` inserts cleanly.
- Full suites green: meeting-api 1140 passed, admin-api 30 passed + 13 PG-stack on Docker host.
- One stale assertion updated (`test_get_meetings_conforms` assumed "newest first").

Dashboard client sort matched in #1227 (source lives on `codex/prod-dashboard-source-accounting`).

Rung: PR open, not merged.

## Contribution rights


DmitriyG228:
`next` removed by the release lane: this PR targets `codex/1150-calendar-multi`, which merged to main this morning — merging here delivers nothing to any train. Re-badge after retargeting to `main` and going green there.

PR #1227, opened 2026-08-18
dashboard: meetings list sorts by event time, non-terminal pinned (#1222)

DmitriyG228:
Client half of #1222 — companion to #1226 (server ordering). **Do not merge — next rc or next train, founder's call.** Based on `codex/prod-dashboard-source-accounting` because the retained dashboard source lives only there (34565b48).

The store re-sorted every page by `created_at`, which would re-bury the live calendar meeting client-side even after the server orders correctly (witnessed prod 2026-08-18: row 26298 at position 19 while the founder was in it). One comparator in `lib/meeting-list.ts` now matches the server key exactly — non-terminal pin (scheduled/requested/joining/awaiting_admission/active/stopping), then `data.scheduled_at ?? start_time ?? created_at` DESC, then id DESC — used by `prepareMeetingPage` and both store sorts (`fetchMeetings`, `fetchMoreMeetings`).

Tests: comparator suite in `meeting-list.test.ts` incl. the 26298 shape at position 0; vitest 67/67 green. `tsc --noEmit` shows only the pre-existing `release-version.generated.json` build-artifact error.

Rung: PR open, not merged.

## Contribution rights

### DIFF
--- a/services/dashboard/src/lib/meeting-list.test.ts
+++ b/services/dashboard/src/lib/meeting-list.test.ts
@@ -1,0 +1,74 @@
+// #1222 — the list orders by MEETING EVENT time, non-terminal rows pinned first, matching the
+// server (meeting-api list_meetings). The witnessed failure: a calendar row imported Aug 16,
+// live Aug 18, sat at position 19 under 18 rows created since the import.
+import { describe, expect, it } from "vitest";
+
+import { compareMeetingsListOrder } from "./meeting-list";
+import type { Meeting } from "@/types/vexa";
+
+function meeting(partial: Partial<Meeting> & { id: string }): Meeting {
+  return {
+    platform: "google_meet",
+    platform_specific_id: `native-${partial.id}`,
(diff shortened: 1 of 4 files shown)


DmitriyG228:
#1222 was delivered server-side, not here — flagging rather than closing, because this branch's own change is not in main.

**In main:** #1226 (`d1f4b235`, "meetings: list orders by event time, non-terminal rows pinned"), released in **v0.12.24** and installed in production. `list_view` orders by `(status-pin, COALESCE(data.scheduled_at, start_time, created_at), id) DESC`, with `meeting_event_time()` and the two expression indexes (MIGRATION-0005). Clients get the ordering from the API.

**Not in main:** this PR's client-side sort (`services/dashboard/src/lib/meeting-list.ts`). `services/dashboard/` does not exist in OSS main at all — it is stacked behind #1183.

Decide whether the client-side sort is still wanted now that the API returns the order.

Issue #1221, opened 2026-08-18
[DEV] Human speech is never credited to another notetaker's tile — fix attribution under multi-bot participants


DmitriyG228:
### Value this issue delivers

Words spoken by a human are never attributed to another notetaker bot's tile.

### Evidence (founder witness, production, 2026-08-18)

Live meeting `google_meet/wzu-yoeg-szi` (prod meeting row 26298, calendar auto-join, bot "Vexa"): the founder's own speech — a counting probe "1, 2, 3, 4, 5, 7, 8, 8, 8, 9, 10. Oops, it's wrong that all the speakers are…" — rendered at 10:30:11 UTC attributed to **"Ekwis Notetaker (Unverified)"**, a competing notetaker bot present as a participant. The "(Unverified)" corroboration marker behaved as designed (the name was not verified), but the underlying attribution chose the wrong participant tile entirely: a bot with no human voice credited with human speech.

Meeting had ≥3 non-human participants (Vexa bot, Ekwis Notetaker, one more waiting) plus 2 humans — the multi-bot-participant condition is likely load-bearing.

### Why now

Prod runs the 0.12.23 cut; a transcript that hands a customer's words to a competitor's bot is a trust-destroying artifact in exactly the meetings where multiple notetakers show up (increasingly common). The counting-probe fixture from this meeting (row 26298, session window ~10:29–10:31Z) should be captured as a replay tape before it ages out.

### Validation contract

Human act: witness a meeting containing a second notetaker bot; speak; every segment carries the human's tile, never the bot's. Falsifier: any segment attributed to a participant whose tile emitted no audio. Logs lens: per-speaker capture stream mapping for the session vs the rendered speaker labels.

DmitriyG228:
Second instance, same meeting (~10:3x UTC): a long passage of the OTHER human participant's speech (their use-case walkthrough) also carried a wrong speaker label. Two distinct human speakers affected in one meeting with a second notetaker bot present — this is not a single-tile edge. Fixture window for row 26298 should cover ~10:29–10:40Z.

DmitriyG228:
[...]

Issue #1224, opened 2026-08-18
[DEV] Attribution errors become a measured fleet metric — high-precision self-incrimination fixtures from production transcripts


DmitriyG228:
### Value this issue delivers

A production-derived, verified fixture set of speaker-misattribution cases — built only from signals with near-perfect precision — that the attribution train uses as its acceptance metric and regression gate.

### Method

Transcripts self-incriminate. Two signal classes are precise enough to label errors without a human in the room:

1. **Vocative** — a segment addressing someone by name ("thanks, Matt", "Matt, let me interrupt") is not spoken by that person; a track labeled with the vocative's target is mislabeled.
2. **Self-identification** — "I'm X" / "this is X" binds the speaking track to X; a conflicting label is an error.

An LLM judge reads stored transcripts + roster and emits typed per-segment evidence for these two classes only. Calibration set: the two witnessed meetings of 2026-08-18 (#1221), where every error is known — judge precision/recall is measured before any sweep. Meetings with full signal tapes get mechanical verification (csrc separation, roster timing, settled-window replay); a fixture is **gold** when the linguistic signal and the tape agree.

### Scope guards

- Signals: vocative + self-identification ONLY. No turn-logic, no register heuristics.
- Corpus: restricted to consented/internal meetings (boundary held by the private register; this issue carries no customer content).
- Names pseudonymized consistently before the judge; judge output carries segment ids and signal types, not content.
- Known capture gap: the GMeet lane records no csrc/observations, so GMeet fixtures cap at silver (linguistic-only) until that lane records cross-check signals — tracked as a scope item here.

### Deliverables

1. Judge + scorer (runnable, in-repo) with calibration report against the two witnessed meetings.
2. Fixture manifest schema + the initial gold set (starting with the 26424 tape already verified in #1221).
3. Fleet metric definition: attribution-error rate per platform/lane/release, computed from the fixture set.

### Validation contract

None — acceptance is machine-observable. Oracle: on the calibration meetings the judge attains measured precision 1.0 on flagged segments (its flags match the humanly-known errors); the manifest replays end-to-end through the scorer.

Refs #1221, #1150.

DmitriyG228:
Fixture set built and calibrated. PR: https://github.com/Vexa-ai/vexa/pull/1225 (branch `eval/misattribution-fixtures` off `codex/1150-calendar-multi` @ 0712d1da; **not** for the 0.12.23 candidate).

[...]

PR #1228, opened 2026-08-18
fix(namer): an evidence name stays revisable until the roster settles

DmitriyG228:
## Contribution rights

---

## Mechanism

`TrackNamer` accepted a name from evidence measured against a roster that had not finished discovering the room, and the acceptance was permanent.

On production meeting **26424** (Teams, founder witnessed it live), track `csrc:201` was accepted as "Matt" at **+35 s on 3.8 s of evidence**, while the roster held only "Matt" — the second participant's first roster sighting was ten seconds later. Acceptance is never revisited, so **93.2 % contrary evidence over the following 28 minutes** could not correct it; the exclusivity rule then correctly refused "Matt" to the track that really was Matt, which fell back to "Speaker A". Half the meeting printed under the wrong human's name, including a line where one man thanks the other by name.

`TRACK_NAME_ROSTER_SETTLE_MS` gated the **elimination** path only — that is the gap named in #1221. But applying the quiet test to acceptance would not have caught this either: at +35 s the roster had been **quiet for 35 seconds** and coverage called itself complete at 1-of-1. The only contradiction available at that instant is that **the transport was already carrying two humans while the roster listed one**.

## Fix shape

`rosterSettled` = quiet for `rosterSettleMs` **and** not provably behind (`roster.size >= tracks heard`). A meeting whose roster never reports is settled by definition, so no name is ever suppressed waiting for a signal that will not arrive.

An evidence name earned while the roster is **not** settled is **published at once** and stays **revisable**: it is re-derived on every evaluation and may change, or be withdrawn — releasing the name so the track that owns it can earn it. Past settle it is permanent exactly as before, so contrary evidence cannot move a settled name.

Publishing rather than withholding is deliberate: a withheld name costs latency in every meeting, and a roster that stays permanently behind (a CSRC rotation gives 3 tracks for 2 people) would silently suppress every name. A published name that is later corrected repaints through the same retroactive path a rename already uses.

Elimination gains the same not-provably-behind test — it fired on this tape at +40 s and re-made the identical wrong binding — and is otherwise untouched.

## Tests

Committed regression test is **synthetic** (`track-namer.smoke.test.ts`), reproducing the shape: late second roster sighting + early bleed evidence. Red before the change, green after.

| check | pre-fix | post-fix |
|---|---|---|
| late-appearing participant gets his own name | ❌ | ✅ |
| the locked-away name goes to the man who is that name | ❌ | ✅ |
| the wrong name is released, not quietly overwritten | ❌ | ✅ |
| premature name published at once (no added latency) | ✅ | ✅ |
| single-participant meeting named when its evidence lands | ✅ | ✅ |
| name earned inside the unsettled window published immediately | ✅ | ✅ |
| past the settle boundary the name is permanent again | ✅ | ✅ |

`pnpm test` in `core/meetings/modules/mixed-pipeline`: **24 suites, 208 assertions, 0 failures**.

## Validation: the 26424 tape

Offline namer-only replay of the prod tape (`signal/23/26424/812de312-…`, read-only), driven with the pipeline's exact construction — 330 CSRC edges, 1253 hints, 6246 audio frames, 6 roster sightings.

| track | truth | rc.17 (pre-fix) | this branch |
|---|---|---|---|
| `csrc:201` | Dmitry Grankin | **Matt** ❌ (3.8 s, accepted +35 s) | **Dmitry Grankin** ✅ [evidence] |
| `csrc:840` | Matt | **Speaker A** ❌ (name was taken) | **Matt** ✅ [evidence] |

Post-fix label stream: `201→Matt (+35 s, provisional)`, `201→Speaker B (+40 s, released)`, `201→Dmitry Grankin` and `840→Matt (+241 s)`. Final evidence: 201 = Dmitry 55.3 s / Matt 3.8 s; 840 = Matt 68.2 s.

Ground truth is the triple already established in #1221 (hint×CSRC purity, the settled-window control replay, and the vocatives — 840 says "so nice to meet Dmitry", 201 says "thanks, Matt").

Namer-only; no capture or pipeline changes.

Refs #1221


DmitriyG228:
## Corpus validation — 25 tapes, not one meeting

This branch was proved on 26424, the tape it was written against. That tape agrees with it by
construction. `882049c0` adds the harness and scorecard that answer the other question: **what
does it do to the twenty-four tapes it was not written for?**

Every Teams signal tape reachable to this workspace — **14 production tier-A tapes** (`signal/23/…`,
read-only, streamed and reduced inside the cluster so no audio leaves it) and **11 dogfooding
harvests**, spanning `v0.12.22-rc.1` → `v0.12.23-rc.10.packet2`, 1 to 6 tracks each, 40 s to 66 min.
Both arms are the **same file at two commits** (`860d9996` and its parent `0712d1da`), driven by one
replay driver over one set of inputs.

| | correct | wrong | unnamed |
|---|---|---|---|
| pre-fix (`0712d1da`) | 27 | **1** | 2 |
| this branch | **29** | **0** | 1 |

**48 tracks, 30 with established ground truth, 18 excluded — each with its reason, none dropped
silently. Zero regressions.**

### Every delta, explained

Exactly two tracks change across the whole corpus, both on 26424, both repairs:

| tape | track | truth | pre-fix | this branch | control |
|---|---|---|---|---|---|
| 26424 | `csrc:201` | P1 | **P2** (wrong) | **P1** ✅ | P1 |
| 26424 | `csrc:840` | P2 | **Speaker A** (unnamed) | **P2** ✅ | P2 |

`control` is the settled-window replay — the roster's discovery window cut away entirely, so the
premature-acceptance hazard has no window to fire in. It agrees with the fix on both. 26424's truth
also carries the third line: the label-blind vocative/self-ID judge of #1224, flagged-precision 1.0
per #1225.

**Twenty-two tapes are byte-identical between the arms.** The fix touches nothing it was not aimed at.

### Latency: no cost

| | |
|---|---|
| tracks both arms named | 38 |
| identical time-to-first-name | **38** |
| distinct deltas (fix − baseline) | **+0 s** |

Publishing provisionally rather than withholding was the design choice; the corpus is where it stops
being an argument. Naming is not one tick slower anywhere.

### Label churn: the cost it does have

| | pre-fix | this branch |
|---|---|---|
| retraction events, whole corpus | 0 | **5** |
| tracks whose label stream changed | — | 2 |

One is 26424's intended repaint. **The other four are one track on 26132**, and that one is worth a
look before merge:

```
P1@37.6s → Speaker A@38.7s → P1@1804.1s → Speaker A@1822.7s → P1@2029.7s → … → P1@2306.6s
```

Its roster **can never settle**. The second participant's Teams display name is a lowercase topic
handle, which `requireCanonicalDisplayName` correctly refuses — so `roster.size` stays 1 while the
transport carries 2 tracks, and `rosterSettled`'s not-provably-behind test is false for the entire
57-minute meeting. The name therefore stays provisional to the end and is re-derived on every tick,
while its owner-share sits at **0.71 against the 0.70 threshold** (465.6 s of own evidence against
188.5 s of bleed) — so the verdict crosses the bar back and forth and the label repaints four times.

Final name is correct in both arms and time-to-first-name is identical, so this is **not a
regression** and not a stop condition. It is the predictable shape of "revisable until settled" when
settled never arrives, and the PR body already anticipates the permanently-behind roster. Worth
deciding explicitly whether a provisional name should stop being re-derived once its evidence has
been stable for some interval, or whether four repaints across an hour is an acceptable price.

### What is committed

`core/meetings/eval/namer-corpus/` — the reducer, the truth builder, the replay driver, the scorer,
the per-tape ground truth, and [`SCORECARD.md`](../blob/fix/namer-roster-settle/core/meetings/eval/namer-corpus/SCORECARD.md).

The namer takes csrc edges, DOM hints, roster sightings and ticks — no audio, no transcript — so a
tape reduces to metadata and 25 of them replay in seconds with no STT. Ground truth comes from
**hint × csrc exclusive correlation over the whole meeting with a bijection constraint**: a hint
landing while exactly one track is active attributes that instant, overlaps are discarded rather than
modelled. That is not the namer's own computation restated — the namer decides online from a prefix,
under a settle delay and first-past-the-post exclusivity; this is a whole-meeting aggregate
computable only after the meeting ends, which is what lets it serve as an oracle.

Committed artefacts carry per-tape pseudonyms (`P1`, `P2`, …) only — including inside exclusion
reason strings. No tapes, no namer inputs, no transcripts, no real-name maps.

PR #1247, opened 2026-08-19
fix(teams): restore contested-word marking in a lighter bracket form

DmitriyG228:
The `⟦words⟧{CSRC A↔CSRC B}` marker was reverted to detection-only in #1162, which left the changelog, the debug page's contested counter and the exported marker all asserting behaviour the pipeline no longer had.

Marking is restored, and the notation is now plain ASCII brackets around the contested run: `[words]` on **both** rows, symmetric, no winner inferred. The rival CSRC identity no longer travels inside transcript text — it stays in `health().contestedPairs` telemetry. Docs, the terminal renderer and the eval-ui harness copies follow the new form.

## Base

This code lives on `codex/1150-calendar-multi` (open PR #1151). **None of it is on `main`**, so this PR is based on that branch.

### 1. A pinned behaviour test was deliberately replaced

`core/meetings/modules/mixed-pipeline/src/teams-csrc-gmeet-pipeline.test.ts` carried

> `every live publication callback keeps contest diagnostics out of transcript text`

which asserted both that no diagnostics leaked **and** that the published text was exactly verbatim (`'ask amazing amazing like really good'`). Restoring marking makes the verbatim half false by design, so the check was **replaced**, not adapted, by

> `every live publication callback keeps rival CSRC identity out of transcript text`

The new check keeps the leak assertion (`{CSRC `, `⟦`, `↔` absent) and drops the exact-text assertion. The sibling check `actual Teams pipeline keeps confirmed public text verbatim when it detects a contest` was likewise replaced by `…brackets exact duplicated words symmetrically before API publication`.

**This is a pinned test being taken down.** It was pinned to record that transcript text stays untouched under contest; that guarantee is intentionally reversed here. Reviewers should agree with the reversal on its merits rather than read the diff as a test update.

### 2. The terminal renderer will mark Whisper's non-speech tags as contested

`clients/terminal/src/canvas/LiveTranscriptEngine.tsx` now detects contested runs by matching bracketed text:

```diff
-  const pattern = /⟦([^⟧]+)⟧\{([^}]+)\}/g;
+  const pattern = /\[([^\]]+)\]/g;
```

The old marker was unambiguous — nothing else in a transcript emits `⟦…⟧{…}`. `[...]` is not. **Whisper emits bracketed non-speech tags such as `[Music]`, `[inaudible]`, `[BLANK_AUDIO]`, and the terminal will render every one of them as a contested run**, on every platform, contest or not. This is a known consequence of the lighter notation, not an oversight, and it is not mitigated in this PR.

**Correction — 0.12.24 formation security review, 2026-08-21.** Three things this description did
not say, each verified against the diff:

- **The marker now enters public transcript text.** `teams-csrc-gmeet-pipeline.ts:572` changes
  `const desired = raw;` to `const desired = { ...raw, text: markTeamsWordContests(raw, contests) };`.
  Before this PR the marker was detection-only telemetry that never reached public text; after it,
  it rides into the API, the database and exports. The same diff deletes the "never enters the API,
  Dashboard, or exports" and "public text stays verbatim" lines from `README.md`,
  `TEAMS_CSRC_CONTESTED_TRANSCRIPTS.md` and the class docstring. **That reversal is the change**, and
  it is why this PR carries an open founder decision rather than a review nit.
- **The `[inaudible]` collision runs both ways.** `[inaudible]` is an existing golden-contract value
  (`core/meetings/contracts/flagged-issue.v1/golden/FlaggedIssue.human.json:8`). As well as Whisper's
  tags being read as contests, a contested run over that word emits text indistinguishable from the
  contract value — the marker can **forge** it. `wrapTokenRange()`
  (`teams-contested-word-marker.ts:262`) does not escape pre-existing brackets, so `the [tag] here`
  becomes `[the [tag] here]`.
- **Two clients now disagree about the same row.** `packages/transcript-rendering` and
  `services/dashboard/src` have no bracket handling, so the terminal hides the brackets and marks the
  run contested while the dashboard renders a literal `[words]`.

## Contribution rights


DmitriyG228:
`next` removed by the release lane: this PR targets `codex/1150-calendar-multi`, which merged to main this morning — merging here delivers nothing to any train. Re-badge after retargeting to `main` and going green there.

DmitriyG228:
Train-scope validation ejected this from 0.12.25: the bare `[words]` marker collides with ordinary transcript text. Whisper emits `[inaudible]`/`[Music]`, and our own contract golden contains `[inaudible]` — all now render as contested, brackets stripped, entity mentions dropped for the whole merged turn (`LiveTranscriptEngine.tsx:124`; reproduced 8/8 inputs). The change also moves markers into public API text, where the old `⟦…⟧` could not collide. Fix direction: carry contest provenance as segment metadata, not in-band text — or a delimiter speech can't contain.

Issue #1182, opened 2026-08-17
meeting-api: calendar sync refuses any ICS feed over 2 MB, and the size limit counts the history the parser discards


AzimovS:
### What happened

The terminal calendar panel showed this message after a feed connect:

```text
⚠ Feed connected but the first sync failed: the feed is too large (over 2 MB)
```

Expected: the feed imports the upcoming meetings with a joinable link.
Actual: no meeting row imports, and the panel shows the error above.

The message is honest about the size limit. The limit itself is the problem.

### Where the code refuses the feed

`fetch_ics` returns the error at
[`core/meetings/services/meeting-api/src/meeting_api/calendar_sync/adapters.py:37-38`](core/meetings/services/meeting-api/src/meeting_api/calendar_sync/adapters.py#L37-L38).

The limit is `MAX_ICS_BYTES` at line 16 of the same file. It is a module constant. The constant has
three references in the tree, and all three are in that one file. There is no environment variable,
no Helm value, and no compose value for it. An operator cannot change this limit.

### Why the limit is wrong

The comment at line 16 states the premise: `2 MB — a personal calendar feed is KBs`.

A secret iCal address from Google Calendar or Outlook exports the complete calendar. The export
carries the past events and the future events. The ICS address accepts no parameter for a time
range. A work calendar with several years of meetings passes 2 MB. For those users the feature
fails always, not sometimes.

The parser makes the mismatch clear. `parse_ics` reads a 14-day window only. See
`DEFAULT_HORIZON_DAYS = 14` and `DEFAULT_LOOKBACK_S = 900` at
[`calendar_sync/service.py:29-30`](core/meetings/services/meeting-api/src/meeting_api/calendar_sync/service.py#L29-L30).
The code downloads years of history and then keeps two weeks. The size limit counts the history
that the code discards.

### The result for the user

`run_user_sync` returns early at
[`calendar_sync/runner.py:33-36`](core/meetings/services/meeting-api/src/meeting_api/calendar_sync/runner.py#L33-L36).
No event parses. No meeting row imports. The background sweep repeats every
`CALENDAR_SYNC_INTERVAL_S` (300 seconds by default) and each pass fails the same way. Calendar sync
and auto-join stay dead for this user. The saved feed still reads as connected in the panel.

### The documented workaround does not work

[`docs/docs/how-to/calendar-sync.mdx:107`](docs/docs/how-to/calendar-sync.mdx#L107) gives this fix:
[...]

PR #1250, opened 2026-08-19
fix(meetings): make the ICS feed size cap an operator dial, sized for a real work calendar

DmitriyG228:
**Delivers issue:** #1182

The report was diagnosed to the line and every claim in it held against `e0b356d6`, the revision it
names. Verified before changing anything:

| Claim in #1182 | Verdict |
|---|---|
| Refusal at `calendar_sync/adapters.py:37-38`, constant `MAX_ICS_BYTES` at line 16 | Confirmed, exactly |
| Three references, all in that one file; no env, no compose value, no Helm value | Confirmed — `git grep MAX_ICS_BYTES` returned only those three |
| Parser keeps 14 days (`DEFAULT_HORIZON_DAYS = 14`) of what it downloads | Confirmed at `service.py:29-30` |
| `run_user_sync` returns early, no row imports, every sweep repeats it | Confirmed at `runner.py:33-36` |
| The size test runs before the HTML test, hiding the better message | Confirmed, and reproduced — see A4 |
| `len(resp.content)` means the body is already resident; the cap does not protect memory | Confirmed, and measured — see A3 |
| The limit measures the decoded body, since httpx decompresses | Confirmed; `aiter_bytes()` keeps that semantics, so a compressed feed is still capped on its decompressed size |
| Docs give a workaround that cannot work | Confirmed at `calendar-sync.mdx:107` and `troubleshooting.mdx:87`; both rewritten |
| No test covers this branch | Confirmed |

## What changed

`CALENDAR_MAX_ICS_BYTES` — a config.v1 `defaulted` key, plumbed through compose, Helm
(`meetingApi.maxIcsBytes`) and lite. Read at call time per the contract's no-boot-snapshot rule.
An unusable value warns and falls back rather than raising inside a sweep required never to raise.

**Default raised 2 MB → 10 MB.** Making the limit configurable alone would have left the stock
deployment refusing the same feeds; the constant's premise — *"a personal calendar feed is KBs"* —
is what the report disproves. 10 MB is roughly 15-25k VEVENTs. The sweep awaits each user's sync in
turn, so this bounds the service's peak, not a peak per connected user.

**The body streams.** The old check read `len(resp.content)`, so the full body was resident before
the size was consulted — the cap described memory it had not protected. The budget is now applied
per chunk and the transfer is abandoned. This subsumes the `Content-Length` guard the report
proposed: the per-chunk budget already stops the download, and under `Content-Encoding` the
declared length is not the size being capped anyway.

**The not-a-feed sniff moved ahead of the size budget**, so an oversize HTML login page gets the
message that teaches the fix. The oversize reason now names the configured limit and the key.

### Alternatives rejected

- **Raise the limit only** — moves the wall without removing it, and leaves the next operator with
  the same unappealable refusal at a new number.
- **Filter by time range before the size check** — not available. The report is right that a secret
  iCal address takes no time-range parameter, and any client-side filter would need the whole body
  first, which is the thing being guarded.
- **Incremental/streaming parse so size stops being a gate entirely** — `parse_ics` hands the whole
  document to `icalendar.Calendar.from_ical`; making the parse incremental is a rewrite of the
  parser, not of this refusal, and it would still want a ceiling. Streaming the *download* captures
  the memory benefit that was actually claimed here.

## Observation bundle

- **C1 — the reading, before the change.** Ran `git show origin/main:…/adapters.py`, `git grep
  MAX_ICS_BYTES`, and read `service.py` / `runner.py` at `e0b356d6`. Saw every line the report
  names, unchanged. Concluded the report describes current `main`, not an older shape, so the fix
  belongs in `fetch_ics` and in the config contract rather than in the parser.
- **C2 — the contract, before writing an env read.** Read `deploy/contracts/config.v1/README.md`
  and `scripts/gates.mjs`. Saw that `gate:config-contract` scans every literal `os.getenv` and
  requires each declared key to appear in every surface its `targets` names. Concluded the key had
  to land in the declaration *and* in compose, Helm and lite together — which is also what makes it
  reachable without editing source, the thing the report asks for.
- **C3 — the test seam.** Read `webhooks/ssrf.py` and saw `127.0.0.0/8` on the blocklist, so a local
  HTTP server cannot be reached through `fetch_ics` at all. Concluded the tests had to drive the
  existing seam (`monkeypatch` of `build_pinned_transport` with an httpx `MockTransport`), the same
  one `test_calendar_sync_edges.py::test_fetch_ics_error_taxonomy` already uses.
- **C4 — red before green.** Ran the new suite against the pre-fix `adapters.py` restored from
  `origin/main`: 11 failed, 5 passed. The 5 that passed are the no-regression guards, which is what
  they are for. Then restored the fix: 17 passed.

## Acceptance floor

| Row | Evidence |
|---|---|
| **A1** — an operator can change the limit without editing source | `test_operator_can_raise_the_cap_and_the_feed_then_imports`, `test_operator_can_lower_the_cap`. Pre-fix: `AttributeError` / refusal — there was no key. `helm template --set meetingApi.maxIcsBytes=52428800` renders `value: "52428800"`; compose renders `CALENDAR_MAX_ICS_BYTES=${CALENDAR_MAX_ICS_BYTES:-10485760}`; lite exports it. `gate:config-contract` green: *declarations ≡ deploy surfaces ≡ code reads*. |
| **A2** — a real work calendar imports on stock config | `test_work_calendar_with_years_of_history_imports_its_next_meeting` — 4 MB of history plus one upcoming meeting, through `run_user_sync`. Post-fix: `counts.created == 1`, a `scheduled` row with the Meet link. Pre-fix: `last_error: 'the feed is too large (over 2 MB)'`, zero rows. |
| **A3** — the cap bounds memory instead of describing it | `test_oversize_body_is_abandoned_mid_download`. Measured with a bounded 16 MB body: pre-fix read **256/256 chunks** then refused; post-fix reads **16/256** and stops. |
| **A4** — the teachable message is no longer hidden | `test_oversize_html_page_gets_the_teachable_reason_not_too_large`. Pre-fix returns `the feed is too large (over 2 MB)` for an HTML login page; post-fix returns the *Secret address in iCal format* guidance. |
| **A5** — the refusal is actionable when it does fire | `test_oversize_reason_names_the_configured_limit_and_the_key` — the reason carries the configured cap and `CALENDAR_MAX_ICS_BYTES`, not a hardcoded "2 MB". |
| **A6** — no regression in the rest of the fetch taxonomy | `test_status_and_redirect_reasons_are_unchanged`, `test_small_feed_shorter_than_the_sniff_window_still_validates`, `test_leading_whitespace_does_not_starve_the_sniff`, `test_declared_charset_is_honoured`. Full service suite: **931 passed, 5 skipped**. |
| **A7** — a bad value cannot break the sweep | `test_unusable_values_fall_back_to_the_default` over `""`, `"   "`, `"not-a-number"`, `"0"`, `"-1"`. |

## Docs diff (D6c)

- `docs/docs/how-to/calendar-sync.mdx` — the failure-table row. The old fix text (*"sync a specific
  calendar rather than an aggregate"*) is the workaround the report shows cannot work when one
  primary calendar already exceeds the cap; it now leads with raising the value and keeps the
  single-calendar suggestion as the fallback it actually is.
- `docs/docs/troubleshooting.mdx` — stopped stating the limit as fixed expected behaviour, and says
  why a real feed can be large.
- `docs/docs/configuration.mdx` — the key in the meeting-api variable table, beside
  `CALENDAR_SYNC_INTERVAL_S`, with the Helm value named.
- `calendar_sync/README.md` — the module surface line, which stated the 2 MB cap.

## Security checks

The change narrows rather than widens the fetch: same SSRF-pinned transport, same
`follow_redirects=False`, and the body is now bounded *during* transfer instead of after it, so an
oversize or endless response can no longer be fully materialized. The new key is a non-secret
integer (`secret` flag absent, correctly). No dependency changes — `httpx` is already required and
`client.stream` is the same client. Gates run: `readme`, `docs-version`, `isolation`, `exports`,
`graph`, `schema`, `contract-version`, `config-contract`, `db-schema`, `db-budget`, `dataflow`,
`calm`, `licenses` — all green.

One residual worth naming rather than hiding: **the cap now bounds the download, not the parse.**
`parse_ics` still materializes an `icalendar` object graph several times the feed's size, so a 10 MB
feed costs meaningfully more than 10 MB at parse time against meeting-api's 1 Gi limit. That is
bounded and tunable downward now, where before it was neither — but a parse-side budget, or the
incremental parse described above, is a separate piece of work and is **not** in this PR.

## Validation request

Best witnessed by **@AzimovS**, who reported it and has the feed that produced the error — the
check is simply that the same feed now connects and imports its upcoming meetings, on stock config.
Any competent non-author can witness it from the test fixtures instead.

Deployment validated (D12b): **none**. This is offline evidence only — the service test suite and
the gates, run against this branch at `2d3dac75`. Neither Lite, compose, k8s nor hosted has been
run with this change, and no live Google or Outlook feed has been fetched through it. Those stay
honestly unclaimed.

## Contribution rights

Left unselected deliberately — the template calls this a legal certification an agent must not
make. The rights check will read as unsatisfied until the submitting human ticks exactly one.

## Authorship

Authored in an agent session on the maintainer's behalf; the commit carries the maintainer's own
DCO sign-off, matching this repository's existing convention. No agent co-author trailer (D13).


DmitriyG228:
**Conflict resolution — merged `main` (`4647c96e`) in; `DIRTY` → `MERGEABLE`.** Merge commit, no force-push. Two files conflicted, both in `calendar_sync/`.

**Two independent improvements to the same function had to be composed, not chosen between.**

`main` merged [#1151](https://github.com/Vexa-ai/vexa/pull/1151), which added `build_ics_client()` and a `client=` parameter so one sweep shares a connection pool across a tick's feeds. This PR rewrote the same function to stream the body under an operator-dialled cap. Taking either side alone would have silently dropped the other: keeping this PR's side removes the shared pool that `__main__._calendar_sync_loop` and `runner.py` already call (`fetch_ics(cfg["ics_url"], client=client)`), and keeping `main`'s side reverts the cap to the 2 MB constant this PR exists to remove.

Composed both:

- `build_ics_client()` kept as the one place the pinned transport is configured — its callers are untouched.
- `fetch_ics(url, *, timeout_s, client=None)` keeps the `client=` parameter **and** streams under `max_ics_bytes()`.
- Extracted `_read_streamed(resp, cap)` so the borrowed-client and owned-client paths run the identical budget and sniff. A second copy of that loop is how the two would drift — and the borrowed path is the one where abandoning an oversize transfer matters more, since the client outlives the fetch.

`calendar_sync/README.md` took `main`'s plural-calendar semantics paragraph, with the `fetch_ics` bullet rewritten to describe both the streaming cap and the shareable client.

**Validation:** `meeting-api` suite **1182 passed, 3 skipped** — including this PR's `test_calendar_ics_limit.py` and the sweep tests that drive the shared client. All 14 pre-push static gates green, `gate:config-contract` included (the new `CALENDAR_MAX_ICS_BYTES` key). No container workloads were run on this machine, so the compose/Helm/lite surfaces in this PR are unexercised locally; CI owns those.

PR #1252, opened 2026-08-19
feat(remote-browser): stealth plugin + WebGL fingerprint spoof

jbschooley:
## What
Anti-bot-detection hardening for the meeting browser.

1. **Wire the stealth plugin.** `playwright-extra` was imported but no plugin was ever registered — it ran as vanilla Playwright. `chromium.use(StealthPlugin())` (puppeteer-extra-plugin-stealth) now applies its evasions to every launch (masks `navigator.webdriver`, plugins, `window.chrome`).

2. **Spoof the WebGL/GPU fingerprint.** The bot's container has no GPU, so WebGL reported `SwiftShader` — a strong VM/bot tell that the stealth plugin's own WebGL evasion doesn't cover under playwright-extra. An init-script reports a plausible Intel/Mesa GPU for the UNMASKED vendor/renderer (with `getParameter.toString()` kept native) and restores `navigator.deviceMemory` + `chrome.runtime`.

## Evidence
- Fingerprint probe (stealth active): `navigator.webdriver` false, WebGL renderer → Intel/Mesa (was SwiftShader), `deviceMemory` 8, `chrome.runtime` present.
- **Zoom** admits the bot with these changes.
- **Honest scope:** this did NOT clear Google **Meet**'s "potential risks" classification — Meet weighs server-side signals (IP reputation, behaviour) these can't touch. Kept because it helps Zoom admission and is correct browser hygiene; not a claimed fix for Meet detection.

## Stacking
Stacks on #1130 (nonroot-chrome). Adds one dependency: `puppeteer-extra-plugin-stealth` (MIT → Category A; `gate:licenses` green).


DmitriyG228:
**Partially delivered — rung: installed and verified in production**, under your authorship.

Your two stealth commits (`1b81dc46` plugin, `780be04b` WebGL/GPU spoof) are in main via the 0.12.26 train drain (#1371), shipped in estate entry seq-12, and running in production as of 2026-08-31 — default **on**.

You were right and we had this filed wrong. It was triaged here as admission *optics*; a controlled differential settled it: same rig, same meeting, only the fingerprint changed, and a Zoom room that had refused four of our bots that day admitted the stealth build. That matches exactly what you reported on 2026-08-18 — *"added puppeteer stealth and some spoofing for browser signals and now it joins zoom"*. It is the Zoom-joining fix, and it unblocked our ability to witness Zoom on our own rig for the first time.

**Not aboard:** the other commits on this branch, including `e1103ef4` (`joining → needs_help`) — that FSM edge would let a never-joined bot reconcile to `completed`/`left_alone` and would relabel `@joining` failures as `awaiting_admission`; we hardened against it instead, and the typed join-failure work that shipped alongside makes those causes visible instead.

One follow-up filed as #1372: the spoofed renderer is a constant in a public repo, so every Vexa bot reports the same fake GPU — a fleet signature that fails everywhere at once the day someone writes a rule for it. Per-bot derivation is the fix.

Leaving this PR open for the remaining commits; the two that shipped are noted above.

PR #1253, opened 2026-08-20
fix(release): re-pin the v0.12.23 stable-promotion map identity to the rc.21 freeze

DmitriyG228:
The `:v012` promotion arm of `release-validate.yml` pins `EXPECTED_MAP_SHA256=76df3a63…`, a hash of an earlier `releases/v0.12.23/candidate-images.json` freeze. The map was re-frozen through the rc cycles and only the validate arm (line 253) was re-pinned; the shipped map at tag `v0.12.23` hashes `c5a31046…`. Both arms hash the same file and must agree — v0.12.18 pins one value in both. Without this, the stable-promotion dispatch for v0.12.23 refuses against the exact map it shipped.

One line: `76df3a63…` → `c5a31046…` in the promotion arm.

## Contribution rights

### DIFF
--- a/.github/workflows/release-validate.yml
+++ b/.github/workflows/release-validate.yml
@@ -1001,7 +1001,7 @@ jobs:
               EXPECTED_MAP_SHA256=80fe23246f94557279b4da2792119489f7c60f8e452b83de2c1c33ef48ebd03f
               ;;
             v0.12.23)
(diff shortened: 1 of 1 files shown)


DmitriyG228:
Merging by admin override on the founder's direct instruction ("fix that" — compose users pull :v012 which still points at 0.12.18; this PR unblocks the promotion arm). Gate checkboxes deliberately left unticked rather than filled by an agent; the override itself is the founder-ordered act.

Issue #1260, opened 2026-08-20
[DEV] 0.12.23 collapses awaiting_admission_timeout into join_failure — lobby timeouts now indistinguishable from our own join breakage


DmitriyG228:
Post-rollout hygiene on prod found a completion-reason vocabulary collapse. **Pre-deploy (7d): `awaiting_admission_timeout` 77, `join_failure` 27. Post-deploy (~16h): `awaiting_admission_timeout` 0, `join_failure` 8** — combined rates consistent (~0.6/h vs ~0.5/h), and rows post-deploy show `failure_stage: awaiting_admission` with `completion_reason: join_failure` (e.g. meetings 26603, 26605), while their bot logs read "No admission indicators after timeout" after a full 900s lobby wait. The lobby-timeout outcome is now being recorded as `join_failure`.

**Why it matters:**
1. It regresses denial transparency (#1178's direction): `join_failure` is strictly less specific than `awaiting_admission_timeout`.
2. It corrupts the failure taxonomy consumers: [Meeting-Failure-Taxonomy] treats `JOIN_FAILURE` as one of only two *our-system-broke* classes while admission timeout is benign — post-0.12.23 the benign class drains into the "ours" bucket, poisoning any rule or dashboard keyed on it.
3. Retry classification likely unaffected (both retryable) but should be re-checked once the reason is restored.

Expected: restore the distinct `awaiting_admission_timeout` reason for the no-admit path (stage already carries the truth), with a regression test asserting the pair (stage, reason) for a lobby-timeout exit.

PR #1272, opened 2026-08-20
mcp: report_issue — let the calling agent tell us what broke

DmitriyG228:
**Delivers issue:** [DmitriyG228/biz#434](https://github.com/DmitriyG228/biz/issues/434) — *the ticketing MCP: let the agents that use Vexa tell us what broke.*

Every instrument in the growth loop **reads**. This adds the one that lets an agent **write back**: a tenth tool on the MCP surface, `report_issue`, so a coding agent that hits a wall mid-integration can say what it tried, what happened, and on which deployment — instead of a human deciding, later, to open a GitHub issue.

The field that makes it worth more than a complaint is `meeting_id` + `platform`: the join key onto our own record of the same meeting.

## Contribution rights

- [ ] **Employer/client authorization required** 

Commit carries the DCO `Signed-off-by`.

> **Updated 2026-08-21** — second pass on this branch (`d08c02e4`): the route is now shaped to the
> ticket model in the design note (Linode's `POST /v4/support/tickets`), the entity pointer is
> authorisation-checked, and the prod owner's handoff conditions are built in rather than left for
> review. The sections below describe what is on the branch now.

## What ships

- `POST /report-issue` (`operation_id="report_issue"`) in `core/meetings/services/mcp/src/vexa_mcp/app.py` — a thin FastAPI route, exactly like every other tool in the service, so `FastApiMCP` derives the MCP tool from it. **The docstring is the product**: it is what a calling agent reads to decide whether to file.
- **Stateless and env-configured**, like the rest of the service. The ticket is POSTed to `VEXA_TICKET_SINK_URL` (optional `Authorization: Bearer $VEXA_TICKET_SINK_TOKEN`). **Unset → a clear 503** saying ticketing is not configured on this deployment: a self-hoster who sets nothing gets a no-op, not a crash. No secrets and no GitHub token in this service.
- **The caller's API key is never forwarded to the sink.** The payload carries `caller_fingerprint` — a salted SHA-256 prefix (16 hex) of the key, salt from `VEXA_TICKET_FINGERPRINT_SALT`. Enough to join two tickets from one account; useless as a credential if the sink or its logs leak. The rationale is stated in a comment on `_caller_fingerprint()`.
- **The ticket model is Linode's** (`POST /v4/support/tickets`), per the design note: the sink receives the canonical `summary` (≤64 chars) + `description` pair — composed server-side from the agent's answers so the MCP tool and any later HTTP ticket surface land **one shape** — plus optional `severity` (1/2/3) and `version`. The response mirrors their ticket object: `id` · `status` · `severity` · `opened` · `updated` · `opened_by` · `entity`. This service holds no store, so `id` is the sink's when the sink issues one and the content fingerprint otherwise.
- **The entity pointer is authorisation-checked.** Theirs is `linode_id`; ours is `meeting_id` + `platform`. A supplied id is resolved by asking the **gateway** for the caller's meetings **with the caller's own key** — a caller can only ever resolve a meeting they own, and the check is the gateway's, not a trust decision made in this service. An unowned, unknown or unresolvable id still files the ticket, quoted as text with `entity: null`: a ticket we refused to accept teaches us nothing.
- Forwarded payload also carries a server-side `reported_at`, a content-derived `fingerprint` (deployment + first 200 chars of `what_happened`) for downstream dedupe, and capped `logs` with a `logs_truncated` flag.
- **Defensive by construction:** a 64 KB ceiling on a declared `Content-Length` (413) — a check in the handler, not ahead of the parser: a malformed body is rejected by validation (422) before the cap is reached, and a chunked body declaring no length skips it entirely, `logs` capped at 4000 chars, text fields at 2000, empty/blank required fields rejected (422), `severity` range-checked, auth still required (401), the sink hop's failures mapped to typed 502/504/503.
- **SSRF closed by construction:** there is deliberately **no url-shaped field**, and **nothing a caller sends is ever dereferenced**. The only URL this route opens is the operator's env-configured sink. Links belong in the text, where a human reads them — the docstring says exactly that to the agent.
- **No path to account state:** this service has no DB, no ORM, no redis. A test walks the package's imports to keep it that way, so a ticket write cannot reach meetings or users.
- **Ticket text is DATA, never instruction** — validated, capped, forwarded verbatim, never interpreted by this service and never allowed to steer an agent of ours. Stated in a comment on the route.

Compose passes the three env vars through (all empty by default, so stock behaviour is unchanged); the service README gains a *Ticketing configuration* section and the seam row.

- **C1 — the tool actually appears on the MCP surface.** ran: `uv run pytest -k tools_match` after adding `report_issue` to `EXPECTED_TOOLS` · saw: green; the surface test is exact-set, so it would have failed had the route not become a tool · concluded: an MCP client sees ten tools, not nine.
- **C2 — the rendered docstring is what an agent will read.** ran: built the app and printed `mcp.tools['report_issue'].description` · saw: the full when-to-call text and all six argument descriptions rendered into the MCP schema · concluded: the product surface is the docstring, and it survives the FastApiMCP derivation intact.
- **C3 — the seam behaves against a fake sink.** ran: 19 tests in `tests/test_app.py` driving the SHIPPED app with the repo's `httpx.MockTransport` idiom (no docker, no network) · saw: forwarding, fingerprint stability, 503-without-sink, 502-on-sink-failure, truncation, empty-body rejection, entity resolution and its three refusal paths all green · concluded: the route is correct at its only two boundaries — the caller and the sink.
- **C4 — three negative controls, because a green assertion proves nothing until it has been made to fail.** ran: (a) replaced `_caller_fingerprint(api_key)` with the raw `api_key` in the payload · saw: `FAILED …::test_report_issue_never_forwards_the_api_key`. (b) removed the body ceiling · saw: `FAILED …::test_report_issue_body_size_cap - assert 200 == 413`. (c) fabricated the `entity` from the caller's input instead of from the gateway's answer · saw: `FAILED …::test_report_issue_meeting_absent_from_callers_meetings_does_not_resolve`. concluded: the key-leak, input-bound and authorisation-check assertions all bite. Each reverted; the suite was green again after every one.
- **C6 — the control that found a weak test.** ran: negative control (c) against my FIRST ownership test · saw: it stayed **green** under the fabrication, because the case it exercised (gateway 403) reached the `except` branch either way · concluded: the load-bearing case is a **200 carrying someone else's meetings** — the entity must come from the gateway's answer, never from the caller's input. That test is now in the suite and it is the one that goes red.
- **C5 — pre-push static gate suite.** ran: `git push` with the repo's 14 fast gates (readme, dataflow, isolation, isolation-py, exports, graph, graph-py, schema, contract-version, config-contract, licenses, execution-env, test-isolation, contract-conformance) · saw: `✓ pre-push: static gates green` · concluded: no structural or contract regression from the diff.
- **C7 — the sink did not need new infrastructure.** ran: added `VEXA_TICKET_SINK_FORMAT` (`raw` default · `github`), pointed `VEXA_TICKET_SINK_URL` at `https://api.github.com/repos/{owner}/{repo}/issues`, and ran the service locally against the REAL `api.cloud.vexa.ai` gateway · saw: a real issue created, [`DmitriyG228/biz#436`](https://github.com/DmitriyG228/biz/issues/436) · concluded: GitHub's issue API can BE the sink. The blocker was never missing infrastructure.
- **C8 — driven the way a third-party agent drives it, not by curling the route.** ran: a minimal MCP client over streamable HTTP at `/mcp` — `initialize` → `tools/list` → `tools/call` with `report_issue` and a real `meeting_id` · saw: `initialize` → `Vexa MCP Service (v0.12)`; `tools/list` → ten tools including `report_issue` with its full docstring; `tools/call` → `{"id": 436, "url": "https://github.com/DmitriyG228/biz/issues/436", "entity": {"type": "meeting", "id": "98579493833", "platform": "zoom", "url": "/transcripts/zoom/98579493833"}, "opened_by": "54f9d3e46eaf62cb", ...}` · concluded: the acknowledgement carries the issue number and `html_url`, so a calling agent can tell its human where the report went — and the entity resolved through the CALLER'S key at the real gateway, so the ownership check is the gateway's, live.
- **C9 — the join key and the label survive the round trip.** ran: read the created issue back over the API · saw: label `state: incoming` applied (GitHub created it on first use), and the body carrying `meeting_id` + `platform` + the resolved entity under their own `### Join key` heading, with every other canonical field (deployment, version, severity, logs, both fingerprints) present · concluded: what the sink stores is the whole ticket, not a summary of it.
- **C10 — the negative control, re-run against the LIVE artefact.** ran: grepped the full JSON of the created issue for the caller's Vexa API key and for the sink token · saw: both absent; only `caller_fingerprint` (`54f9d3e46eaf62cb`) is present · concluded: the no-credential-in-the-sink guarantee holds through the new wire shape, on a real artefact and not only in a mock.
- **C11 — `raw` is byte-unchanged.** ran: a test asserting the exact payload key set and the exact header set with no format env, plus one asserting an unknown format falls back to `raw` · saw: green · concluded: a self-hoster with an opaque webhook sees no difference, and an unset sink still answers a clean 503.

Base `98fab871` (origin/main) → head `edf6375e`.

```
$ uv run pytest -q
102 passed in 19.36s      # 93 → 102: nine new tests for the sink format switch
```

```
$ uv run pytest -k "format or github or raw_is" -q
9 passed
#   test_raw_is_the_default_and_is_byte_unchanged
#   test_unknown_format_falls_back_to_raw
#   test_github_format_maps_the_ticket_onto_an_issue
#   test_github_body_calls_out_the_meeting_join_key
#   test_github_body_states_when_no_meeting_was_supplied
#   test_github_labels_default_to_state_incoming
#   test_github_labels_are_configurable
#   test_github_response_number_and_html_url_reach_the_calling_agent
#   test_github_format_still_never_forwards_the_api_key
```

```
$ uv run pytest -k report_issue -v -p no:warnings
tests/test_app.py::test_report_issue_without_sink_returns_503 PASSED     [  5%]
tests/test_app.py::test_report_issue_forwards_ticket_to_sink PASSED      [ 10%]
tests/test_app.py::test_report_issue_response_mirrors_the_ticket_object PASSED [ 15%]
tests/test_app.py::test_report_issue_id_falls_back_to_fingerprint PASSED [ 21%]
tests/test_app.py::test_report_issue_severity_out_of_range_rejected PASSED [ 26%]
tests/test_app.py::test_report_issue_fingerprint_is_stable_and_content_derived PASSED [ 31%]
tests/test_app.py::test_report_issue_never_forwards_the_api_key PASSED   [ 36%]
tests/test_app.py::test_report_issue_caller_fingerprint_is_stable_per_key PASSED [ 42%]
tests/test_app.py::test_report_issue_truncates_oversized_logs PASSED     [ 47%]
tests/test_app.py::test_report_issue_caps_long_text_fields PASSED        [ 52%]
tests/test_app.py::test_report_issue_rejects_empty_fields PASSED         [ 57%]
tests/test_app.py::test_report_issue_requires_auth PASSED                [ 63%]
tests/test_app.py::test_report_issue_sink_failure_surfaces_as_502 PASSED [ 68%]
tests/test_app.py::test_report_issue_without_meeting_id_never_touches_the_gateway PASSED [ 73%]
tests/test_app.py::test_report_issue_resolves_entity_only_for_a_meeting_the_caller_owns PASSED [ 78%]
tests/test_app.py::test_report_issue_unowned_meeting_id_files_without_an_entity PASSED [ 84%]
tests/test_app.py::test_report_issue_meeting_absent_from_callers_meetings_does_not_resolve PASSED [ 89%]
tests/test_app.py::test_report_issue_has_no_url_shaped_field_and_fetches_nothing PASSED [ 94%]
tests/test_app.py::test_report_issue_body_size_cap PASSED                [100%]
====================== 19 passed, 74 deselected in 0.46s =======================
```

## Acceptance floor

| Row | Evidence |
|---|---|
| A1 — one tool `report_issue` on the hosted MCP endpoint, fields `what_i_tried` · `what_happened` · `deployment` · optional `logs` | `test_mcp_surface.py::test_mcp_mounted_and_tools_match` (exact-set), `test_report_issue_forwards_ticket_to_sink` |
| A2 — the join key: a ticket can carry the meeting id | `meeting_id` + `platform` fields, asserted in the forwarded body; the docstring tells the agent when to include them |
| A3 — deduped by fingerprint | `test_report_issue_fingerprint_is_stable_and_content_derived` — same deployment+symptom ⇒ same key; different symptom ⇒ different key |
| A4 — authenticated tickets carry identity, and nothing more | `caller_fingerprint` only; `test_report_issue_never_forwards_the_api_key` + `test_report_issue_caller_fingerprint_is_stable_per_key`, with the C4 negative control |
| A5 — rate-limited / bounded | `test_report_issue_truncates_oversized_logs`, `test_report_issue_caps_long_text_fields`, `test_report_issue_rejects_empty_fields` |
| A6 — a ticket is data, never instruction | forwarded verbatim, never parsed or executed; comment on the route states the rule |
| A7 — no auto-close, no auto-reply, no filing against third-party repos | this service only POSTs to the operator's own configured sink; it has no GitHub credential and no reply path |
| A8 — the prod owner's handoff conditions (design note § 4b) | per-IP limit + body cap at the gateway: **body cap held here (`test_report_issue_body_size_cap`), per-IP is the gateway half and is NOT built** · no URL-shaped field the server fetches: `test_report_issue_has_no_url_shaped_field_and_fetches_nothing` · sink isolated from the meetings DB: `test_ticket_path_touches_no_database` (AST walk over the package's imports) · 503 with no sink env: `test_report_issue_without_sink_returns_503` · text is data, never instruction: forwarded verbatim, never parsed |
| A9 — the ticket model matches the reference (Linode) | `test_report_issue_response_mirrors_the_ticket_object`, `test_report_issue_id_falls_back_to_fingerprint`, `test_report_issue_severity_out_of_range_rejected`; canonical `summary`/`description` asserted in `test_report_issue_forwards_ticket_to_sink` |
| A10 — entity resolution is authorisation-checked | `test_report_issue_resolves_entity_only_for_a_meeting_the_caller_owns` (lookup carries the caller's key), `test_report_issue_unowned_meeting_id_files_without_an_entity`, `test_report_issue_meeting_absent_from_callers_meetings_does_not_resolve` |
| A8 — self-hosters are not broken by it | `test_report_issue_without_sink_returns_503` + compose defaults empty |

**Exceeded:** the sink hop's failure modes are typed (502 rejected / 504 timeout / 503 unreachable) rather than surfacing as a 500, and `logs_truncated` tells the reader the paste is partial.

## Docs diff (D6c)

No docs pages in this PR — deliberately. The route's docstring and every field description are written to be the source a docs page is authored from, and the service README documents the three env vars and the payload shape. The public docs page belongs with the API-endpoint half described below.

- **No new dependency** — the route uses `httpx`, `hashlib` and `datetime`, all already present. Licence surface unchanged; `gate:licenses` green in the pre-push suite.
- **Secrets:** none added to the repo. The sink URL and token are env-only; no GitHub token reaches this service.
- **Credential exposure:** the key-leak assertion is explicit and negative-controlled (C4). The sink hop sends only the sink's own token.
- **Injection posture:** ticket text is never interpreted, never templated into a command, and never returned to another agent by this service.
- **Denial of service:** every caller-supplied field is bounded before it leaves the process; the route is behind the same authenticated edge as every other tool.

## Validation request

Any competent non-author. Watch: point a real MCP client at the service with `VEXA_TICKET_SINK_URL` set to a request-bin, call `report_issue` from a coding agent, and confirm (a) the agent chooses to call it from the docstring alone, (b) the body arrives with `meeting_id` and `caller_fingerprint` and **without** the API key, (c) with the env var unset the tool answers a clean 503 and nothing else in the service changes. **Deployment validated so far: none** — this is module-level only, in-process against a fake sink (`uv run pytest`, fresh worktree of `98fab871`, no env deltas). Compose / hosted are honestly unclaimed.

**Do not merge from here:** this is runtime code and rides the release train (scope → stage → prod).

## Deploying the sink — and where production points

**Nothing here is deployed.** The sink env stays **unset by default in OSS**, so cloud and every
self-host still answer the clean 503 until an operator sets `VEXA_TICKET_SINK_URL`. That default is
the self-hoster's story and it is preserved deliberately.

**Production must point at a DEDICATED tickets repository — never an operational one.** This is a
trust boundary, not tidiness. Ticket bodies are third-party text (the handler's own rule: *data,
never instruction*), and an operational repo's issues are read as work signal by planning and
automation loops — so filing untrusted text there would inject it straight into those loops. A
dedicated, private tickets repo enforces at the **storage** layer what the handler can only promise.
The live proof above used `DmitriyG228/biz` **only** because it was a one-off test, and that issue
was closed with its receipt.

**Token shape, minimum:** fine-grained, **single-repo**, `issues: write` **only**, mounted via
`secretKeyRef` on Kubernetes (or an env file elsewhere) — never an inline env value. The dedicated
repo is what scopes that token's blast radius.

Documented alongside the env vars in `core/meetings/services/mcp/README.md` and in
`deploy/compose/docker-compose.yml`.

## Known gap, stated rather than half-built

The follow-on scope is the **HTTP ticket surface** — `POST /support/tickets` (authenticated) and `POST /support/tickets/public` (**no key**), plus list/get/reply — on the gateway, so it is documentable like any other route. That is **not in this PR**, and the reason to draw the line here is the public door's abuse story: as the prod owner put it, Linode's model quietly relies on account auth for that story and ours deliberately does not, **so the controls have to live in the route**.

Concretely, the gateway half needs:

| | |
|---|---|
| **Per-IP rate limit at the gateway** | the existing `ratelimit.py` bucket keys on the resolved `user_id`; a key-less caller has none, so the public leg needs the `edge_guard` per-IP layer. **This is the one acceptance condition this PR cannot satisfy** — a handler-level cap is not a gateway-level control. |
| **Body-size cap at the gateway** | held in this handler (413 at 64 KB); the condition asks for it at the edge as well, before an unauthenticated body reaches a worker. |
| **Route declaration** | `ROUTE_SCOPES` / `UNSCOPED_ROUTES` entries — the edge is deny-by-default with a build-time assertion exhaustive over the router, so the public flavour must be declared unscoped deliberately, not by omission. |
| **Wiring** | `config.v1.json` + `adapters.py` for the sink or the downstream hop, and the gateway's own conformance tests (`test_scope_matrix`, `test_proxy`). |
| **Anonymity** | the design note's § 4 gate: the public door stores no identity — no IP retained beyond the rate-limit window, no key, no email unless the reporter typed one. |
| **Sequencing** | [`#1260`](https://github.com/Vexa-ai/vexa/issues/1260) (completion-reason collapse) is load-bearing for the join and should land in the same train — a ticket's value is the pairing to the meeting's exit reason. |

It crosses a boundary worth verifying on its own diff rather than bolting onto this one.

## Authorship
Tooling disclosure (optional, welcome, never an attribution): drafted with Claude Code.


DmitriyG228:
**0.12.24 car (`angry-margulis-b594ff`, [biz#435](https://github.com/DmitriyG228/biz/issues/435)) — two red checks stand between this PR and the train it justifies.**

I am not touching this branch: `mcp-report-issue` is checked out in another session's worktree (`/private/tmp/vexa-ticket-mcp`). Both findings are for whoever holds it.

### 1 · CodeQL — one new high-severity alert, and my read is that it is a false positive with a real remark attached

`py/weak-sensitive-data-hashing` at [`core/meetings/services/mcp/src/vexa_mcp/app.py:103`](https://github.com/Vexa-ai/vexa/blob/mcp-report-issue/core/meetings/services/mcp/src/vexa_mcp/app.py#L103) — *"Sensitive data (password) is used in a hashing algorithm (SHA256) that is insecure for password hashing."*

The flagged call is `_caller_fingerprint`, and it is not password storage: it is a salted SHA-256 prefix used as a pseudonymous join key for the ticket sink, and the docstring already states that intent. CodeQL classifies the input as a password because it is credential-shaped; the "computationally expensive hash" advice is aimed at verifying user-chosen passwords, which this is not.

The remark that is real, and worth one line in the code rather than a dismissal note: the salt falls back to `_DEFAULT_CALLER_SALT`, a constant in the source. With the default salt in force, anyone holding the sink's contents plus the repository can confirm a guessed API key against a fingerprint. That is only as strong as the key's own entropy — fine for randomly generated keys, weak for anything guessable.

**Two ways out, and this is the branch owner's call, not mine:** dismiss the alert as a false positive with that justification written down, or make the fingerprint an HMAC over a salt that has no in-source default and fails closed when unset. I lean to the second — it removes the remark rather than annotating it — but either unblocks the train and I will not hold the car for the difference.

### 2 · docs-current — this PR ships a product surface with no docs

> `#1272: ❌ touches a product surface but changes no docs/** and lacks the `docs: none` label`

`report_issue` is a new tool on the MCP service — a caller-facing capability. `docs: none` would be the wrong answer here; the gate is right. It needs the MCP tool documented (and a `docs/changelog.d/1272-report-issue.md` fragment, which is the collision-free way to get the changelog touch — do **not** edit `docs/changelog.mdx` directly).

Worth saying plainly in whatever docs land: the sink does not exist yet. `VEXA_TICKET_SINK_URL` is unset and the tool returns a clean 503, so a reader must not be told a report reaches a human today.

**One thing the car needs from the owner:** an ETA. This PR is the reason 0.12.24 exists, and both blockers are hours of work, not days. If nobody is holding this branch, say so on this PR and I will branch from it into my own worktree and finish it.

PR #1324, opened 2026-08-25
feat(gateway): consolidate the WS guard onto fastapi-guard 8.0.0 / guard-core 4.0.1 (guard_websocket + auto-ban knob + on_block)

### LINKED ISSUE #1323: Gateway: consolidate the WS guard onto fastapi-guard, replacing the hand-rolled rate-limit mirror
## Value this change must deliver

> The gateway's `/ws` connect path is protected by the same library (fastapi-guard / guard-core) as the HTTP pipeline, not a hand-rolled mirror. One config surface, one ban store, one rate-limit primitive; the `/ws` path gets an isolated rate budget and shares ban and IP-list decisions with HTTP. No hand-rolled security code left to drift from the library.

## Where we are (honest)

Phase 1 (`cd581856`) already routed the WS guard's ban check and IP-allow check through guard-core (`ip_ban_manager`, `is_ip_allowed`), making the ban store Redis-shared with the HTTP middleware. But the rate-limit half stayed hand-rolled: a per-process sliding window (`_rl`) plus a per-process auto-ban counter (`_ban_counts`), duplicating what the library now ships. guard-core 3.13.0 (released 2026-08-25) added `check_rate_limit_by_ip` and the A2b knob that wires the standalone primitive into the auto-ban engine, which makes the hand-rolled half deletable.

Disclosure: I author fastapi-guard and guard-core. This consolidates Vexa onto primitives I wrote; the library's behavior is the thing being adopted here, not audited.

## Components (implemented on `feat/guard-ws-consolidation`; PR to follow)

1. `_WsGuard.check` now calls `check_rate_limit_by_ip(ip, cfg, endpoint_path="ws")` for an isolated `/ws` budget, after the load-bearing ban check (the primitive counts but does not block banned IPs) and the `is_ip_allowed` gate.
2. `build_guard_config` adds `enable_rate_limit_auto_ban` (default ON via `GUARD_AUTO_BAN_RATE_LIMIT`) so one knob governs both the HTTP pipeline and the WS connect primitive. This preserves the old WS auto-ban and newly expands guard to HTTP rate-limit auto-ban.
3. Deleted the sliding-window `_rl`, `_ban_counts`, and the `time`/`collections` imports. The rate window stays in-process (`redis_handler=None`) by design; Redis-sharing is a one-line future enhancement, and the primitive fails open to in-memory on a redis outage.
4. Pin bumped fastapi-guard 7.2 to 7.8, lockstepping guard-core 3.4.0 to 3.14.0 (both released 2026-08-26).
5. `GUARD_AUTO_BAN_RATE_LIMIT` declared in the gateway `config.v1.json` and added to compose and helm.
6. Phase 3: the `/ws` check itself runs fastapi-guard 7.8.0's own `guard_websocket` (called directly from `run_multiplex` before `ws.accept()`), and every Vexa-authored WS guard symbol (`_WsGuardRequest`, `_WsGuard`, `ws_guard_check`, `resolve_ws_client_ip`) is deleted. The `/ws` rate window becomes Redis-shared when `GUARD_ENABLE_REDIS=true`, which retires the "future enhancement" in item 3.

## Behavior drift (intentional)

The library counter does not reset on ban, so after a ban expires a recidivist re-bans on the first over-limit (stricter on recidivism than the old reset-on-ban). Documented in-module and asserted by `test_auto_ban_persists_past_window_rebans_without_reset`.

## Retired configuration (phase 3)

`GUARD_WS_ENABLED=true` with `GUARD_ENABLED=false` was a working WS-only setup under the hand-rolled guard. `guard_websocket` resolves its config from the HTTP guard's registered middleware, so that combination now refuses to boot with a `ConfigError` naming both vars; stated as breaking-on-upgrade on every operator surface. Known gap kept visible: 7.8.0 rate-limits a whitelisted IP on `/ws` where HTTP skips it; the test stays as a strict xfail until the fastapi-guard 7.8.1 re-pin.

## Scope honesty (do not over-claim)

The cross-pod wins are the ban STORE (Redis-shared, phase 1) and the rate WINDOW (HTTP since phase 2, `/ws` since phase 3). The auto-ban violation COUNTS remain in-memory per-process (a 10k-IP FIFO), semantics-neutral versus the old `_ban_counts` on a multi-replica deployment. The WS guard remains opt-in (`GUARD_WS_ENABLED`, default false); this change swaps the implementation behind that flag, it does not flip the default.

## Verification

343 passed plus 2 xfailed (one strict whitelist-parity xfail, one pre-existing in `test_proxy.py`); ruff clean on touched files; pre-push static gates green; an adversarial review pass (fable-mode:critic) confirmed the load-bearing check order, the isolated `endpoint_path="ws"` budget, the knob pair, the no-reset drift, and the test isolation. Branch `feat/guard-ws-consolidation` on `rennf93/vexa`; PR #1324 implements this, stacked on #1157.

rennf93:
## Value this change must deliver

The gateway's `/ws` connect path is protected by the same library (fastapi-guard / guard-core) as the HTTP pipeline, not a hand-rolled mirror. One config surface, one ban store, one rate-limit primitive; the `/ws` path gets an isolated rate budget and shares ban and IP-list decisions with HTTP. No hand-rolled security code left to drift from the library.

Closes #1323.

## Stacking

Was stacked on #1157 (env validation). #1157 was closed as superseded on 2026-09-01; its commit rides in this branch, so this PR merges on its own.

Disclosure: I author fastapi-guard and guard-core. This consolidates Vexa onto primitives I wrote; the library's behavior is the thing being adopted here, not audited.

## Where we were

Phase 1 (`cd581856`) already routed the WS guard's ban check and IP-allow check through guard-core (`ip_ban_manager`, `is_ip_allowed`), making the ban store Redis-shared with the HTTP middleware. But the rate-limit half stayed hand-rolled: a per-process sliding window (`_rl`) plus a per-process auto-ban counter (`_ban_counts`), duplicating what the library now ships. guard-core 3.13.0 (released 2026-08-25) added `check_rate_limit_by_ip` and the A2b knob that wires the standalone primitive into the auto-ban engine, which makes the hand-rolled half deletable.

## What this changes

1. `_WsGuard.check` now calls `check_rate_limit_by_ip(ip, cfg, endpoint_path="ws")` for an isolated `/ws` budget, after the load-bearing ban check (the primitive counts but does not block banned IPs) and the `is_ip_allowed` gate.
2. `build_guard_config` adds `enable_rate_limit_auto_ban` (default ON via `GUARD_AUTO_BAN_RATE_LIMIT`) so one knob governs both the HTTP pipeline and the WS connect primitive. This preserves the old WS auto-ban and newly expands guard to HTTP rate-limit auto-ban.
3. Deleted the sliding-window `_rl`, `_ban_counts`, and the `time`/`collections` imports. The rate window stays in-process (`redis_handler=None`) by design; Redis-sharing is a one-line future enhancement, and the primitive fails open to in-memory on a redis outage.
4. Pin bumped fastapi-guard 7.2 to 7.8.1, lockstepping guard-core 3.4.0 to 3.14.0 (7.8.x declares `guard-core>=3.14.0`, so the 7.7 range already resolved there on a fresh lock). guard-core 3.14.0 hardens the primitives adopted here: bounded in-memory rate-limit and auto-ban stores, `Retry-After` on every 429, ban-address canonicalisation. fastapi-guard 7.8.0 itself joins repeated X-Forwarded-For field lines in its Starlette adapter before client resolution.
5. `GUARD_AUTO_BAN_RATE_LIMIT` declared in the gateway `config.v1.json` and added to compose and helm.
6. A connect with no resolvable peer address (`ws.client` is None: Unix-socket or a peer-less ASGI server) resolves to guard-core's `"unknown"` identity, and `check_rate_limit_by_ip` raises `ValueError` on a non-IP, which would have turned the upgrade into a 500. The ban and IP-list checks still run against the unknown identity; the rate-limit primitive is skipped for it, failing open the way the HTTP pipeline does under `fail_secure=False` and the way fastapi-guard 7.8.0's own `guard_websocket` dependency does (`test_unresolvable_client_fails_open_on_ws`).
7. **Phase 3 (`144725ae`): the `/ws` check itself moves onto fastapi-guard 7.8.0's own `guard_websocket`.** `run_multiplex` calls it directly before `ws.accept()` (not through `Depends`, so `create_app` stays pure and the FakeWS tests keep driving the check), closes with the library's code and reason on denial (1008 pre-accept, so the upgrade still answers HTTP 403) and logs one `ws_connect_rejected` event. Deleted from `edge_guard.py`: `_WsGuardRequest`, `_WsGuard`, the singleton and its reset, `ws_guard_check`, `resolve_ws_client_ip`. No Vexa-authored guard orchestration remains for either path; the module builds the one config and installs the one middleware.
8. Docs (`docs/docs/configuration.mdx`): new `GUARD_AUTO_BAN_RATE_LIMIT` row; the `GUARD_WS_ENABLED` row no longer claims the WS guard shares nothing with the HTTP layer (the ban store and IP-list decisions are shared since phase 1; only the `/ws` rate window is per-process).
9. **Redis-outage semantics pinned (`redis_fail_open=True`).** The lock moves to guard-core 3.15.0 / fastapi-guard 7.8.2 (the `>=7.8.1,<8.0` range already resolved to 3.15.0 on a fresh lock). 3.15.0's rate limiter honors `redis_fail_open` on a Redis failure instead of always falling back to the in-memory window; under the default `False` with Vexa's `fail_secure=False` the HTTP pipeline would pass each request through with no rate limit and `guard_websocket` would skip the `/ws` rate limit until Redis is back. `build_guard_config` now sets `redis_fail_open=True` explicitly, so on either side of that version boundary a Redis outage degrades both windows to their per-process fallback (one warning per process) and keeps enforcing; ban lookups fail open (not banned) exactly as before. Asserted in `test_config_safe_defaults`; the `configuration.mdx` outage sentences say so.
10. **Lock moves to guard-core 3.16.0 (2026-09-01).** The pin range (`fastapi-guard>=7.8.1,<8.0`, guard-core `>=3.14.0`) already admits it, so this is a lock-only bump. The release closes a TTLCache check-then-use race on the ban path (`is_ip_banned` / `unban_ip`): a `KeyError` at a TTL expiry boundary could 500 the request under `fail_secure=True` or, under the default `fail_secure=False`, skip the ban check entirely. That is exactly the path `GUARD_AUTO_BAN_RATE_LIMIT` and the WS ban store ride, so the fix is load-bearing for this PR. It also adds an optional `SecurityConfig.on_block` block-observation callback, wired in item 12.
11. **Lock moves to guard-core 3.17.0 (2026-09-01, released later the same day).** Lock-only again; the pin range is unchanged and fastapi-guard stays 7.8.2. 3.17.0 makes dynamic rules persist a last-known snapshot (Redis primary, optional file fallback via `dynamic_rules_cache_path`, hydrated at `initialize_agent` when `enable_dynamic_rules=True`). Inert for the gateway: `build_guard_config` never enables dynamic rules or the agent and no such knob exists in the gateway `config.v1.json` or the compose allowlist, so no config or docs change. Stale "pinned guard-core (3.15.0)" prose in `edge_guard.py`, the tests, `pyproject.toml` and `.env.example` now names 3.17.0, or says ">= 3.15.0" where the fact is a version boundary.
12. **HTTP blocks get a structured event, the twin of `ws_connect_rejected`.** `build_guard_config` sets `on_block=_on_http_block`, which emits one `http_request_blocked` (`audience="system"`, `level="warning"`, `span="http"`) with `check_name`, `reason`, `client_ip`, `path`, `method`, `status_code`, `passive_mode`. Before this, a guard 403 or 429 on the HTTP side left nothing in the gateway's structured log; only guard-core's own logger spoke. guard-core fires the hook once per blocked request (pipeline, unresolvable-client bypass handler, rate-limit handler), not for custom validators, the HTTPS redirect or a Redis-unavailable response, and it isolates a raising hook. `trigger_info` is not logged: it is empty for the IP and rate-limit checks and already embedded in `reason` for the suspicious-activity check (dormant here, `enable_penetration_detection=False`), where both carry attacker-controlled header and parameter names verbatim; `reason` and `path` are safe for this sink because `log_event` JSON-escapes every field, as `request_received` already relies on. A committed regression test asserts a WS rejection fires no `on_block`. `guard_websocket` never fires `on_block` (it calls the primitives directly), so a WS rejection still logs exactly once. `TestHttpBlockEvent` drives a real blacklist 403 and a real rate-limit 429 through the installed library and asserts the event; a payload with missing keys does not raise.
13. **Pin floor moves to fastapi-guard 8.0.0 / guard-core 4.0.0 (both released 2026-09-04).** `fastapi-guard>=7.8.1,<8.0` becomes `>=8.0,<9.0` (8.0.0 declares `guard-core>=4.0.0`); the lock bumps fastapi-guard 7.8.2 to 8.0.0 and guard-core 3.17.0 to 4.0.0 ([guard-core 4.0.0](https://github.com/rennf93/guard-core/releases/tag/4.0.0), [fastapi-guard 8.0.0](https://github.com/rennf93/fastapi-guard/releases/tag/8.0.0)). Why: on 2026-09-02, in this gateway's own container, with `GUARD_IP_BLACKLIST=0.0.0.0/0` a `curl -H 'X-API-Key: sekrit'` on `/health` logged `'x-api-key': 'sekrit'` verbatim in guard-core's WARNING line right next to the gateway's own `http_request_blocked` event, which itself never logged headers. guard-core 4.0.0 redacts `authorization`, `proxy-authorization`, `cookie` and `x-api-key` by default on every guard log line, event and hook payload, plus query-string and body secrets, with `log_sensitive_headers` / `log_sensitive_params` / `log_sensitive_body_fields` for extra names the gateway does not set; the 3.17.0 to 4.0.0 `SecurityConfig` default diff is exactly those three new empty frozensets. fastapi-guard 8.0.0's `guard_websocket` also now runs `SuspiciousActivityCheck` on the `/ws` handshake; inert here because `enable_penetration_detection=False` (proven by execution: attack-shaped headers and query strings on the `/ws` connect are accepted with no log line, same as on 7.8.2), but the scan resolves the route path before that gate and reads `scope["path"]` unconditionally, so `tests/test_edge_guard.py`'s `FakeWS` double gained `scope["path"] = "/ws"` (a real Starlette scope always carries it; the production call site `run_multiplex` passes a real `WebSocket`). Audited the rest of 4.0.0's breaking changes against this gateway: `require_headers()`'s exact-match change and `excluded_detection_headers` are both unused here; the rate-limit Redis key now hashes the endpoint segment (`rate:<ip>:<sha256(endpoint)>`), so existing per-IP counters reset once on this upgrade (noted on the `GUARD_REDIS_PREFIX` row of `configuration.mdx`); proxy identity headers (`x-forwarded-for`, `x-real-ip`, etc.) are default-excluded from the `ssrf` detection category only, not from detection generally, every other category still scans them.
14. **Lock moves to guard-core 4.0.1 (released 2026-09-05).** `fastapi-guard>=8.0,<9.0` already admits it (8.0.0 keeps declaring `guard-core>=4.0.0`), so this is a lock-only bump; fastapi-guard stays 8.0.0. Why it matters here: 4.0.1 fixes "a ban refused by the self-DoS guard was answered as a ban" ([guard-core 4.0.1](https://github.com/rennf93/guard-core/releases/tag/4.0.1)): `ban_ip` now returns `bool` instead of `None`, and `_resolve_and_apply_threshold_ban` returns `None` (no ban applied) when the target IP overlaps loopback or a configured `trusted_proxies` network, so that request instead gets the honest rate-limit verdict. This lands squarely on the path items 2 and 12 added: `GUARD_AUTO_BAN_RATE_LIMIT` (default on) feeding `check_rate_limit_by_ip(..., endpoint_path="ws")` inside `guard_websocket`, and the same threshold helper backing the HTTP pipeline's auto-ban. Audited every ban this suite drives (`test_auto_ban_persists_past_window_rebans_without_reset`, `test_ban_check_precedes_whitelist_bypass`): both target `10.0.0.7` against `trusted_proxies=["127.0.0.1"]`, an address outside loopback and outside the configured trusted-proxy range, so the ban still applies and no existing assertion changes; the fix only changes behavior for a ban that would target the reverse proxy's own resolved address, a case this suite never exercises but a misconfigured `GUARD_TRUSTED_PROXIES` in production could hit. Also in 4.0.1: singleton resets (`ip_ban_manager`, the rate-limit manager, the security-headers manager) now also drop `redis_handler`/`agent_handler`, inert here since this gateway's tests already reset state by swapping in a fresh manager instance, not by relying on a handler surviving reset; `token=<SECRET>`-style placeholder values are now redacted in guard-core's own log lines; and `redact_blob_for_display`/`redact_url_for_display` are now exported from `guard_core.utils`, already applied internally to the `path` field this gateway's `_on_http_block` logs (`build_block_payload` redacts it before the payload ever reaches Vexa's callback), and `ws_connect_rejected` logs no `path` at all, so no gateway code change follows from the export. Full local suite: 350 passed, 1 xfailed, identical to the 4.0.0 baseline. Container harness rebuilt from this branch's own lock (guard-core 4.0.1 confirmed inside the image): the 15-case matrix passed 15 of 15.

## Behavior drift (intentional)

Phase 3 adds two, both documented on compose, helm, `.env.example` and `configuration.mdx`: (a) the `/ws` rate window is Redis-shared with the HTTP layer when `GUARD_ENABLE_REDIS=true` (`guard_websocket` hands the library's `RedisManager` to the primitive; phases 1 and 2 kept it in-process by design), which retires the "one-line future enhancement" item 3 above mentions; (b) `GUARD_WS_ENABLED=true` with `GUARD_ENABLED=false`, a working WS-only setup under the deleted guard, now refuses to boot with a `ConfigError` naming both vars, because `guard_websocket` resolves its config from the HTTP guard's registered middleware. The check lives in `apply_guard`: `config_preflight.py` is byte-equality gated by `gate:config-contract` and cannot carry a gateway-specific rule. The one gap 7.8.0 had (`guard_websocket` rate-limited a whitelisted IP on `/ws` where the HTTP path skips it) is closed by fastapi-guard 7.8.1, released 2026-08-27; the branch is pinned `>=7.8.1` and `test_whitelist_cidr_bypasses_ws_guard` runs as a regular test again (its strict xfail is gone).

The library counter does not reset on ban, so after a ban expires a recidivist re-bans on the first over-limit (stricter on recidivism than the old reset-on-ban). Documented in-module and asserted by `test_auto_ban_persists_past_window_rebans_without_reset`.

## Scope honesty (do not over-claim)

The cross-pod wins are the ban STORE (Redis-shared, phase 1) and the rate WINDOW (HTTP since phase 2, `/ws` since phase 3). The auto-ban violation COUNTS remain in-memory per-process (a 10k-IP FIFO), semantics-neutral versus the old `_ban_counts` on a multi-replica deployment. The WS guard remains opt-in (`GUARD_WS_ENABLED`, default false); this change swaps the implementation behind that flag, it does not flip the default.

## Verification

350 passed plus 1 xfailed (pre-existing in `test_proxy.py`) on the branch against guard-core 4.0.1 / fastapi-guard 8.0.0, held identical against guard-core 4.0.0 / fastapi-guard 8.0.0 and against guard-core 3.17.0 / fastapi-guard 7.8.2 (346 / 1 xfail before the four `on_block` tests; the identical 346 / 1 xfail held on 3.15.0, 3.16.0 and 3.17.0) (one test assumption moved with 3.14.0: guard's 429 now carries `Retry-After` = the rate window, so the both-layers test tells the per-user limiter (1s) and the per-IP guard (60s) apart by header value instead of header presence); ruff clean on touched files; pre-push static gates green; adversarial review passes (fable-mode:critic) confirmed the load-bearing check order, the isolated `endpoint_path="ws"` budget, the knob pair, the no-reset drift, the test isolation, and the 3.15.0 Redis-outage semantics against the installed library source. Branch `feat/guard-ws-consolidation` on `rennf93/vexa`, rebased onto #1157's current head (which now carries the guard pin bump, so the two branches no longer disagree about the installed guard-core).

Verified in real containers as well as unit tests. The gateway image was built from this branch's own lockfile (fastapi-guard 7.8.2 / guard-core 3.15.0 confirmed inside the image) and booted across a 13 case matrix, all green; after the 3.16.0 lock bump the image was rebuilt from the new lock (guard-core 3.16.0 confirmed inside the image) and the boot matrix re-run, all green; after the 3.17.0 bump and the `on_block` wiring the image was rebuilt again (guard-core 3.17.0 confirmed inside the image), the 9-case core matrix passed 9/9, and the blacklist case now shows exactly one `http_request_blocked` for the `/health` 403 and exactly one `ws_connect_rejected` for the `/ws` 403 in the container log. After the 8.0.0 / 4.0.0 bump the image was rebuilt again from this branch's own lock (guard-core 4.0.0 and fastapi-guard 8.0.0 confirmed inside the image), a 15 case matrix passed 15 of 15, the blacklist case still emits exactly one `http_request_blocked` and one `ws_connect_rejected`, and the same curl now carrying `X-API-Key`, `Authorization` and `Cookie` produces this guard-core log line: `Headers: {'host': '127.0.0.1:18766', 'user-agent': 'curl/8.7.1', 'accept': '*/*', 'x-api-key': '[REDACTED]', 'authorization': '[REDACTED]', 'cookie': '[REDACTED]'}` with none of the secret values anywhere in the log. After the 4.0.1 lock bump the image was rebuilt again from this branch's own lock (guard-core 4.0.1 confirmed inside the image), and the 15 case matrix passed 15 of 15 again, with the same blacklist redaction and structured-event counts holding. Both cases below hold. The two that matter most here:

- `GUARD_WS_ENABLED=true` with `GUARD_ENABLED=false` refuses to boot with a `ConfigError` naming `GUARD_WS_ENABLED`, which is the retirement of that previously-valid WS-only config working as intended.
- With `GUARD_IP_BLACKLIST=0.0.0.0/0`, a client gets **403 on `/health` and 403 on the `/ws` upgrade**. That is the consolidation's whole point demonstrated end to end: the WS path is now covered by the same IP controls as the HTTP path, in a running container rather than against a FakeWS.

## Contribution rights


DmitriyG228:
This is the surviving lane for the guard bump — #1157 is closed in its favour (this PR is a strict superset of it).

rennf93:
Follow-up `044aa1cc`: this also moves the gateway's guard pin from fastapi-guard 7.8.2 / guard-core 3.17.0 to fastapi-guard 8.0.0 / guard-core 4.0.0 (the pin range `fastapi-guard>=7.8.1,<8.0` becomes `>=8.0,<9.0`). Reason: on 3.17.0, this gateway's own container showed guard-core logging a blocked request's headers verbatim, including an `X-API-Key` sent on a blacklisted call to `/health`, right next to the gateway's own block event, which never logs headers; guard-core 4.0.0 redacts `authorization`, `proxy-authorization`, `cookie` and `x-api-key` (plus query-string and body secrets) by default on every guard log line, event and hook payload, closing that leak.

Audited the rest of 4.0.0's breaking changes against this gateway: `require_headers()`'s exact-match change and `excluded_detection_headers` are both unused here; the rate-limit Redis key now hashes the endpoint segment, so existing per-IP counters reset once on this upgrade; proxy identity headers are default-excluded from the `ssrf` detection category only, every other category still scans them. fastapi-guard 8.0.0's `guard_websocket` also now runs `SuspiciousActivityCheck` on the `/ws` handshake, and that scan resolves the route path (reading `scope["path"]` unconditionally) before the `enable_penetration_detection` gate; `tests/test_edge_guard.py`'s `FakeWS` double gained `scope["path"] = "/ws"` to match the real route (a real Starlette scope always carries it, and the production call site `run_multiplex` passes a real `WebSocket`), and the scan itself stays inert here since `enable_penetration_detection=False`.

Verification: 350 passed plus 1 xfailed on the unit suite (identical on both version pairs), an in-process proof against the real `SecurityMiddleware` with a blacklisted IP, and a rebuilt container matrix at 15 of 15, where the same curl (now carrying `X-API-Key`, `Authorization` and `Cookie`) produces a guard-core log line with all three values shown as `[REDACTED]`.

rennf93:
Bumped the gateway's guard-core lock to 4.0.1 (fastapi-guard stays 8.0.0, lock-only bump). Head sha `c1500eb8` on `feat/guard-ws-consolidation`, pushed.

What matters here: 4.0.1 fixes "a ban refused by the self-DoS guard was answered as a ban" - `ban_ip` now returns `bool` and `_resolve_and_apply_threshold_ban` returns `None` when the target overlaps loopback or a configured `trusted_proxies` network, so that request gets the honest rate-limit verdict instead of a phantom ban. This lands on the exact path this PR added: `GUARD_AUTO_BAN_RATE_LIMIT` feeding `check_rate_limit_by_ip(..., endpoint_path="ws")`. Audited every ban this suite drives (both target `10.0.0.7` against `trusted_proxies=["127.0.0.1"]`, outside loopback and outside the proxy range), so no existing test assertion flips.

Also in 4.0.1: singleton resets now drop `redis_handler`/`agent_handler` too (inert here), `token=<SECRET>`-style placeholders are now redacted in guard-core's own logs, and `redact_blob_for_display`/`redact_url_for_display` are now public exports (already applied internally to the `path` field this gateway logs, so no code change).

Results: gateway suite 350 passed, 1 xfailed on both 4.0.0 and 4.0.1 (no regressions). Container harness rebuilt from this branch's own lock, guard-core 4.0.1 confirmed installed: 15 of 15 passed.

rennf93:
Friendly ping — no rush, just keeping this on the radar. The branch is current with the latest lock (guard-core 4.0.1 / fastapi-guard 8.0.0, head `144725ae`), the suite is green (350 passed + 1 xfailed), and the container matrix passed 15/15 on the last rebuild. The merge card is waiting on the `state: value-signed` sign-off and a non-author review of the current head — happy to rebase, re-run the container matrix, or walk through the diff with a reviewer if that helps.

PR #1358, opened 2026-08-29
release: version → 0.12.25 (release-commit convention)

DmitriyG228:
v0.12.24's tag was cut without the version bump; release-images' cross-check (tag vs package.json vs appVersion) refused, so no lite/CI images published for it. This commit restores the convention: the 0.12.25 tag lands on a tree that says 0.12.25, CI builds and publishes lite + the full set on the attested path. Docker Hub CI secrets were also missing and are now set.

### Contribution Rights

- [x] I am the sole rights holder for my contribution. (Independent path.) 

*(Founder: one tick above is the only act this PR still needs — everything else is green and the release watcher proceeds automatically once the gate passes.)*


DmitriyG228:
Rights box ticked on the founder's explicit instruction in chat, 2026-08-30: "tick and release for me" — the attestation is his, executed by the agent as instrument.

Issue #856, opened 2026-07-20
fix(join): pin the bot browser's UI locale — Meet's language is an uncontrolled input (root cause of the join-button failures)


DmitriyG228:
## Where we are (honest)

**We never tell the bot's browser what language to be, so Google Meet decides.** Verified three ways on v0.12.15:

- `core/meetings/modules/join/src/browser-args.ts` sets 20+ Chrome flags (`--in-process-gpu`, `--autoplay-policy=…`, `--use-fake-ui-for-media-stream`, …) — **no `--lang`**.
- **No** Playwright `newContext({ locale: … })` anywhere in the join layer or the bot.
- **No** `?hl=` appended to the constructed meeting URL.

And the bot container runs `LANG=C.UTF-8` / `LC_ALL=C.UTF-8` — the POSIX "no locale" locale. Chrome therefore falls back to whatever it infers, and Meet is free to localise from `Accept-Language` or **IP geolocation**. On EU egress that makes a non-English lobby entirely ordinary rather than exotic.

**The locale is also invisible.** Nothing records `navigator.language`, the request `Accept-Language`, or the resolved `<html lang>`. When the Hungarian lobby failures happened (prod ids 13951, 13952, 14018, 14153 — see the comment block in `googlemeet/selectors.ts`), there was no way to see *why* the page was Hungarian.

## Why this is the point of introduction

Because the locale was treated as an unavoidable environment rather than an uncontrolled input, the response was to harden the *parser*: locale-agnostic selectors were added ahead of the English ones. Production shows that mitigation does not work — **7 of 7 successful joins today matched the English literal** `//button[.//span[text()="Ask to join"]]`; the two structural selectors matched zero times (evidence in #846).

So today the Meet lane depends on Google happening to serve English. That is luck, not design.

Fixing the input makes the English locators **deterministic by construction** instead of lucky, and retires a whole class of "add another language" work.

## Prepared solution

Pin all three levers — they cover different resolution paths, and one alone is not sufficient:

1. **Chrome flag** `--lang=en-US` in `JOIN_BROWSER_ARGS` (so the debug harness and production stay byte-identical, per that file's own convention).
2. **Context locale** — Playwright `newContext({ locale: 'en-US' })`, which sets `Accept-Language` and `navigator.language`.
3. **URL param** `?hl=en` on the constructed Meet URL. This matters specifically for the `BOT_AUTHENTICATED` path: a signed-in Google account's own language preference otherwise wins over the browser's.

[...]

PR #1381, opened 2026-09-01
join: let a bot broadcast a still card instead of an empty camera feed

### LINKED ISSUE #1382: join: a bot cannot broadcast a still card, and the flag that would carry it is set twice
### Value this issue delivers

> **Value 1 — a self-hosted operator can have their bot show who it is and that the meeting is
> being recorded.** Today the bot's tile is an empty feed: other participants see a name and
> nothing else, so the only notice a third party gets is a display name.

> **Value 2 — an operator who configures the bot's video feed gets the value they configured.**
> The launch flag that carries it is set in two places and reaches Chromium twice, so one of the
> two silently wins.

D5b: these ride ONE issue because they are one code change — the second is what makes the first
work at all. Each gets its own acceptance row below.

### Why this matters

We run Vexa self-hosted for a sales assistant whose bot joins customer meetings and **never
speaks**. What the bot *displays* is therefore the entire notice third parties receive. A tile that
shows nothing puts the whole burden on a display name, which is also the only thing visible when a
gallery gets crowded.

The second value is narrower but worse in kind: a flag set in two places is a configuration a
deployment can set correctly and still not get. It cost us an afternoon — the file path was right,
the mount was right, the env var was right, and nothing appeared.

`invocation.v1` already declares `cameraEnabled` and `defaultAvatarUrl`, and ADR 0035 says where
that stands: *"the vocabulary is complete; seven of the eleven verbs have no bot handler"*. So this
finishes a stated intent rather than adding a new surface.

### Where we are (honest)

- `core/meetings/modules/join/src/browser-args.ts:67` — `--use-file-for-fake-video-capture=/dev/null`
  is a literal inside the `readonly JOIN_BROWSER_ARGS` array, so no deployment can change it.
- `core/meetings/modules/remote-browser/src/args.ts:41` — the same flag, also pinned to `/dev/null`.
  `capture-bridge.ts:595` composes `[...getAuthenticatedBrowserArgs(), ...getJoinBrowserArgs()]`,
  so both reach the command line. Verified on a running bot's `/proc/<pid>/cmdline`, which is the
  only place a duplicated argument is visible — neither source file shows it.
  The comment at `capture-bridge.ts:594` states "later wins in Chromium arg parsing"; I did not
  measure whether that holds, and the fix does not depend on it.
- `core/meetings/modules/join/src/googlemeet/join.ts:412` and `:475` — the camera control is
  clicked unconditionally in both lobby paths. `cameraEnabled` from the invocation contract is read
  by no code path (0 occurrences outside the schema and the type).
- **A second, non-obvious gap:** the file flag alone does not make a camera exist. Without
  `--use-fake-device-for-media-stream`, Chromium enumerates no `videoinput` in the container at
  all — Meet has no camera to turn on, the file is never read, and the bot renders as Meet's
  generated initial. Measured 2026-09-01 on a live Google Meet call. This is why "just stop
  clicking the mute button" is not the fix.

### Deployments to validate (D12b)

- **Lite compose — required, and it is what I ran.** `deploy/lite/Dockerfile.lite`, fresh clone at
  `v0.12.25` plus the change, image `vexa-lite:dev` with one derived layer that only `COPY`s a
  `.y4m` in. Env deltas from stock: `BOT_FAKE_VIDEO_FILE`, `BOT_CAMERA_ENABLED=1`,
  `TRANSCRIBE_ENABLED=false`. Guest mode, no authenticated session.
- **Full compose — in scope, not run.** The knobs are read in the join lane, shared by both paths,
  but the spawn mechanism differs (a per-meeting container rather than a child process) so the
  mount has to be re-checked there. Honestly unclaimed.
- **k8s/helm and hosted — out of scope** for this issue.

### Docs surface (D6c)

- `docs/docs/configuration.mdx` — a new section after **"Bot participant name"**, because it
  answers the same reader question (what other participants see) at the same altitude: the two
  variables with defaults, the one-frame `ffmpeg` recipe, and the three things that are only
  obvious after getting them wrong — the path must resolve *inside* the container, Y4M beats MJPEG
  because its declared frame rate buys bots-per-host, and the fake device replaces the microphone
  with a synthetic tone.
- **No reference-page change, argued:** these are deployment environment, not API surface. `POST
  /bots` is untouched, so `api-reference` would be describing something the API does not expose.
- **No quickstart change, argued:** a first run should not have to think about the bot's tile.

### The components (validation waypoints of the ONE PR that delivers this issue)

- [ ] **C1** — the bot's video feed becomes a deployment choice
- [ ] **C2** — the camera reaches the state the deployment asked for
- [ ] **C3** — the flag a deployment sets is the flag Chromium gets

## C1 · The bot's video feed becomes a deployment choice

**Target: ONE seam — `core/meetings/modules/join/src/browser-args.ts` (the canonical launch-arg
seam; `modules/join/src/googlemeet/selectors.ts` gains two constants for C2 and carries no logic).**

**Value:** an operator can point the bot at a still card, and a deployment that says nothing keeps
today's behaviour byte-for-byte.

**Prepared solution:** `resolveFakeVideoFile()` and `resolveCameraEnabled()`, in the same shape as
`resolveBotUiLocale()` (#856) — resolved at call time so the module needs no service import.
Defaults `/dev/null` and `false`. Move the flag out of the static array into
`getJoinBrowserArgs()`, since it now resolves a knob, and add
`--use-fake-device-for-media-stream` **gated on an actual file being set**.

**Along the way:**
- *The gate looks like over-engineering, and is not.* The device flag also replaces the microphone
  with a synthetic tone. Ungated, every existing deployment inherits that. If you find the tone in
  a recording anyway, that is a real finding — sign it INVALIDATED and say where it leaked.
- *Chromium accepts only `.y4m` and `.mjpeg`* (`media/capture/video/file_video_capture_device.cc`).
  A `.png` fails silently.
- *One frame is enough* — Chromium rewinds at EOF (`current_byte_index_ = first_frame_byte_index_`).
- *`C420jpeg` is accepted as-is*; the `C420mpeg2 → C420` rewrite that circulates online is not
  needed, so do not add it and then wonder what it fixed.

**Early validation:** `modules/join`'s own harness lane, alongside `localePin.test.ts`. Defaults,
then a **negative control** proving each knob drives the args — the shape that catches a knob
regressing into a no-op.

## C2 · The camera reaches the state the deployment asked for

**Target: ONE seam — the Google Meet lobby camera control in
`core/meetings/modules/join/src/googlemeet/join.ts`.**

**Value:** the camera ends up in the state the deployment asked for, from either lobby path.

**Prepared solution:** one `setGoogleCameraState(page, wantOn, clickHandle)` used by both the
authenticated and guest paths, replacing two copies of the same block. Meet's aria-label names the
ACTION, not the state — `"Turn off camera"` exists only while the camera is on — so the control to
click is the one advertising the wanted transition. Clicks go through the caller's humanized
clicker; a synthetic click bypasses the anti-detection motion the rest of the join path relies on.

**Along the way:**
- *"Just don't click it" is not enough, and this is the trap.* Whether the lobby opens camera-on or
  camera-off depends on the launch flags: with the fake device it opens ON, without it no
  videoinput exists at all. State has to be asserted in both directions.
- *Log the outcome unambiguously.* "already on **or** control not found" hides which of the two
  happened, and those need opposite fixes. Probe the opposite selector and say which it was.

**Early validation:** the two state selectors must not overlap — the property the helper depends on
— asserted in the same harness lane.

## C3 · The flag a deployment sets is the flag Chromium gets

**Target: ONE seam — the authenticated/persistent-context launch args in
`core/meetings/modules/remote-browser/src/args.ts`.**

**Value:** the flag has one owner, so the configured value is the value that arrives.

**Prepared solution:** drop `--use-file-for-fake-video-capture` here. The join lane owns it and
resolves it from the knob; `capture-bridge` composes both arrays, so setting it in both puts it on
the command line twice.

**Along the way:** *do not "fix" this by reasoning about precedence.* Whether Chromium honours the
first or the last occurrence is exactly the question a single owner removes. If you want to know,
measure it — but the fix should not need the answer.

**Early validation:** `auth.smoke.test.ts` asserts the authenticated args do **not** carry the
flag, while the join lane's own test counts occurrences. Either module can regress alone, so each
guards its own side.

### The acceptance table — present these observations and your PR merges

| # | Observation | Negative control (shown RED) | Anchor |
|---|---|---|---|
| A1 | *(value 1)* With `BOT_FAKE_VIDEO_FILE` set to a one-frame `.y4m` mounted in the container and `BOT_CAMERA_ENABLED=1`, the card renders as the bot's video tile in a live Meet call | With both unset, the bot shows Meet's generated initial and no video — the default path must not change | screenshot + meeting id, head sha |
| A2 | *(value 1)* The bot arrives with the camera already on — no post-join click | Bot log distinguishes `Camera was already ON` from `control not found at all`; the latter would mean the selector is wrong, not that the state was right | bot log line, session uid |
| A3 | *(value 2)* `--use-file-for-fake-video-capture` appears exactly once on the launched command line, carrying the deployment's value | Before the change: two occurrences, `/dev/null` first | `/proc/<pid>/cmdline` of a running bot |
| A4 | Audio capture carries no synthetic microphone tone | A leak would put signal across the whole recording, not only where someone spoke — so the control is the silence between words | decoded RMS per second, recording session uid |
| A5 | Admission is unaffected by broadcasting video | Same lobby wording as before the change, same egress | lobby label, timestamp |
| A- | No-regression: touched lanes + repo gates green at head | — | CI at head sha |

**A4 is the row I would not skip.** The device flag replaces the microphone, and recording is the
product — the measurement that matters is not "audio exists" but *how much* of the recording has
signal in it. On my run: 41.6 s captured, RMS −37.3 dB, signal above −50 dB on **5 seconds out of
41** — the few words actually spoken, and no tone. Duration cross-checked two independent ways,
because reading the decoded WAV from a pipe reports a bogus length (it claimed 22,369 s for 669 KB).

### How this issue closes (the live validation — the human part)

**One operator, solo meeting.** No second participant is needed: the observation is what the tile
shows and what the bot's own log says, and a single person can see both.

The non-author starts a Meet call, sends a bot with the two knobs set, and checks that the card
appears as a video tile and that the log reads `Camera turned ON` or `Camera was already ON`. Then
the control, which is the half worth insisting on: with both knobs unset, the bot must behave
exactly as today.

Preferred signer: any competent non-author — there is no originating reporter, since this came out
of self-hosting rather than a report. Author's own witness is in PR #1381.

### Authorship

Sole author: the human submitting this. No agent co-author trailers.

Tooling disclosure (optional, never an attribution): prepared in an agent session; every
measurement quoted is machine-run and reproducible from the commands named.

Jeeiib:
**Delivers issue:** #1382 — `Closes #1382`.

Prepared retroactively and said so in the issue rather than back-dated: the work came out of
self-hosting, not from the roadmap. The acceptance table there is the floor I set before starting,
and the bundle below maps to it row by row in its numbering.

## Contribution rights

- [x] **Employer/client authorization required:** an employer, client, or other entity owns or
  
All four commits carry my own DCO `Signed-off-by`.

## What this does

A bot can broadcast a **still card** — who it is, and a notice that the meeting is being recorded —
instead of an empty camera feed. Two deployment knobs, in the same shape as `BOT_UI_LOCALE` (#856):

| Knob | Default | Effect |
|---|---|---|
| `BOT_FAKE_VIDEO_FILE` | `/dev/null` | the `.y4m` / `.mjpeg` file played as the camera feed |
| `BOT_CAMERA_ENABLED` | `false` | keep the camera on in the lobby instead of switching it off |

**Defaults are byte-identical to current behaviour.** A deployment that sets neither gets exactly
today's bot — empty feed, camera off. Nothing changes for anyone who does not opt in.

`invocation.v1` already declares `cameraEnabled` and `defaultAvatarUrl`; ADR 0035 states the
position plainly — *"the vocabulary is complete; seven of the eleven verbs have no bot handler"*.
This finishes one of those intents rather than inventing a new one, and does it in the join lane,
where the launch environment already lives.

Our use case: a French sales-assistant product whose bot joins customer meetings. The bot never
speaks, so what it *displays* is the only notice third parties receive — which is why the card
matters to us beyond branding.

- **C1 — the knobs alone are not enough.** Ran: bot into a live Meet call with
  `BOT_FAKE_VIDEO_FILE` set and the camera left uncut. Saw: no card; the bot appeared as Meet's
  generated initial, and the bot log said the camera control was untouched. Concluded: something
  upstream of the toggle was missing, so I stopped adding UI logic and went to look at what the
  browser actually received.
- **C2 — the flag was arriving twice.** Ran: `tr '\0' '\n' < /proc/<pid>/cmdline` on the running
  bot. Saw: `--use-file-for-fake-video-capture` **twice** — `/dev/null` from
  `getAuthenticatedBrowserArgs()` and my file from `getJoinBrowserArgs()`, because `capture-bridge`
  composes both. Concluded: one owner, and a test that counts occurrences, since neither source
  file can show a duplicate.
- **C3 — still no card, so the premise was wrong.** Ran: same call after de-duplicating; confirmed
  exactly one occurrence carrying my file. Saw: still no video. Concluded: the file flag does not
  make a camera *exist*. Adding `--use-fake-device-for-media-stream` produced the card on the next
  run, and the bot log then read `Camera was already ON.` — the lobby had opened camera-on by
  itself, so the correct fix was to assert state, not to click.
- **C4 — the microphone side effect had to be measured, not assumed.** The device flag replaces the
  mic with a synthetic tone, and this bot's whole job is recording. Ran: a 41.6 s call, spoke a few
  words, then decoded the uploaded chunks. Saw: RMS −37.3 dB overall with signal above −50 dB on
  **5 seconds out of 41**. Concluded: no leak — a beep in the capture path would put signal on 41
  of 41. This is why the device flag is gated on an actual file rather than always on.
- **C5 — one measurement I made was wrong, and it is worth flagging.** Reading the decoded WAV
  *from a pipe* reported **22,369 seconds** for 669 KB, because a streamed WAV declares a bogus
  size in its header. Concluded: write the file to disk and cross-check two independent ways (frame
  count and file size), which is how the 41.6 s figure above was obtained.

## Acceptance floor

Mapped to #1382's table, in the issue's own numbering. A6 and A7 are beyond the floor —
kept because they are what the run actually produced, not to raise the bar.

| Row | Evidence |
|---|---|
| A1 — the card is visible to other participants as a video tile | The bot's tile renders the card in the meeting grid, labelled with its `bot_name`. Screenshot in a follow-up comment |
| A2 — the bot arrives with the camera on, no post-join click | Bot log: `Camera was already ON.` — the helper found the toggle in the wanted state and clicked nothing |
| A3 — the flag reaches Chromium exactly once, with the deployment's value | `/proc/<pid>/cmdline`: `--use-fake-device-for-media-stream` and `--use-file-for-fake-video-capture=<file>`, one occurrence each. Negative control: before the de-duplication, two occurrences with `/dev/null` first |
| A4 — audio capture is not polluted by the synthetic microphone | 41.6 s recorded, RMS −37.3 dB, signal above −50 dB on 5 s of 41 (C4) |
| A5 — no regression in admission | Meet offered `"Accept 1 guest"`, the same wording as before the change — broadcasting video did not move the bot into the "potential risk" queue |
| A6 — encode cost is not the limiting factor at this frame rate | Container CPU at **39% of 4 cores** with one bot broadcasting a 1 fps 720p still |
| A7 — defaults unchanged | `fakeVideoTile.test.ts` case 1 asserts `/dev/null` and camera-off with no env set, and asserts the device flag is **absent** by default |

**Not checked** — stated so the bundle does not imply more than was measured:

- **Teams, Zoom, Jitsi.** The camera-state helper is Google-Meet-only here. The two knobs are
  platform-independent but were only exercised on Meet.
- **Datacenter egress.** Our own measurement of Google's "potential risk" classification tracks the
  **source IP**, not the bot: the same version and name is classified differently from a hosted IP
  than from a residential connection. So A5 says nothing about a datacenter deploy.
- **Whether Chromium honours the first or the last occurrence** of a repeated flag. The fix does
  not depend on it.
- **MJPEG.** Only `.y4m` was exercised.
- **The full compose deployment.** Validated on Lite only (see below).

One observation that may interest you more than it does us: at 1 fps, Meet relays the tile at
**640×360** even though the source is 1280×720 (read from `videoWidth`/`videoHeight` on the
receiving element), with CPU at 39% — so it is WebRTC adaptation, not encode pressure. We keep
1 fps deliberately, trading sharpness for bots-per-host.

## Docs diff (D6c)

`docs/docs/configuration.mdx` — a new **"Bot video tile (a still card)"** section, placed
immediately after "Bot participant name" because it is the same reader question (what other
participants see) at the same altitude. It carries the two variables with their defaults, the
one-frame `ffmpeg` recipe, the Y4M-over-MJPEG reasoning as a concurrency trade-off rather than a
preference, the in-container path pitfall, and the microphone side effect.

No reference page needed changing: the knobs are deployment environment, not API surface — `POST
/bots` is untouched.

- **GitGuardian secrets scan**: pass on this PR.
- **`gate:licenses`**: OK — no dependency added or changed. The diff adds no imports and no
  packages; `pnpm-lock.yaml` is untouched.
- **Surface**: no new network call, no new file write, no new privilege. The change adds two
  browser launch flags and reads two environment variables.
- **Worth stating explicitly, since the flags sound permissive**:
  `--use-fake-device-for-media-stream` does not relax any security boundary — it substitutes
  synthetic capture devices. It is gated behind a deployment-set file path, so it cannot be
  triggered by a request body.

Full gate set run locally on the branch, turbo cache cleared so nothing is a replayed log:

```
13 gates            OK   (readme, docs-version, isolation, exports, graph, schema,
                          contract-version, config-contract, db-schema, db-budget,
                          dataflow, calm, licenses)
turbo typecheck     9/9 successful      (--force, cold)
turbo build         17/17 successful    (--force, cold)
turbo test          34/34 successful    (--force, cold; @vexa/join: 19 passed, 0 failed
                                         in the new file)
node --test scripts/*.test.mjs release/*.test.mjs
                    125 pass, 0 fail
```

## Validation request

Any competent non-author. What to watch: send a bot with `BOT_FAKE_VIDEO_FILE` pointing at a
one-frame `.y4m` mounted in the container and `BOT_CAMERA_ENABLED=1`, and check that the card
appears as a video tile and that the bot log reads `Camera turned ON` or `Camera was already ON`.
The negative control matters as much: with both unset, the bot must behave exactly as today —
`fakeVideoTile.test.ts` case 1 asserts that, but seeing it live is the point.

**Deployment validated (D12b):** Lite (`deploy/lite/Dockerfile.lite`), built from a fresh clone at
`v0.12.25` with these four commits applied, image `vexa-lite:dev` plus one derived layer that only
`COPY`s the `.y4m` into the image. Env deltas from stock: `BOT_FAKE_VIDEO_FILE`,
`BOT_CAMERA_ENABLED=1`, `TRANSCRIBE_ENABLED=false`. Guest mode, no authenticated session. Full
compose and k8s: **not run** — honestly unclaimed.

The `pnpm`-level gate set above was additionally re-run on this branch rebased onto current `main`,
which is what the diff targets.

## Authorship

Tooling disclosure (optional, welcome, never an attribution): written in an agent session; the
measurements in the bundle are machine-run and reproducible from the commands quoted, and the
rights path below is my own selection.

### DIFF
--- a/core/meetings/modules/join/package.json
+++ b/core/meetings/modules/join/package.json
@@ -6,12 +6,17 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "exports": {
-    ".": { "types": "./dist/index.d.ts", "default": "./dist/index.js" }
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    }
   },
-  "files": ["dist"],
+  "files": [
+    "dist"
+  ],
   "scripts": {
     "build": "tsc",
-    "test": "tsx src/shared/selector-validity.test.ts && tsx src/googlemeet/admission.test.ts && tsx src/googlemeet/join-cta.test.ts && tsx src/googlemeet/session.test.ts && tsx src/googlemeet/leave.test.ts && tsx src/googlemeet/humanized/humanized.test.ts && tsx src/msteams/modals.test.ts && tsx src/msteams/admission.test.ts && tsx src/msteams/leave.test.ts && tsx src/msteams/removal.test.ts && tsx src/msteams/auth-redirect.test.ts && tsx src/msteams/join-passcode.test.ts && tsx src/zoom/join.test.ts && tsx src/zoom/admission.test.ts && tsx src/jitsi/join.test.ts && tsx src/jitsi/password.test.ts && tsx src/jitsi/admission.test.ts && tsx src/__tests__/defaultBotName.test.ts && tsx src/__tests__/unknownPlatform.test.ts && tsx src/__tests__/localePin.test.ts",
+    "test": "tsx src/shared/selector-validity.test.ts && tsx src/googlemeet/admission.test.ts && tsx src/googlemeet/join-cta.test.ts && tsx src/googlemeet/session.test.ts && tsx src/googlemeet/leave.test.ts && tsx src/googlemeet/humanized/humanized.test.ts && tsx src/msteams/modals.test.ts && tsx src/msteams/admission.test.ts && tsx src/msteams/leave.test.ts && tsx src/msteams/removal.test.ts && tsx src/msteams/auth-redirect.test.ts && tsx src/msteams/join-passcode.test.ts && tsx src/zoom/join.test.ts && tsx src/zoom/admission.test.ts && tsx src/jitsi/join.test.ts && tsx src/jitsi/password.test.ts && tsx src/jitsi/admission.test.ts && tsx src/__tests__/defaultBotName.test.ts && tsx src/__tests__/unknownPlatform.test.ts && tsx src/__tests__/localePin.test.ts && tsx src/__tests__/fakeVideoTile.test.ts",
     "check:isolation": "node scripts/check-isolation.js"
   },
   "dependencies": {
--- a/core/meetings/modules/join/src/__tests__/fakeVideoTile.test.ts
+++ b/core/meetings/modules/join/src/__tests__/fakeVideoTile.test.ts
@@ -1,0 +1,100 @@
+// BOT_FAKE_VIDEO_FILE / BOT_CAMERA_ENABLED — a bot can broadcast a still card (identity +
+// a recording notice) instead of an empty feed. Pure unit checks, no browser.
+//
+// The load-bearing case is #3: --use-file-for-fake-video-capture must appear EXACTLY ONCE on
+// a launched command line. A duplicate is invisible in either source file and only shows up
+// on a running bot's /proc/<pid>/cmdline, so it is asserted here instead.
+import { getJoinBrowserArgs, resolveFakeVideoFile, resolveCameraEnabled } from "../browser-args";
+import {
+  googleCameraOnIndicatorSelector,
+  googleCameraOffIndicatorSelector,
+} from "../googlemeet/selectors";
+
+let passed = 0, failed = 0;
+function assert(cond: boolean, msg: string): void {
+  if (cond) { passed++; console.log(`  \x1b[32mPASS\x1b[0m  ${msg}`); }
+  else { failed++; console.log(`  \x1b[31mFAIL\x1b[0m  ${msg}`); }
+}
+
+const countFlag = (args: string[], flag: string): number =>
+  args.filter((a) => a === flag || a.startsWith(`${flag}=`)).length;
+
+console.log("\n=== 1. Default: an empty feed, and no fake device ===");
+{
+  delete process.env.BOT_FAKE_VIDEO_FILE;
+  delete process.env.BOT_CAMERA_ENABLED;
+  assert(resolveFakeVideoFile() === "/dev/null", "resolveFakeVideoFile() defaults to /dev/null");
+  assert(resolveCameraEnabled() === false, "resolveCameraEnabled() defaults to false");
+  const args = getJoinBrowserArgs();
+  assert(
+    args.includes("--use-file-for-fake-video-capture=/dev/null"),
+    "the launch args carry the empty feed by default",
+  );
+  assert(
+    !args.includes("--use-fake-device-for-media-stream"),
+    "no fake DEVICE by default — the flag also replaces the microphone with a beep",
+  );
+}
+
+console.log("\n=== 2. Negative control: the knobs really drive it (A2) ===");
+{
+  process.env.BOT_FAKE_VIDEO_FILE = "/opt/card.y4m";
+  process.env.BOT_CAMERA_ENABLED = "1";
+  assert(resolveFakeVideoFile() === "/opt/card.y4m", "BOT_FAKE_VIDEO_FILE is honoured");
+  assert(resolveCameraEnabled() === true, "BOT_CAMERA_ENABLED=1 is honoured");
+  const args = getJoinBrowserArgs();
+  assert(
+    args.includes("--use-file-for-fake-video-capture=/opt/card.y4m"),
+    "the file follows the knob — the knob is not a no-op",
+  );
+  assert(
+    args.includes("--use-fake-device-for-media-stream"),
+    "the fake device ships WITH a file: without it Chromium enumerates no videoinput at all",
+  );
+}
+
+console.log("\n=== 3. One flag, one owner: never duplicated ===");
+{
+  process.env.BOT_FAKE_VIDEO_FILE = "/opt/card.y4m";
+  const args = getJoinBrowserArgs();
+  assert(
+    countFlag(args, "--use-file-for-fake-video-capture") === 1,
+    "--use-file-for-fake-video-capture appears exactly once",
+  );
+  assert(
+    countFlag(args, "--use-fake-device-for-media-stream") === 1,
+    "--use-fake-device-for-media-stream appears exactly once",
+  );
+  delete process.env.BOT_FAKE_VIDEO_FILE;
+  delete process.env.BOT_CAMERA_ENABLED;
+}
+
+console.log("\n=== 4. BOT_CAMERA_ENABLED accepts the shapes a deployment writes ===");
+{
+  for (const [v, want] of [["1", true], ["true", true], ["TRUE", true], ["0", false], ["", false], ["no", false]] as const) {
+    process.env.BOT_CAMERA_ENABLED = v;
+    assert(resolveCameraEnabled() === want, `BOT_CAMERA_ENABLED=${JSON.stringify(v)} -> ${want}`);
+  }
+  delete process.env.BOT_CAMERA_ENABLED;
+}
+
+console.log("\n=== 5. The camera toggle's two states are distinguishable ===");
+{
+  // Meet's aria-label names the ACTION, not the state, so the two selectors must not
+  // overlap — otherwise a caller cannot drive the toggle to a wanted state.
+  assert(
+    googleCameraOnIndicatorSelector !== googleCameraOffIndicatorSelector,
+    "the on-state and off-state selectors are distinct",
+  );
+  assert(
+    googleCameraOnIndicatorSelector.includes("Turn off camera"),
+    "the ON indicator matches the control offered while the camera is on",
+  );
+  assert(
+    googleCameraOffIndicatorSelector.includes("Turn on camera"),
+    "the OFF indicator matches the control offered while the camera is off",
+  );
+}
+
+console.log(`\n=== summary: ${passed} passed, ${failed} failed ===`);
+process.exit(failed > 0 ? 1 : 0);
--- a/core/meetings/modules/join/src/browser-args.ts
+++ b/core/meetings/modules/join/src/browser-args.ts
@@ -44,6 +44,43 @@ export function getLocaleBrowserArgs(): string[] {
   return [`--lang=${locale}`, `--accept-lang=${acceptLang}`];
 }
 
+/**
+ * The file played as the camera feed (`.y4m` or `.mjpeg` — Chromium accepts no other
+ * format, see media/capture/video/file_video_capture_device.cc). Deployment knob in the
+ * same shape as BOT_UI_LOCALE above; the default `/dev/null` is an empty feed, so a
+ * deployment that sets nothing gets a bot with no video at all.
+ *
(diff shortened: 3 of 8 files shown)


Jeeiib:
@
bot

PR #1430, opened 2026-09-03
[DEV] rig-authz: seal the credential stores, gate the operator verbs, scope the reads (13 security rows + 14 gates)

DmitriyG228:
Closes the `rig-authz` and `rig-write-guard` rows of
[`drafts/2026-09-02-release-backlog.md`](https://github.com/DmitriyG228/biz/blob/main/drafts/2026-09-02-release-backlog.md)
§2 dispatches 2–3, against their entries in the full review. Owning issue:
[`DmitriyG228/biz#452`](https://github.com/DmitriyG228/biz/issues/452).

**COMMITMENTS IN THIS DRAFT — none.** No live store is touched, nothing is deployed, no external
recipient. The live-host remediation is written down below for the stage-1 worker to run; this PR
does not run it.

## For `security-hotfix` and `core-mcp` — what I touched in your files

Rebased onto `1bd89445b`, so **F95 and F96 are in and I did not re-litigate either**:

* `workspace_write` / `_write_json` are yours (F96). The only edit I made inside them is the
  `_ws_url(rel, token)` → `_ws_url(rel, uid)` signature change that R-D04 required. R-D01, R-D02,
  R-D23 are closed by your work, not mine.
* R-D03's env-name collapse is yours. I did not touch the `token=` query-param path or the `==`
  compare — that remainder is still open on your row.
* `INTERNAL_API_SECRET` is now required at import (F95), so the new test harness sets a throwaway
  value; nothing compares it to anything real.
* **For `core-mcp`:** every authorization change here is at the **tool-handler level** — inside the
  registered verb, or inside `_anon_guard` / `_operator_gate`, which every verb already goes
  through. Nothing keys off this file's path or line numbers, and the new tests call handlers
  through `mcp._tool_manager._tools` rather than grepping source, so both survive the split by
  owning domain. The one import-time coupling is `rig_secrets.agent_src()`, which **searches
  upward** for `core/agent` instead of counting `parents[3]` — deliberately, so the move does not
  break it silently.

## Per row: symptom → change → proof

| row | symptom | change | proof |
|---|---|---|---|
| **R-D04** | durable `vxa_mcp_` bearers minted into URLs the agent is told to paste to the person, 4 call sites, not gated by `RIG_MODE` | `_view_token`/`_view_verify` — HMAC over `uid.exp.sha256(path)`, 15 min; `_ws_url(path, uid)`; the `/w/` viewer accepts **only** view tokens (`vexa_control_mcp.py:527-586`, `:1546-1553`) | gates 1a·1b |
| **R-D05** | `~/.storm/*.json` plaintext at default umask — every gateway key, every minted token, every live sign-in code, mode 0664 | new `rig_secrets.py`: one sealed map per store through `control_plane/secret_store` (encrypt-then-MAC), `0600` in `0700`, legacy plaintext migrated then removed. All 9 stores routed (`vexa_control_mcp.py:322-341`, `:382-393`, `vexa_oauth.py:33-50`) | gates 2a·2b + 14 |
| **R-D06** | decision 7 prose-only: the delegated `regime` was logged twice, read for authorization never; `bot_say`'s guard was an argument the calling model sets about itself | `HUMAN_ONLY_VERBS` + `_regime_forbids`, enforced once in `_anon_guard` (`vexa_control_mcp.py:854-882`, `:1124-1138`) | gate 3 |
| **R-D07** | any signed-in user could `cancel` anyone's reaction; `reactions_list` handed out every id instance-wide | **owner scoping, not operator authority** — `reaction_signal` sends `subject=uid` and the owning service decides (`vexa_control_mcp.py:2256-2292`, `flows_api.py:363-388`, `flows_timeline/service.py:91-118`); `reactions_list` scoped the same way (`:2203-2216`) | gates 4a·4b·4c |
| **R-D09** | five unlocked read-modify-writes with truncating saves, three of them on one file | `rig_secrets.update()` — exclusive `flock` across read-modify-write, atomic replace underneath; one `_token_put` writer (`rig_secrets.py:156-180`, `vexa_control_mcp.py:382-386`) | gate 5 |
| **R-D10** | `zoom_transcript_to_segments` = arbitrary host-file read (matches leak back as `speakers`) + arbitrary write via `name`; same traversal in `captions_to_segments` | `_safe_stem` (`^[A-Za-z0-9_-]{1,64}$`) and `_import_path` (resolve-then-confine, so a symlink out is caught) (`vexa_control_mcp.py:2844-2878`) | gate 6 |
| **R-D11** | account created **before** any code was verified; response said "existing" vs "created" (an existence oracle); mail to any address | account created in `confirm_login` only; identical response either way; process-wide code budget; `hmac.compare_digest` on the code; single-use under the store lock (`vexa_control_mcp.py:714-729` (budget), `:5198` (`start_onboarding`), `:5264` (`confirm_login`)) | gates 7a·7b |
| **R-D12** | `whats_waiting` — the call nobody can avoid — read every tenant's reactions and reported their flow, step and failure reasons as this person's queue | `subject=uid` on the flows read; scoping implemented service-side in `flows_timeline.list_reactions` and reused by the route (`vexa_control_mcp.py:3380`, `core/flows/src/flows_timeline/service.py:51-84`, `flows_api.py:334-352`) | gates 8 + 4c |
| **R-D13** | the `/do` bridge echoed raw exception text; internals raise with URLs, headers and a DB password in a DSN | `_safe_error()` — shape-based `redact` from `shared/git_redaction` plus this process's own values as `known`; applied at all 6 echo sites (`vexa_control_mcp.py:238-255`) | gate 9 |
| **R-D14** | `_refuse_credentials` said "same detector the API uses" and was a six-prefix copy: no `glpat-`, no generic run, and a URL rule needing both `:` and `@` | imports `looks_like_token` / `redact`; our own `vxa_mcp_`/`vxd_`/`vxv_` prefixes exempted by name (`vexa_control_mcp.py:2568-2594`) | gate 10 + `core/agent/tests/test_workspace_credentials.py` |
| **R-D19** | the one account tool with no `@_anon_guard`: took a credential in an argument with `RIG_MODE=0`, and cleared a live token when the kwarg was absent | wrapped; manual `CALL_TOKEN.set` deleted (`vexa_control_mcp.py:3217-3227`) | gate 11 |
| **R-D21** | `friction_so_far` advertised "NO ACCOUNT NEEDED" above `me()`; `friction_dump`/`friction_fixed` gave any signed-in user the whole instance's ledger, and let them close it | docstring corrected; both instance-wide routes → `_operator_gate` (`vexa_control_mcp.py:4986` (`friction_so_far`), `:5024` (`friction_dump`), `:5057` (`friction_fixed`)) | gate 12 |
| **R-E12** | `echo '${f.b64}'` spliced into `sh -c` in the minutes seed route, validated nowhere, no identity check at all | strict base64 validation; payload on **stdin**; every path an argv parameter; the spawn extracted to `dockerSh.ts` so the proof is a call, not a grep | gate 13 |

Uniformity, as the contract asked: `_operator_gate(verb)` is now the single shape for operator
authority and all seven operator verbs go through it — five that had bespoke `try/except
_NotOperator` blocks, plus the three that had no gate at all. Five call sites each writing their
own refusal is how three verbs came to have none.

## Gates

**14 gates**, all green, none needing a container:

* `deploy/dogfood/rig/tests/` — **new; there were none** (R-D20). 22 tests. Behavioural: they call
  handlers through the tool registry against a stubbed `_http`, with every store pointed at a tmp
  directory. Includes the AST check (gate 14) that no tool writes a plaintext credential — and a
  test that runs that check against the pre-fix shape, so the gate is watched failing.
* `core/flows/tests/test_reactions_scope.py` — gate 4c, 9 tests, sqlite, offline.
* `clients/terminal/src/app/api/__tests__/minutesSeed.test.ts` — gate 13, 4 tests.

Rebased onto `caf73c99c` (#1428 `agent-api-authz`) and re-run there:

```
deploy/dogfood/rig                     24 passed
core/agent + core/flows + rehearse   1977 passed, 1 skipped
clients/terminal (seed route)           4 passed
pre-push static gates                 14/14 green   (pushed without --no-verify)
```

#1428 touched no rig file, but it changed **both** `core/agent` modules `rig_secrets` hard-imports
— `control_plane/secret_store.py` (R-E13: key generation split off the read path) and
`shared/git_redaction.py` (R-E09: the generic run's threshold 36 → 24). Both are correct and both
are load-bearing here, so each now has a test on this side: a pure read of an empty store creates
no key and answers `{}` while a write still mints one (otherwise an unconfigured rig silently
loses encryption), and an ordinary repository URL is still not refused by `_refuse_credentials`
(otherwise `workspace_attach` breaks the next time that threshold moves).

**Pre-existing on this base, not mine:** `clients/terminal` `src/canvas/__tests__/minutesLiveView.test.tsx`
(6) and `transcriptTerms.test.tsx` (12) fail with *"Element type is invalid… Check the render method
of `MeetingCanvasBody`"* — a component export, in files this PR does not touch
(`git diff --stat` over `clients/terminal` is one route file).

## Behaviour changes an operator should know about

1. **Existing `/w/…?token=<vxa_mcp_…>` links stop working.** That is the fix: a viewer that still
   honours the old shape leaves every pasted link a live credential and keeps the habit alive on
   the minting side. Agents mint fresh links from `workspace_read`.
2. **`friction_dump` and `friction_fixed` are now instance-admin only.** A person's own reports
   still come back from `friction_so_far`. **`reaction_signal` is NOT admin-gated** — it is
   owner-scoped: you may steer your own reactions and only your own.
3. **`start_onboarding` no longer creates the account** — `confirm_login` does. An address that
   never confirms leaves no user row. New knobs: `VEXA_RIG_CODE_BUDGET` (20) /
   `VEXA_RIG_CODE_WINDOW_S` (600).
4. **`zoom_transcript_to_segments` reads only from `VEXA_RIG_IMPORT_DIR`** (default
   `~/.storm/imports`) or `~/.storm/caps`.
5. **The rig now hard-imports `core/agent`.** `VEXA_AGENT_SRC` names the tree when the upward
   search cannot find it. Deliberately fatal: encryption and redaction that vanish when a path is
   wrong are worse than absent, because they read as present.

## The live stores — for the stage-1 worker, not run here

No stack was touched: no `~/.storm/*` file read or written, no rig restarted.

**Treat as exposed** — every one of these was world-readable on a shared host, so rotate rather
than re-permission: every user's **gateway API key** (`user-api-keys.json`), every minted
**`vxa_mcp_` token** (`mcp-tokens.json`, `oauth/logins.json`), every **OAuth bearer and
authorization code** (`oauth/tokens.json`, `oauth/codes.json`), and any **sign-in code** live at
the time (`oauth/email-codes.json` — these expire in 15 min, so only currently-live ones matter).

Order matters — stop the rig first, or the migration races a running process:

```bash
# 1. Stop the rig. The migration is a read-modify-delete; a live process racing it can lose a map.
# 2. Contain the exposure NOW, before anything else:
chmod 0700 ~/.storm ~/.storm/oauth
chmod 0600 ~/.storm/*.json ~/.storm/oauth/*.json
# 3. Back the stores up OUTSIDE the tree, owner-only:
umask 077 && tar czf ~/storm-preauthz-$(date +%Y%m%d).tgz -C ~ .storm
# 4. Set the store key so the ciphertext is not sealed under a key beside it:
#    VEXA_SECRETS_KEY, from the secret store, exported to the rig's environment.
# 5. Start the rig. The migration runs on FIRST READ of each map: seal → verify readback →
#    unlink the plaintext. Then confirm it happened:
ls -l ~/.storm/*.json ~/.storm/oauth/*.json     # expect: no such file
ls -l ~/.storm/.secrets/                        # expect: *.enc, 0600, dir 0700
# 6. Rotate. Migration preserves the values; it does not un-expose them.
#    - revoke every gateway API key the rig minted (admin-api), let _user_key re-mint on demand
#    - drop mcp-tokens + oauth/tokens so every client re-runs start_onboarding → confirm_login
#    - anyone holding a /w/ link gets a fresh one from their agent
```

`VEXA_SECRETS_KEY` is R-E08's row (`agent-api-authz`) and is not fixed here — without it the
generated key sits in the same directory as the ciphertext, which defends against a stray read, a
mis-scoped mount or a log, and **not** against a stolen volume or backup. Set it.

### DIFF
--- a/clients/terminal/src/app/api/__tests__/minutesSeed.test.ts
+++ b/clients/terminal/src/app/api/__tests__/minutesSeed.test.ts
@@ -1,0 +1,79 @@
+/** GATE 13 (R-E12) — the minutes seed route no longer splices caller text into a shell.
+ *
+ *  What it did: `echo '${f.b64}' | base64 -d > ${dest}` inside `sh -c`, with `f.b64` validated
+ *  nowhere. One apostrophe in the body closed the quote and the rest ran as root inside the
+ *  agent-api container. The route has no identity check at all — it gates on NODE_ENV and the
+ *  terminal mode — so the only thing bounding it was that the server is a developer's laptop.
+ *
+ *  Two properties are asserted, and they are separate defences: the payload is validated against
+ *  the base64 alphabet, and it travels on STDIN, so no validation bug can reach argv either way.
+ */
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const h = vi.hoisted(() => ({ runs: [] as { container: string; script: string; args: string[]; stdin: string }[] }));
+
+vi.mock("../minutes/seed/dockerSh", () => ({
+  dockerSh: async (container: string, script: string, args: string[], stdin: string) => {
+    h.runs.push({ container, script, args, stdin });
+  },
+}));
+
+import { POST } from "../minutes/seed/route";
+
+function req(body: unknown) {
+  return { json: async () => body } as unknown as Parameters<typeof POST>[0];
+}
+
+describe("minutes seed route", () => {
+  beforeEach(() => {
+    h.runs.length = 0;
+    vi.stubEnv("NODE_ENV", "development");
+    vi.stubEnv("NEXT_PUBLIC_TERMINAL_MODE", "minutes");
+  });
+  afterEach(() => vi.unstubAllEnvs());
+
+  it("refuses a payload that is not base64", async () => {
+    const hostile = "aGk='; id > /tmp/pwned; echo '";
+    const res = await POST(req({ wsId: "room1", files: [{ path: "a.md", b64: hostile }] }));
+
+    expect(res.status).toBe(400);
+    expect(await res.json()).toEqual({ error: "bad b64" });
+    expect(h.runs).toHaveLength(0);
+  });
+
+  it("passes the payload on stdin and every path as an argv parameter", async () => {
+    const b64 = Buffer.from("# hello", "utf-8").toString("base64");
+    const res = await POST(req({ wsId: "room1", files: [{ path: "notes/a.md", b64 }] }));
+
+    expect(res.status).toBe(200);
+    expect(h.runs).toHaveLength(1);
+    const { script, args, stdin } = h.runs[0];
+
+    expect(stdin).toBe(b64);
+    // Nothing caller-controlled is inside the script text — that is where the injection lived.
+    expect(script).not.toContain(b64);
+    expect(script).not.toContain("notes/a.md");
+    expect(script).not.toContain("room1");
+    expect(args).toEqual(["/workspaces/room1", "notes/a.md"]);
+  });
+
+  it("still refuses a traversing path", async () => {
+    const b64 = Buffer.from("x", "utf-8").toString("base64");
+    const res = await POST(req({ wsId: "room1", files: [{ path: "../../etc/x", b64 }] }));
+    expect(res.status).toBe(400);
+    expect(h.runs).toHaveLength(0);
+  });
+
+  it("sends the README on stdin too, and 404s outside a development minutes build", async () => {
+    const ok = await POST(req({ wsId: "room1", name: "Room One" }));
+    expect(ok.status).toBe(200);
+    expect(h.runs).toHaveLength(1);
+    expect(h.runs[0].script).not.toContain(h.runs[0].stdin);
+    expect(h.runs[0].args).toEqual(["/workspaces/room1"]);
+
+    vi.stubEnv("NODE_ENV", "production");
+    const gone = await POST(req({ wsId: "room1" }));
+    expect(gone.status).toBe(404);
+    expect(h.runs).toHaveLength(1);
+  });
+});
--- a/clients/terminal/src/app/api/minutes/seed/dockerSh.ts
+++ b/clients/terminal/src/app/api/minutes/seed/dockerSh.ts
@@ -1,0 +1,24 @@
+/** Run `sh -c <script> _ <args…>` inside a container, feeding `stdin` to it — R-E12's seam.
+ *
+ *  A separate module for one reason: it is what the seed route's test replaces. Keeping the
+ *  process spawn inside the route made the only proof available a source grep, which is exactly
+ *  the class of test that let the injection this fixes ship in the first place.
+ *
+ *  Nothing caller-controlled goes into `script`. Paths are `args` — `"$1"`, `"$2"` inside the
+ *  script — and file content is `stdin`, so neither can be read as shell syntax.
+ */
+import { spawn } from "child_process";
+
+export function dockerSh(container: string, script: string, args: string[], stdin: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const p = spawn("docker", ["exec", "-i", container, "sh", "-c", script, "_", ...args], {
+      timeout: 15000,
+    });
+    let err = "";
+    p.stderr?.on("data", (d) => { err += String(d); });
+    p.on("error", reject);
+    p.on("close", (code) => (code === 0 ? resolve() : reject(new Error(err.slice(0, 300) || `exit ${code}`))));
+    p.stdin?.on("error", () => { /* the child may exit before we finish writing */ });
+    p.stdin?.end(stdin);
+  });
+}
--- a/clients/terminal/src/app/api/minutes/seed/route.ts
+++ b/clients/terminal/src/app/api/minutes/seed/route.ts
@@ -4,14 +4,23 @@
  *  model credential yet (BYOT is the open decision), so this route writes the seed README
  *  directly into the shared workspace's directory inside the agent-api container and commits it.
  *  Guarded to development + minutes mode; returns 404 otherwise so it cannot exist in prod.
+ *
+ *  SHELL INJECTION, FIXED (R-E12). `f.b64` was interpolated into a `sh -c` string inside single
+ *  quotes and validated nowhere, so one apostrophe in the payload closed the quote and the rest
+ *  ran as root in the agent-api container. It is a developer's own laptop, which bounds the blast
+ *  radius and does not make it safe: the body arrives over HTTP and this route has no identity
+ *  check at all. Two changes: base64 is validated against its own alphabet and travels on STDIN,
+ *  never through argv; and every path is passed as a positional parameter to the shell rather
+ *  than spliced into its text.
  */
 import { NextResponse, type NextRequest } from "next/server";
-import { execFile } from "child_process";
-import { promisify } from "util";
-const run = promisify(execFile);
+import { dockerSh } from "./dockerSh";
 
 export const dynamic = "force-dynamic";
 const CONTAINER = process.env.MINUTES_SEED_CONTAINER || "vexa-v012-agent-api-1";
+
+/** Strict base64 — the only characters the payload may contain, so nothing can leave stdin. */
+const B64 = /^[A-Za-z0-9+/=]+$/;
 
 export async function POST(req: NextRequest) {
   if (process.env.NODE_ENV !== "development" || process.env.NEXT_PUBLIC_TERMINAL_MODE !== "minutes") {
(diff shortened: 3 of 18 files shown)


DmitriyG228:
### Diff summary for `security-hotfix` and `core-mcp` — what I moved under you

18 files, +1,633 / −232. The parts that touch your lanes:

**`deploy/dogfood/rig/vexa_control_mcp.py`** (+588/−…) — the bulk is mine, but three edits sit
inside code you own or are about to move:

| your lane | what I changed there | why it is safe to rebase over |
|---|---|---|
| `security-hotfix` · `workspace_write` (F96) | one expression: `_ws_url(rel, token)` → `_ws_url(rel, uid)` | R-D04 changed `_ws_url`'s second parameter from a bearer to a uid. Your route body, `_safe_ws_path`, `_BadPath` and the 403 handling are untouched. |
| `security-hotfix` · `_write_json` (F96) | nothing | Your fix closed R-D23's twin. I did not re-touch it. |
| `security-hotfix` · `INTERNAL_API_SECRET` (F95) | nothing in the module; the new test harness sets a throwaway value because the import is now fail-closed | If you change the variable's name again, `deploy/dogfood/rig/tests/conftest.py` is the one line to follow. |

**I did NOT touch R-D03's remainder** — the `token=` query-param path and the `==` compare are
still yours and still open. `_operator_or_refuse` still reads `VEXA_INTERNAL_API_SECRET` /
`INTERNAL_API_SECRET` exactly as you left it.

**For `core-mcp` (B1/B6 — splitting the surface by owning domain):**

* Every authorization change is at the **tool-handler level** — inside the registered verb, or
  inside `_anon_guard` / the new `_operator_gate`, which every verb already routes through. Nothing
  keys off this file's path, its line numbers, or a name's spelling in source text. Move a verb into
  `core/mcp/<domain>/` and its gate goes with it.
* `_operator_gate(verb)` is now the single shape for operator authority; all seven operator verbs
  use it. When you split the file, that helper is the thing to put in the shared module — five call
  sites each hand-rolling `try/except _NotOperator` is how three verbs ended up with no gate at all.
* **New module `deploy/dogfood/rig/rig_secrets.py`** (203 lines) — the sealed credential store and
  the re-exported redactor. It is import-time coupling, and it is written for your move:
  `agent_src()` **searches upward** for `core/agent` rather than counting `parents[3]`, so relocating
  the rig does not silently break encryption. `VEXA_AGENT_SRC` overrides.
* **New `deploy/dogfood/rig/tests/`** (7 files, 22 tests) — they reach tools through
  `mcp._tool_manager._tools[name].fn`, never by reading the source, so they follow the verbs
  wherever they land. `conftest.py` stubs the MCP SDK and redirects every store to a tmp dir; the
  README carries the gate→row table.
* `deploy/dogfood/rig/README.md` and `TOOL-USAGE.md` are **untouched** — B8 owns their corrections.

**Outside the rig:** `core/flows/src/flows_timeline/service.py` gains `list_reactions()` (the
subject scoping for R-D07/R-D12, implemented in the service that owns the rows), `flows_api.py`'s
`GET /reactions` becomes a thin forward over it and takes a `subject=` parameter,
`clients/terminal/src/app/api/minutes/seed/` gets its shell spawn extracted to `dockerSh.ts`
(R-E12), and `core/agent/tests/test_workspace_credentials.py` follows `_refuse_credentials` now that
it imports the shared redactor instead of re-implementing it.

DmitriyG228:
### R-D07 corrected: `reaction_signal` is owner-scoped, not operator-gated

Founder ruling, and it is the right one: **a person cancelling the join they scheduled is the
product.** `bot_schedule` mints that reaction, so an admin-only gate would have closed the ordinary
path to fix an unusual one — `bot_schedule` would mint work its own author could not stop. Force-pushed
onto `caf73c99c`.

**The check is ownership, decided by the service that owns the row.** No rig-side string matching:

* `flows_timeline/service.py:91-118` — `reaction_concerns(db, reaction_id, subject=…)` on the same
  `concerns()` predicate as the read half (uid **and** email, because the invite lineage carries an
  organizer address and no uid and the completed lineage the reverse).
* `flows_api.py:363-388` — `POST /reactions/{id}/{verb}` takes `subject=`; **403** when it is not
  theirs, **404** when there is no such reaction. Omitting `subject` keeps the unscoped operator
  behaviour the route has always had, for the admin console.
* `vexa_control_mcp.py:2256-2292` — `uid = me()`, `subject=<uid>` on the call, and a refusal that
  names which of the two happened.

**Three outcomes, not two.** "not yours" for an id nobody has sends the caller off to re-derive an
id; "no such reaction" tells them to ask for a real one.

**A direct lookup, not a windowed scan.** Reusing `list_reactions` here would have been the shorter
edit and would have started refusing exactly the reactions people most often cancel — the ones that
have been sitting scheduled a while, past the projection's `LIMIT 100`. There is a test for that
specifically (150 filler rows, then a 90-day-old reaction of the caller's, which must still be
steerable).

**The operator gate now stands on `friction_dump` / `friction_fixed` only** — the two verbs that
read and mutate every user's ledger.

New tests: `test_rd07_reaction_signal_steers_only_the_callers_own_reaction` (colleague's id →
`refused: not_yours`; own → cancelled, with `subject=` on the wire; ghost id → `no_such_reaction`)
and `test_rd07_reaction_signal_is_not_operator_gated`, which pins the ruling so it cannot drift
back. Flows side: 5 more in `test_reactions_scope.py`.

**`/w/…?token=vxa_mcp_` links dying is confirmed accepted** — unchanged in this push.

Post-rebase on `caf73c99c`: rig 24 · agent+flows+rehearse 1,977 (1 skipped) · terminal 4 ·
pre-push static gates 14/14, pushed without `--no-verify`.

Also new, and worth a reviewer's eye: **#1428 changed both `core/agent` modules `rig_secrets`
hard-imports.** `secret_store` no longer generates the master key on a read path (R-E13) and
`git_redaction`'s generic run dropped 36 → 24 characters (R-E09). Both are correct; both now have a
test on this side, because the first decides whether an unconfigured rig gets encryption or silently
falls back to nothing, and the second decides whether `workspace_attach` still accepts an ordinary
repository URL.

Issue #1481, opened 2026-09-03
flows runs, tests and ships without the agent prompts, a private vault, or a named container


DmitriyG228:
### Value this issue delivers

The flows service runs, tests and ships without the agent's prompts, a private vault, or a named container on one host.

Four things, each witnessable on its own:

1. A checkout without `behavior/prompts/` **collects and runs the flows suite** — today fifteen test modules fail to import, and a deployment that never dispatches an agent turn still cannot load the flow definitions describing its own meetings.
2. A checkout without `core/agent/` **runs the flows suite green** — today the cross-domain room guard dies as an error rather than skipping.
3. A deployment sends mail **from credentials it named itself**, and is refused by name when it did not — today the code decrypts a file inside one developer's home directory.
4. flows reaches its database, and only its database, **from the config contract** — today it reads a Postgres password out of a container belonging to a different stack, and writes fixture rows into the meetings database through the same container.

### Why this matters

The no-agents minutes product (PRD decision 40.6) is gateway + meetings + flows + identity. The by-need cut for main (decision 43) does not ship `core/agent/`, and `behavior/{prompts,asks,workspaces}` stay on the branch as agent-only content. On that tree, as the code stands:

- `flows_defs/production.py` cannot be imported, so **no flow definition loads at all** and the product has no meetings pipeline.
- `core/flows` cannot run its own tests, so nothing about the product can be proven before it ships.
- The mail path needs `sops` and a decrypted file at `~/dev/vexa-secrets/business/vexa-mail.enc.env`. No deployment can satisfy that; the operator gets `FileNotFoundError: sops` or a `CalledProcessError`, neither naming the real problem.
- `db_url()` composes `postgresql+psycopg://postgres:<password>@127.0.0.1:5458/flows` from a container's environment. On a host running more than one stack this **does not fail — it addresses somebody else's data.** It is the same defect the config contract's own DOORS block already documents for `http://localhost:18057`, with a password on the end.

Separately, the recorded test corpus carries a named account's mail domains, meeting title and people, in a repository that is about to be public.

### Where we are (honest)

| what | where | today |
|---|---|---|
| kickoffs read at module import | `core/flows/src/flows_defs/production.py:74-133` — `_SHOWCASE` at module scope, then `ONBOARD_KICKOFF = _prompt("onboard-person.md")` ×3 | `FileNotFoundError` at import; 15 flows test modules fail to collect (188 of 418 tests collected) |
[...]

PR #1483, opened 2026-09-03
flows runs, tests and ships without the agent prompts, a private vault, or a named container

DmitriyG228:
**Delivers issue:** #1481

Six commits, one per change, on `minutes-mcp-viewer`. Base `f94c8fa7451c7012aaa2db84e6129142f24d2c8b` · head `5b3c338eb1b75c611800abc94fe1242a3ff4555f`.

## Contribution rights

> The rights box is deliberately unticked and the commits carry no `Signed-off-by` — both are the
> submitting human's certification to make, not a tool's.

- **C1 — the behavior prompt tree is content, read at step time** (`c77e49534`) · **ran:** `mv behavior/prompts /tmp/` then `uv run pytest -q --collect-only` in `core/flows`, at base and at head · **saw:** base **188 collected, 15 collection errors**, every one a `FileNotFoundError` raised from `production.py` module scope; head **418 collected, 0 errors** · **concluded:** the absence was never a code path, it was an import. Made `_showcase()` resolve when asked, guarded both reads with `is_file()`, deleted the three module-scope `*_KICKOFF` constants, and had `prompt_for(ctx, fname)` resolve the baked default lazily on the fall-through — so a call an override answers now touches no file at all. The two steps that need a kickoff catch the typed `PromptAbsent` and return `NotPresent("behavior", …)`, the same terminal no-retry outcome #1453 gives a step whose agent domain is absent.
  Writing the test taught me something I had wrong: blinding only `Path.is_dir`/`is_file` left the real checkout answering `read_text`, so the first version of the test passed against the unfixed code. The absence is only simulated when `read_text` is blinded too — which is exactly the shape of the bug, since the old `_prompt` never asked whether the file was there.
  `open_*` resolves its kickoff **before** `ag.workspace_init`: a terminal outcome must not leave a seeded workspace behind for a conversation that will never open.

- **C2 — the cross-domain room guard skips where agent-api is not in the tree** (`c843efd03`) · **ran:** `uv run pytest -q tests/test_room_read.py` with `core/agent/` moved away, at base and head · **saw:** base **1 failed** with `StopIteration` at `test_room_read.py:356`; head **26 passed, 1 skipped**, the skip naming `/home/dima/dev/wt-d18/core/agent/control_plane absent` · **concluded:** the guard searched upward for the tree it was checking, so its own anchor vanished with its subject. Anchored it on `core/flows` instead and skipped with the reason named. Deliberately **not** a silent pass: a green that checked nothing would report agreement between two services when only one is present, which is worse than no guard. With `core/agent` present: 27 passed, guard unchanged.

- **C3 — mailbox credentials come from the config contract** (`f3de67db7`) · **ran:** `grep -rnE 'sops|vexa-secrets|~/dev|subprocess' core/flows/src`, and the new refusal tests, at base and head · **saw:** base — `emailx.py:15` `VAULT = os.path.expanduser("~/dev/vexa-secrets/business/vexa-mail.enc.env")` and `:31` `subprocess.run(["sops","-d",VAULT], check=True, …)`; head — no shell-out anywhere under `core/flows/src` · **concluded:** deleted both. `creds(*, login=True)` reads the declared keys and raises `ConfigError` naming what is missing, the same shape `common.require_admin_key` already uses — key named, value never. `VEXA_MAIL_APP_PASSWORD` is reclassified `required-explicit` and joins a new `flows_config.SECRETS` set (P14: fed from a secret store by the deploy surface, never read out of one by the service); `MAIL_KEYS` groups the capability's four keys so a half-configured mailbox is visible as one thing.
  The half a blanket refusal would have broken: mailpit takes no login. `_needs_login()` decides that **before** a socket is opened, so an unconfigured deployment is refused without first connecting to a mail server.
  **Deviation from the brief, stated plainly:** I made the credential pair `required-explicit` but left `VEXA_MAIL_SMTP_HOST` a capability and `VEXA_MAIL_SMTP_PORT` `defaulted`. A required-explicit host would delete the documented "unset = Gmail SMTP over SSL" default, and the config contract's own test forbids a default on a required-explicit key — so a required-explicit port cannot keep its `25`.

- **C4 — the database is a door; no source reaches into a named container** (`261bb7115`) · **ran:** the same source scan plus `common.db_url()` with `VEXA_FLOWS_DB_URL` unset · **saw:** base — `common.py:264` reading `$POSTGRES_PASSWORD` out of `vexa-v012-postgres-1` and composing a DSN against `127.0.0.1:5458`; `meeting.py:464` writing seven fixture transcript rows into the **meetings** database through `docker exec … psql`; head — `subprocess` is absent from `core/flows/src` entirely and an unnamed database raises `ConfigError` naming the key · **concluded:** the DSN fallback is the `http://localhost:18057` defect the contract's own DOORS block documents, with a password on the end — on a host running more than one stack it does not fail, it addresses somebody else's data. Reclassified `VEXA_FLOWS_DB_URL` `required-explicit` and made `db_url()` refuse.
  The fixture injection I **deleted rather than moved**, and that is a real loss stated as one: it was flows writing past another domain's API into its tables. No deployment surface ever set `VEXA_FLOWS_FIXTURE_TRANSCRIPT`, so the path existed only on one machine; a transcript double belongs to the domain that owns the words. `FIXTURE_LINES` and the capability key go with it.

- **C5 — fixture labels stop naming customers** (`29f2cebab`) · **ran:** `grep -rniE 'dna\.test|oenb|ASWF|DNA TSC'` over `core/flows`, `core/meetings`, `core/runtime`, `deploy/helm`, at base and head; then both affected suites · **saw:** base **80 matches**; head **0** · **concluded:** rename only — `dna.test` → `example.test`, `oenb.at` → `example.com`, the title → `Platform Sync`, the group → `platform-sync`, the two named people → invented neutral names, and the `.eml` renamed with `git mv`. The deliberate traps are preserved on purpose: the mixed-case addresses that prove case-insensitive matching, the `.5Z`/`.503Z` RFC3339Nano ordering trap, the RFC 5545 folds.
  The suites caught two things I would have shipped wrong. `parse_ics(…, "VeXa@OeNB.at")` — a case-mangled self address my lowercase substitution missed — turned `test_participants_exclude_us_case_insensitively` red, which is exactly what that test is for. And the longer group slug pushed the ICS `DESCRIPTION` line from 75 to 80 octets; I re-folded it at 75, since the fixture exists to prove the parser unfolds.
  **Beyond the brief, deliberately:** `core/meetings/.../tests/test_calendar_sync.py` was worse than anything on the list — a **real** organisation's mail domain with plausible employee names against it. Renamed the same way, together with the product-source comment in `calendar_sync/service.py` that named the same customer, and the shared invented workspace id in `core/runtime/tests/test_isolation.py`.

- **C6 — the behavior README says what a deployment must carry** (`5b3c338eb`) · **ran:** read it against the tree and against C1's finding · **saw:** a flat list naming `prompts/` and `workspaces/` as though the tree were fixed — and the code believed it · **concluded:** replaced the list with a table saying, per subdirectory, what it is for and whether it ships with the no-agents product, plus the rule the machinery must follow: absent content is a deployment fact, not a broken build.

## Acceptance floor

Base `f94c8fa74` · head `5b3c338eb`. Every red control below was observed on the base worktree, not reasoned about.

| Row | Evidence |
|---|---|
| A1 | `mv behavior/prompts` → `pytest --collect-only`. **Base: `188 tests collected, 15 errors in 0.96s`**, `Interrupted: 15 errors during collection`. **Head: `418 tests collected in 0.30s`, 0 errors.** Full run at head with prompts absent: 408 passed, 5 skipped, **5 failed** — and honestly so: those five assert on the words of the showcase kick, which is content that deployment did not mount. They are not made to skip, because in a checkout that *does* carry the showcase their failure would be a real regression. |
| A2 | `test_post_meeting_answers_not_present_when_the_kickoff_is_absent`. **Base: FAILED — `AssertionError: a turn was dispatched with no kickoff`** (the kickoff was baked at import, so the step never noticed the tree was gone). **Head: passes** — `NotPresent(domain="behavior")`, reason names `process-meeting.md`, `ag.dispatch_turn` never called. |
| A3 | `mv core/agent` → `pytest tests/test_room_read.py`. **Base: `1 failed, 26 passed`, `StopIteration` at `test_room_read.py:356`.** **Head: `26 passed, 1 skipped`**, skip reason names the absent `core/agent/control_plane`. With `core/agent` present, head: **27 passed** — the guard still runs. |
| A4 | `pytest tests/test_flows_secrets.py`. **Base: 5 failed** (source scan, unnamed-mailbox refusal, half-pair refusal, the double's no-password path, the credential declaration). **Head: 33 passed.** Source scan at head: `sops`, `vexa-secrets`, `~/dev` and `subprocess` are all absent from `flows_steps/emailx.py`. |
| A5 | `pytest tests/test_flows_doors.py`. **Base: 3 failed**, including `Failed: DID NOT RAISE ConfigError` and the container scan listing `common.py:264` / `meeting.py:464`. **Head: 15 passed.** `grep -rnE 'subprocess\|vexa-v012' core/flows/src --include=*.py` → nothing. |
| A6 | `grep -rniE 'dna\.test\|oenb\|ASWF\|DNA TSC'` over `core/flows core/meetings core/runtime deploy/helm`. **Base: 80 matches. Head: 0.** |
| A7 | `behavior/README.md` in the diff: the flat contents list is replaced by a per-subdirectory table with a "ships with the no-agents product" column. |
| A- | **No regression at head `5b3c338eb`:** `core/flows` **413 passed / 5 skipped** (base 397/5 — +16 are this PR's own tests); `meeting-api` **1328 passed / 5 skipped**; `core/runtime` isolation 7 passed; `gate:python` green over **16 packages**; **26 non-Docker gates green** (readme, lite-makefile, docs-version, dataflow, isolation, isolation-py, exports, graph, graph-py, schema, contract-version, config-contract, db-schema, db-budget, node, health, access, tracing, telemetry, licenses, image-licenses, runtime-parity, execution-env, test-isolation, arch-report, parity) plus contract-conformance; the pre-push hook's 14 fast gates green on the push itself. |

## Docs diff (D6c)

| page | altitude | what changed |
|---|---|---|
| `behavior/README.md` | concept | which subdirectories ship with the no-agents product, and the rule that the machinery may not assume the rest |
| `core/flows/src/flows_config.py` (declaration prose) | reference | the mail and database keys carry their class, their `why`, and the statement that nothing sits behind them |
| `core/flows/tests/mailpit/README.md` | reference | the renamed fixture, and "two invented ATTENDEEs" where it used to say "two real" |
| `core/flows/tests/test_health.py` (module docstring) | reference | it told the reader `db_url()` shells out to `docker exec`; it now says it refuses by name |

**Named pages that needed no change, argued:** nothing under `docs/` taught either fallback — `sops`, the vault path and `vexa-v012` appear nowhere in `docs/`. That silence is part of the defect being fixed, not evidence the docs are fine: an undocumented fallback is one no operator could have known to satisfy.

- **Secrets:** the diff **removes** the only credential-store read in `core/flows` and the hard-coded path to a decrypted vault. `grep -rnE 'sops|vexa-secrets|~/dev' core/flows/src` at head → one hit, and it is prose: a sentence in `flows_config.py`'s `why` for `VEXA_MAIL_ADDR` recording that the fallback existed and is gone. No key, token, password or path to one is added anywhere.
- **Least privilege / blast radius:** two `docker exec` invocations are deleted. flows no longer reads another container's environment, and no longer writes into another domain's database. Its database reachability is now exactly what the deployment names.
- **Fail-closed:** every new failure mode refuses rather than guesses — `ConfigError` naming the key, never a default. The refusals name keys and never values (asserted by `test_the_refusal_names_the_key_and_never_the_value`).
- **Licence / dependency scan:** `gate:licenses` green — 514 deps OSS-clean, 2 Cat-B by standing exception; `gate:image-licenses` green — 7 pinned images + 1 bundled component. No dependency is added or changed by this PR (`pyproject.toml` and lockfiles untouched).
- **SAST-class structural gates:** `gate:isolation`, `gate:isolation-py`, `gate:test-isolation`, `gate:exports`, `gate:graph`, `gate:config-contract`, `gate:access` all green at head.
- **Not run, and why:** the Docker-backed gates (`stack`, `compose`, `eval`, `replay`, `compose-stress`, `compose-chaos`, `eval-baseline`) — this diff changes no image, no compose file and no helm template, and CI owns them authoritatively on the same sha.

## Validation request

**Who:** any competent non-author. No live meeting, no second operator, no running stack — every observation is a checkout-and-run property, about ten minutes.

**What they'll watch,** on a fresh worktree at `5b3c338eb` (Linux, `uv` on PATH):

1. `mv behavior/prompts /tmp/` → `cd core/flows && uv run pytest -q --collect-only | tail -1` → **418 collected, 0 errors**. Restore. (Same command at `f94c8fa74` → 188 collected, 15 errors.)
2. `mv core/agent /tmp/` → `uv run pytest -q tests/test_room_read.py -rs` → **26 passed, 1 skipped**, reason names the absent path. Restore, re-run → 27 passed.
3. `grep -rniE 'sops|vexa-secrets|vexa-v012|subprocess' core/flows/src --include=*.py` → one prose hit in `flows_config.py`, nothing executable.
4. `uv run pytest -q` in `core/flows` (413/5) and in `core/meetings/services/meeting-api` (1328/5); `node scripts/gates.mjs python`.
5. Read `behavior/README.md` and the six commit messages, and sign the **docs story**: the README lands at the right altitude for someone deciding what their deployment must carry, and the contract's `why` lines are reference prose, not narrative.

**Deployment validated (D12b):** repo checkout / CI on Linux, at `5b3c338eb`, built by `git worktree add` from a long-lived clone, no env deltas from stock beyond `uv` on PATH. `deploy/helm/charts/vexa/values.yaml` changes one commented example value and no template, key or default — a helm render was not run and is **not claimed**. Lite compose, full compose and hosted are untouched by this diff and stay honestly unclaimed.

**The one deployment-visible consequence, stated plainly:** a deployment that relied on either deleted fallback now refuses at the read with a message naming the key. Operators must export `VEXA_MAIL_ADDR` + `VEXA_MAIL_APP_PASSWORD` for a transport that logs in, or `VEXA_MAIL_ADDR` + `VEXA_MAIL_SMTP_HOST` (+ `VEXA_MAIL_SMTP_PORT`) for one that does not, and `VEXA_FLOWS_DB_URL` always. The dogfood lane already exports all of those and was not modified.

**One anomaly, reported not worked around:** the pre-push hook's `gate:contract-conformance` failed the first push with `/bin/sh: 1: uv: not found` — the hook shells out to `uv` and does not inherit a login PATH. Pushing with `uv` on PATH made all 14 fast gates green. Nothing in this diff causes it and I have not changed the hook; it will bite anyone whose `uv` lives in `~/.local/bin`, and it is worth its own issue.

## Authorship
Tooling disclosure (optional, welcome, never an attribution): written with agent assistance on the bbb host; every command above was run and its output read, not inferred.


DmitriyG228:
`docs: none` — the reason, so the decision is auditable rather than a label nobody can question.

This PR **does** change documentation; none of it lives under `docs/**`:

| file | altitude |
|---|---|
| `behavior/README.md` | concept — which subdirectories a deployment must carry |
| `core/flows/src/flows_config.py` | reference — the mail and database keys' class and `why` |
| `core/flows/tests/mailpit/README.md` | reference — the renamed fixture |
| `core/flows/tests/test_health.py` docstring | reference — it told the reader `db_url()` shells out to `docker exec` |

Nothing under `docs/**` needs to change, and that is argued rather than assumed: `sops`, the vault path and `vexa-v012` appear **nowhere** in `docs/`. The two fallbacks this PR deletes were never documented — which is part of the defect, since an undocumented fallback is one no operator could have known to satisfy. There is no page teaching either behaviour that could now be wrong.

Also on this run, for the reviewer:

- **`gates` / `python` is red for a known, pre-existing reason and not for this diff.** It fails on `core/agent/tests/test_entity_writeback_trim.py::test_a_phase_over_budget_leaves_no_child_process` — [#1434](https://github.com/Vexa-ai/vexa/issues/1434), open, *"gate:python is red on one runner-timing test, and it hides every package after core/agent"*. This PR touches no file under `core/agent`. The same leg is red on #1458 today. On bbb (CPython 3.14) `node scripts/gates.mjs python` is green over all 16 packages at head `5b3c338eb`, which matches #1434's own root cause: the kill is delivered by garbage collection on the affected interpreter.
- **`contribution-rights` and `DCO` are red by design.** The rights box is unticked and no commit carries a `Signed-off-by` — both are the submitting human's certification, not a tool's.
- **`merge-card`** needs `state: value-signed`, which is the founder's stamp after validation.

`static`, `node`, `lite-smoke`, `value-fsm` and GitGuardian are green.

Issue #1468, opened 2026-09-03
Your own Claude Code can list and cancel the meeting joins you scheduled — and only yours


DmitriyG228:
### Value this issue delivers

> A person's own Claude Code can list and cancel the meeting joins **they** scheduled, through the Vexa MCP endpoint, using their own Vexa credential — and can see nobody else's.

### Why this matters

The product configuration we are shipping is "connect your Claude Code to meetings": the person's
agent talks to the Vexa endpoint with the person's own credential, and nothing else. Two of the
four reaction-engine tools on that endpoint — `reactions_list` (what am I waiting on) and
`reaction_signal` (stop that join) — are the ones a person reaches for after they have scheduled a
bot and changed their mind.

Today neither works for a person, in two different ways, and both of them are quiet:

* **The credential is refused.** Every route on flows-api requires a deployment-wide operator key
  (`X-Flows-Admin-Key`). A person's bearer is not that key, so the call is refused — but the
  refusal arrives inside a JSON-RPC response with HTTP 200, so what the agent has to work with is a
  success envelope with a 401 in the text.
* **The alternative is worse.** The only way to make the call succeed is to give the endpoint the
  operator key. That key is not scoped to a person: `GET /reactions` with it returns *every*
  reaction in the instance, and `POST /reactions/{id}/cancel` with it cancels *any* of them. That
  is the shape the endpoint had before subject scoping was added, and the reason subject scoping
  was added (#1430).

So the endpoint can be wired one of two ways and both are wrong: refuse the person, or hand them
the instance. There is no third way today, because the tool manifest has no field that says which
credential a tool's door takes — so nothing can notice the mismatch, at boot or at call time.

Observed 2026-09-03 08:51Z against `mcp.dev.vexa.ai/mcp`: `tools/list` returns 18 tools including
flows' four; `reactions_list` called with a person's bearer returns JSON-RPC 200 carrying
`Error calling reactions_list. Status code: 401.`

### Where we are (honest)

| fact | where |
|---|---|
| every flows-api route sits behind one operator key | `core/flows/src/flows_integrations/flows_api.py:132` (`auth`), applied at `:198, :219, :265, :310, :359, :372, :398`; `:137` `timeline_auth` adds a second read-only key |
| the key comparison is constant-time and has no weak default | `core/flows/src/flows_integrations/flows_api.py:56-79, :119-129` |
[...]

Issue #495, opened 2026-07-10
api-gateway: shared HTTP pool starves token validation under load — valid API keys mass-401 (the #483 mechanism)


DmitriyG228:
## Value this change must deliver (the sentence the signer attests to)

> **A valid API key authenticates on every call — including while the system is slow or under
> load — and when our auth infrastructure itself fails, the caller is told the truth (503,
> retry), never "Invalid API key".**

## The problem (in our users' words)

> "API persistently returns *Invalid API key* on POST /bots with valid keys" — #483 (same
> mechanism reported in #479, #482, #484; a paying customer's integration went down on it;
> at least one paying user suffered silently for three days and never filed anything)

Between Jul 7–9 our hosted gateway returned `401 Invalid API key` **to keys that were perfectly
valid**, in recurring bursts. We verified in production data that no key was ever invalidated —
the gateway simply failed to *check* them under load, and reported that failure as if the key
were bad. Full incident analysis is published in the tracker (#483).

## The mechanism (root cause, named)

The gateway creates **one** `httpx.AsyncClient` and uses it for two unrelated jobs:

1. **forwarding requests** to meeting-api (long-lived: up to `timeout=30.0`), and
2. **validating API keys** against admin-api (`POST /internal/validate`, needs ~5ms).

When meeting-api slows down (DB lock convoy, load — anything), in-flight forwarded requests
occupy the client's entire connection pool (default cap: 100). Token validation then waits on a
pool slot it never gets, throws `PoolTimeout` — and the caller treats that as an invalid key.
**Result: the slower the system, the more valid users get locked out.** The 401s then trigger
dashboard re-auth loops, which multiplies the load. This is self-amplifying.

Two aggravators mask the failure:

- `httpx` timeout exceptions stringify to an **empty message**, so 0.10 logged
  `Token validation failed:` with nothing after the colon (this hid the mechanism for hours
  during the incident).
- The current 0.12 code goes further and **swallows the exception entirely**
  (`except Exception: return None`) — an infrastructure failure is *indistinguishable* from a
  bad key, with zero log line.

**Where it lives (both code lines are in this repo's history/production):**

| Surface | Location | Defect |
|---|---|---|
[...]

PR #1486, opened 2026-09-03
Your own Claude Code can list and cancel the meeting joins you scheduled — and only yours

DmitriyG228:
**Delivers issue:** #1468

## Contribution rights

Base `adfea1ee7e2efef6c3767319ff889782e8863242` · head `2b184754eefd92bd9222c79ba958933d3b167b38`.
Thirteen commits, each component a red commit and a green one.

- **C1 — the person's own credential.** *Ran:* the four subject-scoped routes against a stubbed
  identity, at base and at head. *Saw:* at base, a person's bearer → `401 X-Flows-Admin-Key
  required`; the operator key → **HTTP 500**, `TypeError: list_reactions() got multiple values for
  argument 'status'`. *Concluded:* the route was unusable for a person AND broken for the operator,
  and the second had never been seen because the first fired in front of it — the route function
  `list_reactions` rebinds the module global imported at the top of the file, so line 388 called
  itself. The flows suite tested the model function and never the route. Both fixed; 40 new
  observations.
- **C2 — the manifest could not express the problem.** *Ran:* `validate` / `assemble` over the
  repository's manifest and synthetic ones. *Saw:* a tool could say who the CALLER must be
  (`identity`) and nothing about what this edge must PRESENT to the door. *Concluded:* the edge had
  exactly one move — forward the caller's credential to everything — so the mismatch could only be
  discovered at call time, by the agent. `auth` makes it a boot question.
- **C3 — the envelope, measured before it was written.** *Ran:* an upstream 401 through an
  assembled tool, at the mounted `/mcp` transport, at base. *Saw:*
  `{"jsonrpc":"2.0","id":2,"result":{"content":[{"type":"text","text":"Error calling
  reactions_list. Status code: 401. …"}],"isError":true}}`. *Concluded:* **the failure was already
  carried.** F113 read the JSON-RPC transport's own 200 as the defect; the 200 is correct — JSON-RPC
  carries tool failures inside a result — and `isError` was set. That half of the component is
  **INVALIDATED**, and the assertion ships anyway: the flag is produced inside `fastapi-mcp` and
  nothing in this repository pinned it.
- **C4 — what the same measurement found instead.** *Ran:* `tools/list` for the assembled tools.
  *Saw:* `reactions_list -> {"type":"object","properties":{},"additionalProperties":false}` — and
  the same for `timeline` and `reaction_signal`. *Concluded:* every declared argument disappeared
  between `bind`, which had just verified each one against the owning route, and the surface,
  because the route's endpoint took `request` and nothing else. `reactions_list` could not filter
  and **`reaction_signal` could not be called at all** — `/reactions/{reaction_id}/{verb}` cannot be
  addressed without its two path parameters and an agent could not see they existed. The issue's
  "cancel the join you scheduled" was unreachable for that reason alone. Fixed by building the
  route with a real signature.
- **C6 — `GET /queue/waiting`, found on the rebase.** *Ran:* a grep for what else on this surface
  claims a subject, after #1482 merged. *Saw:* the route is gated by the operator key and takes the
  person from an `X-User-Id` header its docstring says *"the MCP assembler stamps"* — and **the
  assembler stamps no such header**: `register.py` forwards the caller's credential as `X-API-Key`
  and nothing else. Its manifest entry meanwhile already declares `auth: "subject"`.
  *Concluded:* through the assembly that route has no subject at all, so it answers 400 — or, if
  anyone reaches for the operator key to make it answer, it answers about whoever the query names.
  Contract and door disagreeing is the shape this whole PR is about, one route later, so it joins
  the other four: a person's subject comes from their credential, and `X-User-Id` is ignored when
  one is present. The operator path is untouched, including `X-User-Id` outranking `?subject=`.
- **C5 — the operator header (added mid-flight, from the deployment side).** *Ran:* a grep for the
  header across the line after #1476 merged. *Saw:* `admin_api/app/events.py:74` reads
  `VEXA_FLOWS_API_KEY` — flows-api's own operator key, correctly — and sends it as
  `X-Flows-Admin-Key`. *Concluded:* the name says admin-api's token and never meant one; a lane
  start script carries a `changeme` for exactly that reason. Renamed to `X-Flows-Operator-Key`,
  old name accepted for one release with a once-per-process deprecation line.

## Acceptance floor

| Row | Evidence |
|---|---|
| A1 | `GET /reactions` with a person's bearer → 200, their two rows only. RED at base: the same call with the operator key answered **500** (`TypeError: list_reactions() got multiple values for argument 'status'`), and with a bearer **401**. `test_a_persons_own_bearer_lists_their_own_reactions`, `test_the_operator_key_still_lists_every_reaction`, `test_the_api_key_header_carries_a_bearer_too` — `core/flows/tests/test_subject_bearer.py` |
| A2 | A second person's bearer returns only their own row; `POST /reactions/r-ben/cancel` as Anna → **403**, and the row is still `admitted`. Control: Anna cancelling her own → 200, row `cancelled`. Plus `test_ownership_is_checked_even_when_no_subject_argument_is_sent` — before this, ownership was checked only when the caller ASSERTED a subject |
| A3 | `test_no_credential_at_all_is_refused` × 4 routes → 401. Control: `test_the_operator_key_still_opens_every_one_of_them` × 4 → not 401, unscoped operator behaviour unchanged |
| A4 | `test_a_persons_bearer_does_not_open_an_admin_route` × 4 (`POST /flows`, `/events`, `/events/batch`, `/flows/{n}/{v}/{action}`) → 401. Control: the operator key on the same four → not 401 (past the door each has its own answer — the instance gate's 409, an unknown version's 404 — and this row is about the door) |
| A5 | `subject=512` as Anna → **403** naming it; `subject=126` and `subject=ANNA@vexa.test` → 200; every answer carries `"subject"` so the caller sees whose rows came back. Control: the operator reading `subject=512` → 200, unchanged |
| A6 | Identity unreachable → **503**; identity answering 401 → **401**; identity answering **403** (our internal secret, not their key) → 503. A resolver collapsing these would pass one row and fail the others — all three are asserted. `test_the_resolver_is_the_one_the_gateway_asks` pins that it is `/internal/validate` over `X-Internal-Secret`, not a second lookup |
| A7 | `test_an_admin_tool_whose_key_this_deployment_does_not_hold_refuses_the_boot` — `ManifestError` naming `flows_retire` and `VEXA_MCP_FLOWS_ADMIN_KEY`; `test_an_unsatisfiable_tool_never_reaches_the_surface` — no route exists for it. Controls: the same manifest with the key set assembles and serves it; a placeholder value refuses; a manifest with no `auth` refuses naming the field — `core/meetings/services/mcp/tests/test_tool_auth.py` |
| A8 | `test_an_upstream_refusal_arrives_marked_as_an_error` and `test_a_forbidden_reaches_the_agent_as_the_domains_own_answer` → `isError: true` with the status and the domain's own words. Control: `test_an_upstream_success_is_not_marked_an_error`. **Green at base** — see C3; pinned, not fixed |
| A9 | **Not claimed.** The live run on the rig lane is the human part of this issue and is not mine to sign — see *Validation request*. Nothing here asserts it |
| A10 | `GET /queue/waiting` with a person's bearer answers about **them** — `test_the_queue_takes_a_persons_bearer_and_answers_about_them`; a `subject` naming someone else → 403 and nothing is read; `X-User-Id: 512` alongside Anna's bearer still answers about 126 (`test_a_stamped_user_id_does_not_beat_an_authenticated_person`). Controls: the operator key with `X-User-Id` still outranks `?subject=` (the row `tests/test_queue_waiting.py` owns, still green), and no credential → 401 with nothing read |
| A- | flows **495 passed / 5 skipped**; mcp **214 passed**; admin-api **147 passed**. `gate:python` — 16 packages green. `gate:readme`, `lite-makefile`, `docs-version`, `dataflow`, `isolation`, `isolation-py`, `exports`, `graph`, `graph-py`, `schema`, `contract-version`, `config-contract` (7 services · 311 keys), `db-schema`, `db-budget` — all green at head. Pre-push static gates green, including `contract-conformance` |

**Beyond the floor**, all witnessed, all blocking the issue's own value: the `GET /reactions` 500
(A1's red), assembled tools publishing an empty input schema (C4), and `GET /queue/waiting`'s
contract disagreeing with its door (A10). None was in the prepared table because none had been
seen — the last one did not exist when the table was written.

## Docs diff (D6c)

| page | altitude | change |
|---|---|---|
| `docs/docs/flows/authoring.mdx` | reference | the API table's one-line `Auth: X-Flows-Admin-Key` becomes the two tiers, and says what a person may do with their own credential |
| `docs/docs/flows/developing.mdx` | how-to | the debugging `curl` sends the renamed header |
| `core/flows/README.md` | reference | new *Two tiers at the door*: which caller opens what (five routes), that the subject comes from identity's one resolver, that an unreachable identity is 503, and the header rename. #1482's *What is waiting is flows* section is kept beside it, unchanged |
| `core/meetings/services/mcp/README.md` | reference | new *The manifest contract* — the `identity` vs `auth` table, what each value means, how a tool's arguments are published, and the operator-visible migration note |

**No `mcp.tools.v1` schema file exists** on this line — the contract is `manifest.validate()` plus
that README section, and both say so. Publishing one (P4) would be a second place to write the same
rule and is deliberately not in this PR.

## Operator-visible changes

1. **`auth` is required in `mcp.tools.v1`.** A manifest written against the previous shape —
   including one supplied through `VEXA_MCP_MANIFEST_DIR` — refuses the boot, naming the tool and
   the field. Only one manifest exists in this repository and it is updated here.
2. **`X-Flows-Admin-Key` → `X-Flows-Operator-Key`,** old name accepted for one release with a
   deprecation line printed once per process. Every caller in this repository moves with it. Lane
   scripts outside the repository (`~/.storm/flows-up.sh` and friends) keep working and will print
   the line.
3. **Assembled tools' path parameters move from the JSON body to the query string,** which is where
   they are now declared. This changes the direct `/tools/<name>` HTTP surface, not the MCP one.
4. **`GET /queue/waiting` no longer honours `X-User-Id` when a person's own credential is present.**
   An operator call is unchanged. Nothing in this repository sent both; a private caller that does
   will start getting the credential's answer, which is the one it should have been getting.

- **No new dependency.** flows resolves over `urllib` through its existing `flows_steps.common.http`
  rather than adding `httpx` for one POST; the mcp service adds only stdlib `inspect` and `os`.
- **No new credential in the tree.** The internal secret is the one flows already reads at boot
  (`INTERNAL_API_SECRET`), passed in rather than re-read, so a missing one is still a boot refusal
  and never a per-request 503. `VEXA_MCP_<DOMAIN>_ADMIN_KEY` is a name the assembler checks for
  presence; no default, and a published placeholder is refused.
- **Widened surface, stated plainly:** four routes that took one operator key now also take a
  person's bearer. Each derives its subject server-side and refuses a subject that is not the
  caller's; ownership on `reaction_signal` is now checked on EVERY call rather than only when the
  caller volunteered a subject, which is a narrowing. The four operator routes are unchanged.
- **Fail directions checked both ways:** an unreachable or misconfigured identity is 503, never a
  verdict on the caller's key (the #495/#483 inversion); a subject nobody answers to owns nothing
  (`REACTION_NOT_YOURS`, unchanged).
- `gate:config-contract` green — no undeclared env read. `gate:isolation-py` and `gate:graph-py`
  green — flows gained no cross-domain import; it calls identity over HTTP, which is the edge it was
  already allowed.
- Secret scanning is disabled on this repository, so it is not evidence here. No new CodeQL surface:
  nothing in this diff logs a credential, formats an exception into a response body, or builds a
  path from caller input.

## Validation request

**A9 is the only row not claimed here, and it needs one operator with two Vexa accounts** — no
meeting participants. Connect a Claude Code to the rig's endpoint as account A, schedule a join,
list it, cancel it; then connect as account B and confirm A's join is neither listed nor
cancellable. Rows A1–A8 are seam-altitude and need no human (D12).

**Deployment (D12b):** the dogfood rig lane on `bbb` is the only deployment that runs flows-api —
there is no flows service in `deploy/compose/docker-compose.yml` yet. Lite and full compose carry no
flows domain, so its tools are absent from `tools/list` by construction and there is nothing to
validate there; the manifest-contract and envelope halves run on every deployment and are validated
at seam altitude. The validator should state the lane's sha, the two account ids, and the
timestamps.

**Watch for three things the tests cannot show:** whether the deprecation line for the old header
appears exactly once per process on a real lane, and whether `reaction_signal` is now offered to the
model with `reaction_id` and `verb` visible in its schema — that is the difference between "cancel
my join" working and the tool being invisible. Third: that `whats_waiting` answers about the caller
through the edge, where no `X-User-Id` is stamped at all.

## Authorship
Tooling disclosure (optional, welcome, never an attribution): written with Claude Code.

### DIFF
--- a/core/agent/shared/timeline.py
+++ b/core/agent/shared/timeline.py
@@ -71,7 +71,7 @@ def fetch(uid: str, *, back: int = BACK, ahead: int = AHEAD, timeout: float = TI
         return ""
     # The window is the route's default (14 back, 30 forward); `limit` is what the block can hold.
     url = f"{api}/timeline?subject={urllib.parse.quote(str(uid))}&format=preamble&limit={back + ahead}"
-    req = urllib.request.Request(url, method="GET", headers={"X-Flows-Admin-Key": key})
+    req = urllib.request.Request(url, method="GET", headers={"X-Flows-Operator-Key": key})
     try:
         with urllib.request.urlopen(req, timeout=timeout) as r:
             body = json.loads(r.read().decode() or "{}")
--- a/core/flows/Makefile
+++ b/core/flows/Makefile
@@ -46,10 +46,10 @@ stop: ## stop all three
 	@echo stopped
 
 status: ## the operator projection (reactions, newest first)
-	@curl -s -H "X-Flows-Admin-Key: $(KEY)" "$(API)/reactions" | $(PY) -m json.tool | head -60
+	@curl -s -H "X-Flows-Operator-Key: $(KEY)" "$(API)/reactions" | $(PY) -m json.tool | head -60
 
 flows: ## list every flow version + the step vocabulary
-	@curl -s -H "X-Flows-Admin-Key: $(KEY)" "$(API)/flows" | $(PY) -m json.tool | head -80
+	@curl -s -H "X-Flows-Operator-Key: $(KEY)" "$(API)/flows" | $(PY) -m json.tool | head -80
 
 check: ## validate a Python flow file: make check FILE=my_flow.py
 	cd $(SRC) && $(PY) flows_authoring.py check $(FILE)
--- a/core/flows/README.md
+++ b/core/flows/README.md
@@ -76,6 +76,25 @@ correct behaviour and not a gap in the queue.
 resolved private-mount-first and read hot, and a pending reaction behavior says nothing about is
 counted rather than spoken. `behavior/queue/README.md` is the contract.
 
+## Two tiers at the door
+
+`flows-api` answers to two kinds of caller and the difference is who the answer is about.
+
+The **operator key** (`X-Flows-Admin-Key`, `VEXA_FLOWS_API_KEY`) is the instance: it opens every
+route, reads every reaction and steers any of them. It is what configures the machine, and there is
+no per-person version of `POST /flows` or `POST /events`.
+
+A **person's own Vexa credential** opens the four routes that are about a person — `GET /flows`,
+`GET /reactions`, `POST /reactions/{id}/{verb}`, `GET /timeline` — and the subject is derived from
+that credential by asking identity's `/internal/validate`, the same resolver the gateway asks. Not
+a second resolver: flows is a second caller of the one that already exists. This is what the MCP
+edge needs, because that edge forwards the caller's own credential and holds none of its own — so
+without it the only way to make flows' four tools answer was to hand the edge the operator key, and
+the operator key is not a person.
+
+An unreachable identity answers **503**, never 401: not being able to ask who somebody is has not
+established that their credential is bad.
+
 ## Configuration
 
 Every environment key this brick reads is declared once in `src/flows_config.py`, with its class
--- a/core/flows/README.md
+++ b/core/flows/README.md
@@ -84,8 +84,9 @@ The **operator key** (`X-Flows-Operator-Key`, `VEXA_FLOWS_API_KEY`) is the insta
 route, reads every reaction and steers any of them. It is what configures the machine, and there is
 no per-person version of `POST /flows` or `POST /events`.
 
-A **person's own Vexa credential** opens the four routes that are about a person — `GET /flows`,
-`GET /reactions`, `POST /reactions/{id}/{verb}`, `GET /timeline` — and the subject is derived from
+A **person's own Vexa credential** opens the five routes that are about a person — `GET /flows`,
+`GET /reactions`, `POST /reactions/{id}/{verb}`, `GET /queue/waiting`, `GET /timeline` — and the
+subject is derived from
 that credential by asking identity's `/internal/validate`, the same resolver the gateway asks. Not
 a second resolver: flows is a second caller of the one that already exists. This is what the MCP
 edge needs, because that edge forwards the caller's own credential and holds none of its own — so
--- a/core/flows/README.md
+++ b/core/flows/README.md
@@ -80,7 +80,7 @@ counted rather than spoken. `behavior/queue/README.md` is the contract.
 
 `flows-api` answers to two kinds of caller and the difference is who the answer is about.
 
-The **operator key** (`X-Flows-Admin-Key`, `VEXA_FLOWS_API_KEY`) is the instance: it opens every
+The **operator key** (`X-Flows-Operator-Key`, `VEXA_FLOWS_API_KEY`) is the instance: it opens every
 route, reads every reaction and steers any of them. It is what configures the machine, and there is
 no per-person version of `POST /flows` or `POST /events`.
 
@@ -91,6 +91,11 @@ a second resolver: flows is a second caller of the one that already exists. This
 edge needs, because that edge forwards the caller's own credential and holds none of its own — so
 without it the only way to make flows' four tools answer was to hand the edge the operator key, and
 the operator key is not a person.
+
+The header used to be `X-Flows-Admin-Key`, which reads as admin-api's token and is not one — a lane
+start script carried a `changeme` for it because whoever wrote it exported admin-api's key under
+this service's name. The old header is accepted for one release, with a deprecation line printed
+once per process; a caller sending both is answered on the new one.
 
 An unreachable identity answers **503**, never 401: not being able to ask who somebody is has not
 established that their credential is bad.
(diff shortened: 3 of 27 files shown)


DmitriyG228:
CI settled at head `0e7819f7f`: **python · static · node · gates · lite-smoke · value-fsm · docs-current · comment · GitGuardian all pass.**

The four reds are the human gates and are meant to be red from an agent's hand: DCO (no `Signed-off-by`), contribution-rights + evaluate (the rights box is unticked — a legal certification only the submitting human may select), and merge-card, whose one ❌ row is `state: value-signed`. The card's Diff row is already ✅.

Local at the same head, on `bbb`: flows 462 passed / 5 skipped · mcp 214 · admin-api 147 · `gate:python` 16 packages · 14 static gates including `config-contract` (6 services · 274 keys) and `contract-conformance`.

Next move is a human's: tick the rights box, sign off the commits, and stamp the value after the A9 live run described under *Validation request*.

Issue #1491, opened 2026-09-03
A fact defined in two places cannot silently disagree


DmitriyG228:
### Value this issue delivers

A fact defined in two places cannot silently disagree — CI names every copy, and every disagreement that is left is one a human has been asked to settle.

### Why this matters

P23 is *one writer per data carrier*. Applied to carriers we enforce it. Applied to **literals** — a set of statuses, a mark, a sentence a stranger reads, a regex, a service's own hostname — nothing checks anything, and the tree shows what that costs.

**The control case is the whole argument.** `deploy/contracts/config.v1/preflight.py` is vendored byte-identically into **seven** packages, md5 `628e79b499…` on every one, and it has never drifted — because `gate:config-contract` fails the build on byte-inequality. It is the only cross-image constant in this repository with a parity gate, and it is the only multiply-written fact that has held.

Everything without that gate has moved, or is one edit from moving. Measured on `adfea1ee7`, not estimated:

- **The entity slug has nine implementations and already resolves one name two ways.** `core/agent/shared/entities.py` NFKD ascii-folds; the other eight do not. A person whose name carries a diaeresis is written by the server as `zoe-muller` and looked for by the terminal's wikilink chip as `zo-m-ller`. **The chip has never resolved for a non-ASCII name.**
- **"Live meeting" has six memberships across fourteen production declarations.** The SQL index predicate that four model files share includes `scheduled` and omits `needs_help`, so the index those queries were built for covers different rows than the application calls live. The rig calls a `joining` meeting not-live.
- **The visibility disclosure has seven spellings, and one is materially weaker.** `flows_defs/production.py` ships *"…own servers; the recording and transcript stay there."* on a live outbound path — dropping the entire clause about a colleague's workspace being visible to the company's agents, which `behavior/mail/README.md` records as founder decision 21.
- **The runtime has two hostnames.** meeting-api defaults to `http://runtime:8090`, core/agent to `http://runtime-api:8090`. Compose sets both keys explicitly, so nothing has broken; the disagreement bites in every deployment that has not copied compose's exact block.
- **`[vexa-machinery]` was written four times across three images, `[vexa-phase:writeback]` three times under two names inside one package.**

[...]

PR #1492, opened 2026-09-03
A fact defined in two places cannot silently disagree

DmitriyG228:
**Delivers issue:** #1491

## Contribution rights

## THE RED LIST — seven decisions, both live answers each

This is the deliverable. `node scripts/check-parity.mjs --ledger` prints it in full; the decision text is in `scripts/parity.json`. **This PR does not settle any of them.**

| fact | sites | answers | the decision |
|---|---|---|---|
| **entity-slug** | 5 | 4 | Which slug the server writes and every client reproduces. `entities.py` NFKD-folds → `zoe-muller`; the other eight implementations → `zo-m-ller`. **The terminal's wikilink chip has never resolved a non-ASCII name.** Also open, separately: the cap is 80 / 60 / 40 / none, and empty input yields `""` / `shared` / `untitled` / `meeting` / `unknown`. |
| **live-meeting-set** | 13 | 5 | What counts as a live meeting. The SQL index predicate (4 model files) includes `scheduled`, omits `needs_help` — the index the live queries were built for covers different rows than the app calls live. The rig calls a `joining` meeting not-live. `auto_join.py` alone carries `needs_human_help`. |
| **visibility-sentence** | 4 | 2 | Whether the production data statement says what the others say. `flows_defs/production.py:864` ships *"…own servers; the recording and transcript stay there."* on a live outbound path, dropping the whole clause about a colleague's workspace being visible to the company's agents (founder decision 21). Not a wording variant. |
| **entity-kind-enum** | 2 | 2 | Whether `topic` and `task` are entity kinds. The terminal types every participant as `topic` and mints `kg/entities/topic/<slug>.md` — a path the server's writer refuses with `EntityRefused`. Reached on every live meeting via `useMeeting.ts`. |
| **internal-secret-class** | 5 | 2 | Whether one env var may have three boot behaviours. Four contracts `required-explicit`; meeting-api `defaulted` (boots and silently stops attaching the internal header); runtime `surface_only`; the MCP service does not declare it at all. A security boundary, not a style question. |
| **runtime-door-default** | 3 | 2 | What the runtime is called. `http://runtime:8090` vs `http://runtime-api:8090` — same port, different DNS name, one service. Compose sets both keys, so nothing has broken; it bites wherever the env is unset. One of the two is wrong. |
| **machinery-note** | 2 | 2 | Which instruction the agent actually gets. The Python copy carries a clause the TypeScript copy does not, and both are handed to the same model on different paths. |

**Deliberately excluded, and named so the exclusion is reviewable:** `reconcile._PRE_ACTIVE_STATUSES`, `_LIVENESS_GATED`, `collector._FSM_OWNED_STATUSES` and `meeting.tsx` `BADGE_STATUSES` share the live set's vocabulary and answer different questions. Folding them in would assert a unification nobody decided.

**The Python↔TypeScript seam, stated as one.** For the cross-language facts there is no shared module to import (ADR-0036 item 10). The manifest compares the regex/literal **source string**. Two engines can read one source differently, so this is the strongest check available without a mechanism that does not exist — and strictly stronger than the nothing there is today.

- **C1 — the checker.** Ran: a first manifest keyed on `os.getenv`-style single anchors. Saw: patterns that matched twice (five config contracts carry the placeholder list once per key) and patterns that matched nothing (a comment that had been reworded). Concluded: both must be **failures**, not tolerated — a manifest that has quietly stopped describing the tree reports green, which is worse than no manifest. Added `"all": true` for the legitimately-repeated case, which also catches a file whose own copies disagree.
- **C1 (2) — prose.** Ran: a byte comparison over the visibility sentence. Saw: five spellings where a reader sees one — a Python implicit-concat splits at whichever word hit the margin, markdown hard-wraps and prefixes `> `. Concluded: compare the sentence with the carrier stripped, but **never** normalise the apostrophe — U+2019 vs U+0027 is a real difference in text a stranger reads, and the rig's copy differs in exactly that way.
- **C2 — the marks.** Ran: `grep` for the two literals, then read the three Dockerfiles. Saw: every copy carried a comment saying the duplication was deliberate because the sides ship in different images — and `core/agent/shared` is COPY'd into agent-api, the worker **and** lite, with two of the three modules already importing from it at module scope. Concluded: the stated reason was true of the TypeScript copy and never of the Python ones. Single-sourced the Python side; the TS copy stays and is now gated instead of hoped about.
- **C2 (2) — the placeholder list, an assumption that did not survive.** Ran: a set comparison across all ten sites. Saw: they **agree** — value- and order-identical. Concluded: this is not a refactor candidate. There is nothing in `deploy/contracts/config.v1` to import from (the schema types `forbidden_values` as an unconstrained string array), so extracting a shared artefact is a contract change and a design choice. Enforced it instead, which is the moment it costs nothing and buys everything.
- **C3 — the ledger, and the red-first pass that caught a real hole.** Ran: the three negative controls. Saw: two went red; the third did not. Flipping `core/agent`'s runtime hostname from `runtime-api` to `runtime` left **both** values present, so the value-set baseline saw no change while a site had just switched camps. Concluded: a ledger entry that cannot say **who** gives **which** answer is not a baseline. Re-pinned on the mapping and added the test; the same control now reds and names the file that moved. That commit is separate and deliberate — the gate was wrong until it went red on the input it exists to catch.
- **C4 — the suite.** Ran: `time node scripts/gates.mjs fact-parity` → **0.11 s**. Concluded: it clears the pre-push wall-clock bar with room to spare.
- **Not ours, untouched.** The `vexa-compose-gate` compose project running on bbb belongs to `~/dev/wt-fc` (branch `flows-compose`). This work ran no Docker gate and started no container.

## Acceptance floor

| Row | Evidence |
|---|---|
| A1 | `node scripts/gates.mjs fact-parity` at head → `✓ 4 enforced fact(s) agree across 19 site(s) · 7 on the drift ledger`. **RED control:** one letter changed in the terminal's `MACHINERY_MARK` → exit 1, `machinery-mark … is written 2 times and they DISAGREE (2 answers): (1) [vexa-machinary] clients/terminal/src/canvas/actions.ts:96 (2) [vexa-machinery] core/agent/shared/marks.py:37`, plus the canonical file and how to agree. Restored. |
| A2 | **RED control:** `clients/terminal/src/_stray.ts` containing the mark → `machinery-mark: clients/terminal/src/_stray.ts:1 writes "[vexa-machinery]" and is not a declared site — read it from core/agent/shared/marks.py instead, or add the site … and say why a third writer is right.` Deleted. |
| A3 | **RED control:** `core/agent/shared/config.py` switched from `runtime-api` to `runtime` → red, printing recorded-vs-found **with the files under each answer**, so the moved site is visible. Both values still existed; the earlier value-set baseline was green on this exact input, which is why the mapping is pinned. Restored. |
| A4 | `node --test scripts/check-parity.test.mjs` → 15/15. "a pattern that matches NOTHING fails" and "a pattern that matches TWICE fails" are two of them; `"all": true` has its own pair, including a file whose own copies disagree. |
| A5 | Same run: prose passes a Python concat split, a markdown hard wrap and a blockquote of one sentence; the identical fixture with U+2019 is **red**. |
| A6 | `grep -rn '"\[vexa-machinery\]"\|"\[vexa-phase:writeback\]"' --include=*.py core/agent \| grep -v tests` → **no output**. `core/agent` pytest → **1588 passed**, including `test_chat_intents.py` and `test_user_text_field.py`, which pin those literals. |
| A7 | `time node scripts/gates.mjs fact-parity` → `real 0m0.113s`. |
| A- | No-regression at head: 19 static (non-Docker) gates green (`readme · docs-version · isolation · exports · graph · schema · contract-version · config-contract · db-schema · db-budget · dataflow · licenses · isolation-py · graph-py · test-isolation · execution-env · contract-conformance · domain-doors · fact-parity`); `node --test scripts/*.test.mjs release/*.test.mjs` → **170 pass / 0 fail**; `core/agent` pytest → **1588 passed**. Docker gates not run — they are CI's, and bbb is shared. |

## Docs diff (D6c)

- `docs/docs/governance/architecture.mdx` § 4 — the `gate:fact-parity` row (the table's own rule is that every gate in the `GATES` map appears there), plus a paragraph at concept altitude saying what it is for: **a duplicated fact needs a gate, not a comment** — four places in this tree assert their copies are "kept identical on purpose" and in each case the copies had already diverged.
- `docs/adr/0037-…` § Consequences — **the config-key rule recorded at the #1483 merge**, contract doctrine rather than a code change: `VEXA_FLOWS_DB_URL` is `required-explicit` because a service without its database is not a deployment; the mail keys stay a `capability` (`mailbox`) because a deployment may legitimately run without mail credentials — the dogfood stack's flows lanes run a mail double (SMTP to mailpit, no password by construction) and `required-explicit` would refuse their boot. Stated generally: **a key is `required-explicit` only if no valid deployment can lack it; anything a valid deployment may omit is a `capability` with a declared degrade.** The test is not "is this important" — the mail credential is important — it is "does a deployment that lacks this still make sense".
- No user-facing page changes: nothing a user of Vexa does changes. The reader here is a contributor.

- **Dependency/licence:** no dependency added — the checker is `node:fs` / `node:path` / `node:url` only; the sole `package.json` change is one `scripts` entry. `pnpm gate:licenses` at head → 514 deps OSS-clean, unchanged.
- **Secrets:** the diff adds no value of any kind. `scripts/parity.json` holds repo paths, regexes, env-var and constant NAMES, and two published-placeholder lists that are already in five committed contracts — no credential. GitGuardian runs on the diff. Note the one security-adjacent **finding**, recorded on the ledger and not changed here: `INTERNAL_API_SECRET` has three boot behaviours across seven contracts, and `core/flows/src/flows_integrations/flows_api.py` still carries a four-value pre-F95 refusal list missing both `vexa-internal-secret` and `lite-internal-secret`.
- **SAST-shaped:** the new code executes nothing, spawns nothing and opens no socket — it reads files and compares strings. The one production change is two Python modules importing a constant instead of retyping it, covered by 1588 tests. `gate:config-contract`, `gate:db-schema` and `gate:execution-env` (P14 — secrets are references only) green at head.

## Validation request

Any competent non-author with a terminal. **The gate is the cheap part; the ledger is what needs you.**

1. `node scripts/gates.mjs fact-parity` at head → green.
2. `node scripts/check-parity.mjs --ledger` → read the seven. The question is not *are these real* — the file:line evidence settles that — but **is each `decision` field the right question, and is anything missing from a fact's site list?** A fact whose sites are incomplete under-reports its own drift, and that is the failure mode this gate cannot catch by itself.
3. Change one letter of `MACHINERY_MARK` in `clients/terminal/src/canvas/actions.ts`, run it again, watch it name both answers and both files. `git checkout` the file.
4. `time node scripts/gates.mjs fact-parity` → confirm it is under 2 s on your machine; it is in your pre-push path now.
5. Sanity-check the one production change: `core/agent/shared/marks.py` and its four importers, against `core/agent/services/agent-api/Dockerfile`, `core/agent/worker/Dockerfile` and `deploy/lite/Dockerfile.lite` — all three COPY `core/agent/shared`.

**Deployment validated (D12b): none, and deliberately.** A static CI gate that runs no service and starts no container behaves identically everywhere because it runs nowhere. Build provenance for every run above: branch `parity-gates` at head, based on `minutes-mcp-viewer` `69d5ba3a2`, fresh `git worktree` on the bbb host, `pnpm install --frozen-lockfile`, `uv run pytest` for `core/agent`, no env deltas from stock. Lite, compose, k8s and hosted are honestly unclaimed: nobody ran them, and this change cannot reach them.

## Authorship
Tooling disclosure (optional, welcome, never an attribution): drafted with Claude Code.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -17,7 +17,7 @@ unset GIT_DIR GIT_WORK_TREE GIT_INDEX_FILE GIT_OBJECT_DIRECTORY GIT_COMMON_DIR G
 # will prove authoritatively on the same sha. The stack/python/build classes cost 5-10 min per
 # push and once blocked a release tag on a LOCAL stale-docker flake. Full suite on demand:
 # `pnpm gates`. Bypass: `git push --no-verify`.
-FAST_GATES="readme dataflow isolation isolation-py exports graph graph-py schema contract-version config-contract domain-doors licenses execution-env test-isolation contract-conformance"
+FAST_GATES="readme dataflow isolation isolation-py exports graph graph-py schema contract-version config-contract domain-doors fact-parity licenses execution-env test-isolation contract-conformance"
 echo "▶ pre-push → static gates ($(echo $FAST_GATES | wc -w | tr -d ' ') fast checks; CI runs the full suite)…"
 for g in $FAST_GATES; do
   node scripts/gates.mjs "$g" || {
--- a/.github/workflows/gates.yml
+++ b/.github/workflows/gates.yml
@@ -40,6 +40,7 @@ jobs:
       - run: pnpm gate:contract-version
       - run: pnpm gate:config-contract
       - run: pnpm gate:domain-doors
+      - run: pnpm gate:fact-parity
       - run: pnpm gate:db-schema
       - run: pnpm gate:db-budget
       - run: pnpm gate:dataflow
--- a/core/agent/control_plane/chat_intents.py
+++ b/core/agent/control_plane/chat_intents.py
@@ -45,8 +45,9 @@ SILENT_KINDS = frozenset({"highlight"})
 # real answer of every scaffolded chat. `worker/engine.py` says exactly this at its own definition —
 # "Distinct from MACHINERY_MARK and not a replacement for it" — and it is the whole reason there are
 # two literals rather than one flag.
-MACHINERY_MARK = "[vexa-machinery]"
-PHASE_MARK = "[vexa-phase:writeback]"
+# ONE SOURCE, three images (shared/marks.py). Re-exported under the names this module has always
+# published, so every reader of `chat_intents.MACHINERY_MARK` is unmoved.
+from shared.marks import MACHINERY_MARK, PHASE_MARK  # noqa: E402 — re-export, see the note above
 SILENT_PREFIX = MACHINERY_MARK + " " + PHASE_MARK + " "
 
 
--- a/core/agent/control_plane/scaffolds.py
+++ b/core/agent/control_plane/scaffolds.py
@@ -89,12 +89,12 @@ TTL_SECONDS = 14 * 24 * 3600
 # did not type a paragraph of instructions, and on 2026-09-02 the founder saw exactly that paragraph
 # painted as his own chat message (ledger F7). The terminal already filters a user turn carrying
 # this mark out of what it renders (`clients/terminal/src/canvas/actions.ts` MACHINERY_MARK /
-# MACHINERY_NOTE); the string is duplicated here rather than shared because the two sides ship in
-# different images, and it is the SAME string on purpose — a rename on either side is a visible
-# regression in the other, which is the cheapest coupling available.
+# MACHINERY_NOTE). The literal now lives once, in `shared/marks.py`, which all three Python images
+# already carry; the terminal's TypeScript copy is the one that genuinely cannot import it, and
+# `gate:fact-parity` compares the two on every push instead of hoping a rename is noticed.
 #
 # "the human sees turns, the agent sees instructions" (founder, 2026-09-02).
-MACHINERY_MARK = "[vexa-machinery]"
+from shared.marks import MACHINERY_MARK  # noqa: E402 — re-export under this module's long-standing name
 MACHINERY_NOTE = (
     "\n\n" + MACHINERY_MARK + " This opening was composed by the product from the link this person "
     "clicked; they did not type it and they cannot see it. Answer it as their first ask, in your own "
--- a/core/agent/control_plane/workspace_reader.py
+++ b/core/agent/control_plane/workspace_reader.py
@@ -102,11 +102,11 @@ _TEMPLATE_BANNER = (
 _TURNS_SIDECAR = "{session}.turns.jsonl"
 
 # The write-back phase runs in the SAME harness session as the turn it follows, so its prompt and
-# its reply are in this transcript. The phase declares itself with this mark (``engine.WRITEBACK_MARK``
-# — duplicated, not imported: the worker ships in its own image; ``tests/test_user_text_field.py``
-# pins the two literals together). Everything from a marked prompt until the next thing a person
-# actually said is bookkeeping the founder was never meant to read back as his own conversation.
-PHASE_MARK = "[vexa-phase:writeback]"
+# its reply are in this transcript. The phase declares itself with this mark — ONE literal now, in
+# ``shared/marks.py``; the worker reads the same constant under its historical name WRITEBACK_MARK.
+# Everything from a marked prompt until the next thing a person actually said is bookkeeping the
+# founder was never meant to read back as his own conversation.
+from shared.marks import PHASE_MARK  # noqa: E402 — re-export under this module's long-standing name
 
 # The harness's own auto-continue. It is a user line nobody typed — the runner nudging a turn that
 # stopped early — and it rendered as a grey USER bubble reading "Continue from where you left off."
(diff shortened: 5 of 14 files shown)


DmitriyG228:
Integrated by branch merge, not through this PR (line-integration ruling 2026-09-03). GitHub auto-closed this as merged when `ffe47682c` became reachable from `minutes-mcp-viewer` via **`63b9cddda`** — `merge: parity-gates — delivers #1491`, a `--no-ff` merge of the branch onto the line. Suites and counts on #1491.

PR #1498, opened 2026-09-03
[DEV] flows: split agent-only flow definitions out of production.py

DmitriyG228:
`core/flows/src/flows_defs/production.py` registered seven flows in one `build(reg, db)`. Four of them have no meaning in a deployment without the agent domain, so a `no-agents` cut had to edit them out of eighteen hundred lines. They are now a separate module, registered behind the presence signal that already exists.

## The split, and the property it splits on

Not tidiness — **whether the flow still does anything when the agent domain is absent** (PRD decisions 40.6 / 40.7).

| | Flows | With no agent domain |
|---|---|---|
| `production.py` **keeps** | `invite_intake` (v3) · `post_meeting` (v4) · `live_meeting` (v1) | they still run: the invite is accepted, the bot joins, the meeting is recorded, and the agent-reaching steps answer `agent:not_present` — 40.7 working |
| `production_agent.py` **takes** | `meeting_prep` (v1) · `email_chat` (v1) · `desk_setup` (v1) · `desk_claim` (v1) | **not registered at all** |

The four degrade to nothing. `desk_setup` and `desk_claim` react to events only agent-api publishes, so in that deployment the fact never arrives either; the other two would exist to write a queue row saying "there is no agent here", per subject, forever.

## What moved

**Steps** — `prepare_meeting` · `feedback_turn` · `email_reply` · `await_scaffold` · `await_claim`.
**Helpers** — `_untrusted_mail` (only `feedback_turn` calls it) and the constant `CLAIM_BOOK` (only `await_claim` reads it).

**Promoted to module level in `production.py` rather than moved** — `_their_clock` and `_scaffold_refs`, which were `build()` closures. Both halves mint scaffolds, and a recipe with two homes is the defect `_note_path` was written to close; they read nothing from the closure, so promoting them changes no name resolution. `platform_user_id` now has no caller in `production.py` and is kept imported there deliberately (see below).

**Deliberately NOT moved** — `_conversation` and the four steps it registers (`open_person`, `drive_person`, `open_group`, `drive_group`). They are agent-only, but they belong to `onboard_person`/`onboard_group`, retired by decision 29, and to none of the seven flows. Moving them was outside this change's rule ("every step/helper that ONLY those four use"); they stay registered and harmless. Worth a follow-up.

**Also not moved** — the `EventType` constants (`UPCOMING`, `MAIL_REPLY`, `DESK_UNSCAFFOLDED`, `CLAIM_PROPOSED`). They are the module's published vocabulary, `production_agent` reads them back through it, and `core/agent/tests/test_desk_events.py` pins `flows_defs/production.py` as the one file under `core/flows/src` that may spell `desk.unscaffolded` / `claim.proposed` at all. Moving them would have forced a change in another domain's test.

## The presence signal

**`flows_steps.common.domain_present("agent")`**, defined at `core/flows/src/flows_steps/common.py:38`. It is the same predicate the engine consults for every `needs=("agent",)` step, reading the same configuration key (`VEXA_FLOWS_AGENT_API_URL`, class `capability`, no default) and never probing — a health check would make "agent-api is restarting" and "there is no agent-api" the same answer, and the second is a shipped product.

The seam is `production._register_agent_flows`, called last in `build()`:

- registers nothing unless the domain is present;
- `importlib.util.find_spec` first, so **a cut that deletes `production_agent.py` outright is supported** — a missing optional module is a fact about the tree, a broken one is a defect, and `find_spec` keeps the two distinguishable where `except ImportError` would not;
- hands `production_agent.build` **this module object** (`home=`), because the moved steps read every shared collaborator through it.

Verified both ways on the branch:

```
agent ABSENT   flows: ['invite_intake', 'live_meeting', 'post_meeting']   prepare_meeting in vocab: False
agent PRESENT  flows: ['desk_claim', 'desk_setup', 'email_chat', 'invite_intake', 'live_meeting', 'meeting_prep', 'post_meeting']
FILE DELETED, agent PRESENT -> flows: ['invite_intake', 'live_meeting', 'post_meeting']
```

## One finding worth reading before the next module split

`production_agent` must read `setting`, `mint_scaffold`, `scaffolded`, `ws_file`, `ag`, `mt` and the event types **through the production module**, not `from .production import …` — the idiom `common.agent_door` states out loud (*"Reads the MODULE attribute so a test can set the world with one `monkeypatch.setattr`"*). A `from … import` binds the real functions past every fake the suite installs on `production`.

That was not sufficient. A module-scope `from . import production` resolves `sys.modules["flows_defs.production"]` **once, at import** — and `tests/test_flows_api_service.py:180` deletes every `flows_defs.*` and `flows_steps.*` entry from `sys.modules` mid-run. After it, a re-import builds a **second** production module while every already-collected test module still holds the first: the fakes land on one object, the steps read the other, and twelve tests reached a real socket at `127.0.0.1:1`. Every one of them passed in isolation.

`sys.modules[__name__]` at call time does not fix it either — it resolves to whatever module currently answers to the name, which is the second one. The module object is captured at **import** (`_SELF = _sys.modules[__name__]`, where the interpreter guarantees it is self) and passed down. Anything that splits a module the suite monkeypatches will hit this.

## Tests

**525 passed / 5 skipped, before and after** — `uv run pytest -q` in `core/flows`, on bbb, base `88b158103` vs head `261d2e68c`. **No test file changed.** Every registry assertion that pins the full vocabulary — `test_no_agents.AGENT_STEPS`, `test_no_meetings.MEETINGS_STEPS`, `test_queue_waiting`'s desk-card flows, `test_link_loop`'s `prepare_meeting` cases, `test_untrusted_mail` — goes on reading it from `production.build(reg, db)`, which now chains, and the suite declares the agent door in `conftest.OFFLINE_DOORS`.

Gates, from the worktree root on bbb (all static, no containers): `isolation-py` ✓ · `graph-py` ✓ · `test-isolation` ✓ · `exports` ✓. The pre-push hook's sixteen static gates are green too.

## Nothing else changed

Three files, all under `core/flows/`: `src/flows_defs/production.py`, the new `src/flows_defs/production_agent.py`, and `src/flows_defs/README.md`.

## Undone

- `_conversation` and its four retired onboarding steps are still in `production.py` (above).
- `core/agent/control_plane/claims.py:27` has a comment pointing at "`flows_defs/production.py` CLAIM_BOOK". The constant is in `production_agent.py` now; the comment is one word stale and was left alone rather than touch another domain in this PR.

### DIFF
--- a/core/flows/src/flows_defs/README.md
+++ b/core/flows/src/flows_defs/README.md
@@ -5,3 +5,22 @@ FUNCTIONS (a typo is a registration error, never a 2pm KeyError; strings exist o
 database). One file per flow; reviewed like any product change; a new version is a new
 registration — in-flight reactions keep the version stamped at admission, new events select the
 newest (Registry.match).
+
+## `production.py` and `production_agent.py`
+
+The production definitions are in TWO modules, split by one property: whether the flow still does
+anything when the **agent domain is not deployed** (PRD decisions 40.6/40.7).
+
+| | Flows | With no agent domain |
+|---|---|---|
+| `production.py` | `invite_intake` · `post_meeting` · `live_meeting` | they still run — the invite is accepted, the bot joins, the meeting is recorded, and the agent-reaching steps answer `agent:not_present` |
+| `production_agent.py` | `meeting_prep` · `email_chat` · `desk_setup` · `desk_claim` | **not registered at all** — a conversation with an agent and two cards on a desk have nothing to degrade to |
+
+`production.build()` calls `production_agent.build()` last, and only when
+`flows_steps.common.domain_present("agent")` — the same predicate the engine consults for every
+`needs=("agent",)` step, reading the same key (`VEXA_FLOWS_AGENT_API_URL`) and never probing. A cut
+that deletes `production_agent.py` outright is supported: the seam checks `find_spec` first.
+
+Shared helpers stay in `production.py`, and `production_agent` reads every collaborator **through
+the module object handed to its `build(reg, db, home=…)`** — never `from .production import …`. One
+`monkeypatch.setattr(production, …)` has to reach both halves.
--- a/core/flows/src/flows_defs/production.py
+++ b/core/flows/src/flows_defs/production.py
@@ -50,6 +50,7 @@ from __future__ import annotations
 import json
 import logging
 import os
+import sys as _sys
 import time
 from pathlib import Path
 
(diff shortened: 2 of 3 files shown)


DmitriyG228:
Tracked by [#1499](https://github.com/Vexa-ai/vexa/issues/1499) — the D3 issue for the no-agents cut this split unblocks.

DmitriyG228:
Merged manually as part of the merge worker's dispatch (worktree merge, not GitHub's merge button, since the target `minutes-mcp-viewer` isn't `main`) — content landed at `4d3448074` on `minutes-mcp-viewer`, with one merge-time test fix on top (see [issue #1499](https://github.com/Vexa-ai/vexa/issues/1499#issuecomment-5526650237) for the full judgment + suite results). Closing this PR as superseded by that merge.

PR #1543, opened 2026-09-04
feat(transcription-service): per-window language re-detection (AIM-949)

ludger-aimable:
## Why
Capture's batch transcript sends the whole recording as one request with no language, so faster-whisper detects the language once on the first 30 s and decodes everything in it. A MavenBlue meeting (Vexa meeting 118) that opened in English and continued in Dutch came out as 45 minutes of translated, garbled English.

## What
- `multilingual=True` on the transcribe call: language is re-detected per 30 s window from that window's encoder output. Skipped automatically when a request pins `language`.
- Env flag `MULTILINGUAL` (default `true`) to turn it off; logged at startup; documented in `.env.example`.
- `faster-whisper>=1.1.0` (the parameter exists since 1.1; pii runs 1.2.1).

## Measured (pii, int8, prod decode params, 3 meetings)
| meeting 118, English word share | whole-file (today) | per-window |
|---|---|---|
| 0–5 min, really English | 100% | 100% |
| 5–60 min, really Dutch | 81–100% | 0–5% |

Two Dutch-only meetings: output word-for-word identical. Wall time 57 s → 63 s on 76 min. Pinning `nl` instead translates the English opening into Dutch, so that is not the fix.

Linear: AIM-949

### DIFF
--- a/deploy/compose/docker-compose.yml
+++ b/deploy/compose/docker-compose.yml
@@ -19,9 +19,13 @@ services:
 
   postgres:
     image: postgres:17-alpine
-    # idle_in_transaction_session_timeout=60s auto-kills orphaned transactions
-    # (defense-in-depth against future DB session leaks in meeting-api).
-    command: ["postgres", "-c", "idle_in_transaction_session_timeout=60000"]
+    # 60s was too aggressive: /meetings/{id}/transcribe holds the request's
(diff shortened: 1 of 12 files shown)


ludger-aimable:
Opened against the wrong repo by mistake, sorry for the noise. This belongs on the aimable-ai fork.

PR #1546, opened 2026-09-04
mcp: standing notices ride meeting tool results; flows: notice-flagged queue items

DmitriyG228:
A say-file under `behavior/queue/` may now open with `notice: true` in front-matter.
That marks what stays true **between** calls rather than what just happened — an
admin's edit in the tree that already decides whether an item is spoken at all, with
no deploy on either side of it.

**flows** — `GET /queue/waiting` carries `notice` on each item; `GET /queue/notices`
is a smaller answer behind the same door: only the flagged say texts, deduped, cheap
enough to ask on every call.

**mcp** — the five meeting tools carry the caller's notices out on their results: a
`notices` field plus one `Notice: …` line, once. Bounded at 2s; an absent or slow
domain, a refusal, or an unreadable answer leaves the result exactly as it was.

No new tool (still 21), no new argument, and nothing flagged in this repo.

Draft: opened for review, not for merge.

### DIFF
--- a/behavior/queue/README.md
+++ b/behavior/queue/README.md
@@ -34,6 +34,32 @@ So: adding a file is how a situation becomes person-facing, and deleting one is 
 are commits here, and neither is a deploy. The count of silent reactions comes back as `quiet`, so
 an operator can always tell *nothing is happening* from *behavior is saying nothing about it*.
 
+## Front-matter: `notice: true`
+
+A file may open with a fenced block of `key: value` lines. There is one key:
+
+```markdown
+---
+notice: true
+---
+Whatever a person's agent should keep in front of it.
+```
+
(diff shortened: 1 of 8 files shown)


DmitriyG228:
Baseline note, not caused by this branch: `gate:dataflow` already fails on the base commit `acd7a25c0` — *completeness: 'core/flows/contracts/flows.v1' exists on disk but is not registered in architecture.calm.json*. Verified by running the gate on a clean worktree of that commit. This branch was pushed with `--no-verify` for that reason; nothing here touches `architecture.calm.json` or the contracts directory.

Suites, both green with no new failures: flows 544 passed (511 on base, +33 new), mcp 279 passed (243 on base, +36 new). The 23 pre-existing flows failures are identical on base and branch.

Issue #1316, opened 2026-08-24
[DEV] Calendar import stores scheduled_at +1h off for some TZID events — dispatch reads the true time, retry/grace windows read the wrong one


DmitriyG228:
Observed on prod 2026-08-24: a calendar-imported occurrence stored `scheduled_at: 10:00:00+00:00` while the meeting verifiably ran at **09:00Z** (bot dispatched 08:58, `active` 08:59:50, transcript rows from 09:00; the humans were there). A sibling occurrence imported the same day stored its time correctly — so the error is per-event, consistent with TZID handling on some ICS events (+1h, DST-shaped).

Two consequences:
1. **Dispatch evidently does not read `scheduled_at`** (it fired at the true start−2min despite the stored value being an hour later) — two sources of truth for one event's time.
2. **Retry/backoff/grace windows DO key on `scheduled_at`** (#1200/#1211 semantics) — for affected events the occurrence window is shifted +1h: retries can fire during the hour after the real meeting ended, and a genuine early failure could be denied its retry because the system thinks the meeting hasn't started.

Expected: one canonical event-time on the row that both dispatch and the retry window read; TZID normalization test for ICS imports covering DST-offset zones.

Issue #1249, opened 2026-08-19
[DEV] BLOCKER: #991 needs one production first-save readback on a fresh identity — human-only


DmitriyG228:
### Value this issue delivers

The one observation that closes #991 against the reporter's actual state, rather than against an account that already worked. Until it is run, "the calendar fix cleared his mismatch" remains an inference — and #991 was already closed once on an inference.

### Why this is blocked on a human

The check needs a hosted identity that has **never had a calendar connected**. That identity cannot be produced by an agent: account creation is a human-only action, and the one production key on hand belongs to a live account whose feed URL is masked on read-back and therefore not restorable after an overwrite.

Prior attempts and why they do not substitute, from the 2026-08-17 production run on #991:

- both runs used an account that **already had a working calendar**, so a first-ever-save fault would not have been caught;
- admin-api is not publicly exposed, this host has no cluster context, and staging tokens 401 against production.

### The exact sequence

Against `api.cloud.vexa.ai`, with a **fresh account that has no calendar**, using a valid published Outlook/M365 ICS feed (a `.ics` that answers 200 with `BEGIN:VCALENDAR` and no redirect):

| # | Call | Expected |
|---|---|---|
| 1 | `PUT /user/calendar` with `{"ics_url": "<feed>"}` | `200`, `ics_url_set: true`, masked host echoed back |
| 2 | `POST /user/calendar/sync` — **immediately**, same API key | **not** `404 no calendar feed connected`. Either `200` with a stamp (`counts`, or a `last_error` naming a *feed* problem), or `503 calendar_config_discovery_unavailable` if #1090 has landed by then |
| 3 | `GET /user/calendar/sync` | a non-empty stamp, not `{}` |
| 4 | repeat 1–2 once | same result both times |

### What each outcome means

- **No 404 at step 2** → the reported mismatch does not reproduce on the reporter's exact path, and #991 can close on an observation instead of an inference.
- **404 at step 2** → the mismatch survives on the first-save path specifically. That is a live production defect on a path the 2026-08-17 run could not reach, and it reopens the incident rather than closing it.
- **Empty `{}` at step 3 after a non-404 step 2** → the stamp is not being stored; a separate defect, worth its own row.

### Constraints

- `auto_join` should be left `false` throughout so nothing joins a real meeting.
- The feed URL is a credential; do not paste it into this issue or any comment. Report only status codes, the masked host, and counts.
- Delete the account's calendar afterwards and confirm the account returns to its prior state.

[...]

Issue #1232, opened 2026-08-18
[DEV] Hostile calendar feeds cannot exhaust the control plane — bound ICS fetch size, time, and VEVENT count


DmitriyG228:
### Value this issue delivers

A hostile or huge calendar feed cannot exhaust control-plane memory — the size and time caps bound the fetch, not just report on it after the fact.

### Finding (rc.18 security review)

**S2 — 2 MB ICS cap fires after full buffering.** The code does `resp.content` then `len(...)`: a 200 MB feed buffers entirely (+530 MB RSS measured) before the cap "refuses" it. `timeout=15` is per-operation, not total, so a slow-drip feed is unbounded in time. Reachable via `POST /user/calendars/{id}/sync` with no cooldown behind a 40 rps/user bucket → OOM of the shared control plane.

**S3 — unbounded VEVENT count.** A feed *under* the 2 MB cap yields ~14,000 planned rows/user, ~8 MB GIN-indexed JSONB, ~1.1s event-loop CPU — every 300s, forever. (#1213 item 3 triages the row *inflation* LOW; the row *count* is the larger half and not in that description — this supersedes/expands it.)

### Fix

Stream the body with a hard byte ceiling that aborts mid-read; add a total-deadline alongside the per-op timeout; cap VEVENT count per feed (reject or truncate with a surfaced error). Sync cooldown per calendar.

### Validation contract

None — machine-observable. Oracle: a 200 MB feed aborts at the ceiling with bounded RSS; a slow-drip feed hits the total deadline; a 14k-VEVENT feed is capped. Red tests reproduce each measurement.

PR #1636, opened 2026-09-06
feat(calendar_sync): read a user's calendar over the Google and Graph APIs, not just an ICS feed

DmitriyG228:
**Draft.** The seam only — nothing calls these yet. Opened as the findable artifact for a pushed
branch rather than as a review request; mark ready on the founder's word.

## COMMITMENTS IN THIS PR

**None.** No published terms, no dates, no rights, no spend, no external promise. Additive code
behind no route and no flag: two new pure functions, their tests, and the module's own docs.
The commitments in the wider calendar-OAuth effort (the privacy-policy amendment, entering Google's
verification programme) live in `~/dev/biz/drafts/2026-09-06-calendar-oauth-review-clock.md` and are
gated separately — none of them is decided here.

## Why

Connecting a calendar today means pasting a secret ICS address. Google Workspace hides that field
under the default admin policy and Outlook requires publishing a calendar, so the setup step for the
buyer we care about is "file a ticket with IT". Six open issues also exist purely because we
dereference a user-supplied URL and parse RFC 5545 ourselves (#1182, #1231, #1232, #1316, #991,
#1249), and a polled feed has a 5-minute latency floor with no push.

## What this is

`calendar_sync` was already split so the pipeline is reader-agnostic: `sync_user` owns one row per
UID, adoption by link, occurrence disposition and retirement, and it consumes a plain dict. So a
calendar API is a **new reader**, not a new pipeline.

```
parse_ics(text)              -> {"events": [...], "cancelled_uids": [...]}
events_from_google(items)    -> {"events": [...], "cancelled_uids": [...]}   NEW
events_from_microsoft(items) -> {"events": [...], "cancelled_uids": [...]}   NEW
```

Pure — no network, no tokens, no clock of their own. Recurrence expansion is the provider's job
(`singleEvents=true`, `/calendarView`); we then apply the same rule `parse_ics` applies: group by the
series-stable id, keep the earliest occurrence in the window, because two scheduled rows on one
native id violate `uq_meeting_active_user_platform_native`.

## The decision worth reviewing

**Keyed on `iCalUID` / `iCalUId`, not on the per-occurrence `event['id']`.** That is the same value
the ICS feed carries, so a calendar reconnected from ICS to OAuth **adopts its existing rows**.
Keying on the instance id would have re-imported every existing user's whole calendar as duplicates
on the day they switch. Executed rather than asserted —
`test_a_calendar_reconnected_over_oauth_adopts_its_own_ics_rows` runs `sync_user` over a feed and
then over the API payload for the same meeting and asserts `created == 0`.

## Two deliberate divergences from the ICS reader

| | ICS reader | These readers | Why |
|---|---|---|---|
| Event snapshot | copies every VEVENT property | bounded allowlist | #1213 item 3 — "no test that the VEVENT snapshot is redacted"; an API returns far more per event than a feed (extended properties, attachments, ACL hints) |
| Naive timestamp | n/a | resolved against its stated zone, else **UTC** | never the server's local zone — that is #1316 and backlog R-B10. Wrong by a known offset beats wrong by wherever the pod runs |

Attendees fold onto the ICS PARTSTAT vocabulary (`accepted` / `declined` / `tentative` /
`needs-action`) so nothing downstream has to learn two dialects; rooms and equipment are dropped on
both sides, matching `service._attendees`.

## Tests

`tests/test_calendar_providers.py`, 28 cases. The parity ones are the point: same PlannedEvent keys
as `parse_ics`, same attendee vocabulary, same one-row-per-uid rule, same "a link-less event still
imports" rule, and three that drive the real `sync_user` over the in-memory store.

```
meeting-api: 1450 passed, 5 skipped   (28 new)
```

Run locally against the repo venv, no containers.

## Not in this PR

No OAuth flow, no token storage, no route, no config-record change, no push subscriptions, no UI.
Nothing imports these functions yet. Token storage in particular has a hard prerequisite: #876 —
per-user secrets sit plaintext in `users.data` JSONB, and a leaked refresh token is a different
category of problem from a leaked ICS URL. Envelope encryption lands before or with the connect flow.

## Second commit: the I/O half

`provider_io.py` — fetch a window of events, refresh a token. `fetch_ics` dereferences a
user-supplied URL and must ride the SSRF-pinned transport; these talk to two fixed hosts, so the
risk inverts and what matters is that stored user input never escapes into the URL:

- calendar ids are path-escaped against a constant base (tested with `../../tokeninfo?x=`)
- redirects off, so an `Authorization` header cannot be moved to another host
- Graph's `@odata.nextLink` is host-checked **before** it is followed with a token attached

Same error contract as `fetch_ics` — `(value, human_reason)`, never raises, because one user's dead
network or revoked grant must not stall every other user's calendar in the same sweep. The reason is
what the calendar panel shows, so it reads "reconnect the calendar", not "401".

An `invalid_grant` refresh is treated as **terminal**, not transient. Retrying a dead grant on a loop
hammers the identity provider with a credential that will never work again.

Scopes are `calendar.events.readonly` + `calendar.calendarlist.readonly` and a test pins that they
stay read-only. We never write to a calendar, and a wider scope is a bigger consent prompt, a slower
Google review, and more to lose if a token leaks.

Pagination is exhaustive but bounded at 20 pages: past ~5000 events in a 14-day window this is not a
person's calendar any more.

17 further tests, offline through a stub client. **meeting-api: 1467 passed, 5 skipped.**

## Note on `value-fsm`

It failed once on this branch (`test_mock_silence_left_alone` → `POST /bots` 500), then **passed on
re-run against the identical head**. Not caused by this change: `meeting_api.calendar_sync` and
`meeting_api.__main__` both import cleanly with the new eager import, and the suite is green locally.
Flagging it because a gate that flakes on a mock scenario is worth someone's attention on its own.

## Collision check

- #1410 (`fix/rrule-expand-in-event-timezone`) touches `service.py` + `test_calendar_sync.py` — no
  overlap with these files. Its timezone reasoning is adjacent to `_iso_to_utc` here; if it changes
  how a naive time is read, these readers should follow.
- #1250 (`1182-ics-size-limit`) also edits this module's `README.md`, at the `fetch_ics` bullet
  (old line ~21). This PR edits the header and adds the reader table above it. Different regions;
  expected to merge cleanly.

---

## Contribution rights

independent path means no individual CLA is required.

> Section restored by an agent because the PR body was written without the repo template. **Nothing
> is ticked and nothing will be** — the template says an agent may explain the choices but must not
> select one, so this is left for a human. The commits also lack `Signed-off-by`, which is the same
> kind of certification; sign them with:
> `git rebase --signoff origin/main && git push --force-with-lease`

### DIFF
--- a/core/meetings/services/meeting-api/src/meeting_api/calendar_sync/README.md
+++ b/core/meetings/services/meeting-api/src/meeting_api/calendar_sync/README.md
@@ -1,10 +1,25 @@
-# calendar_sync — ICS feed → planned meetings
+# calendar_sync — a user's calendar → planned meetings
 
-One concern: turn a user's secret ICS calendar feed into PLANNED meeting rows (intent status
-`scheduled`, complete `data.calendar_sources[].event` provenance) so the Meetings surface shows what's coming and the
-auto-join sweep sends the bot when each meeting starts. No OAuth — the user pastes the
-secret-address ICS URL Google Calendar / Outlook already provide (`PUT /user/calendar`, identity
-domain).
+One concern: turn a user's calendar into PLANNED meeting rows (intent status `scheduled`, complete
+`data.calendar_sources[].event` provenance) so the Meetings surface shows what's coming and the
+auto-join sweep sends the bot when each meeting starts.
+
+**How we learn the calendar is a READER, and there are three.** Everything below the reader — one
+row per UID, adoption by link, occurrence disposition, retirement — is identical in every case,
+because every reader emits the same `{"events": [PlannedEvent], "cancelled_uids": [uid]}`:
+
+| Reader | Input | The caller must |
+|---|---|---|
+| `parse_ics` | secret-address ICS feed text | — (expands RRULEs itself) |
+| `events_from_google` | Google Calendar `events.list` items | pass `singleEvents=true` + `showDeleted=true` |
+| `events_from_microsoft` | Graph `/calendarView` items | send `Prefer: outlook.timezone="UTC"` |
+
+The two API readers key on the series-stable UID (`iCalUID` / `iCalUId`) — the same value the feed
+carries — so **a calendar reconnected from ICS to OAuth adopts its existing rows instead of
+duplicating them.** That migration path is executed in `tests/test_calendar_providers.py`, not
+merely asserted. They also store a BOUNDED per-event snapshot rather than the ICS reader's
+unbounded property copy (`#1213` item 3), and they never read a naive timestamp in the server's
+local zone (`#1316`).
 
 ## Public surface
 - `parse_ics(text, now, horizon_days=14)` → `{"events": [PlannedEvent], "cancelled_uids": […]}`
--- a/core/meetings/services/meeting-api/src/meeting_api/calendar_sync/__init__.py
+++ b/core/meetings/services/meeting-api/src/meeting_api/calendar_sync/__init__.py
@@ -1,10 +1,13 @@
-"""calendar_sync — ICS feed → planned meetings (see README.md).
+"""calendar_sync — a user's calendar → planned meetings (see README.md).
 
-Public surface: ``parse_ics`` / ``sync_user`` (pure logic), the production I/O adapters
-``fetch_ics`` / ``fetch_configs``, and the shared one-user pass ``run_user_sync`` (+ stamp
+Public surface: the READERS ``parse_ics`` / ``events_from_google`` / ``events_from_microsoft``,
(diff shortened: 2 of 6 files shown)


DmitriyG228:
`docs: none` reason (D6c): this PR adds two pure functions that **nothing imports yet** — no route, no config field, no flag, no user-visible behaviour. The module's own `README.md` is updated in the diff. User-facing docs (`docs/docs/how-to/calendar-sync.mdx`, which today documents the secret-ICS-address flow) change when the connect flow ships, not here — documenting an OAuth option that does not exist would be the wrong kind of current.

Issue #1670, opened 2026-09-13
Bug: Meeting Capture Status


timjuedemann:
### What happened

In the terminal's **Meetings** view, every past meeting carries the phrase **"nothing captured"** —
including meetings that recorded audio and have transcript segments. The little line to the right of
the row (duration · notes/lines) stays empty for the same rows.

**Expected:** a meeting whose audio was captured reads as captured, and the row shows its
duration/notes.
**Actual:** it reads "nothing captured" and the right-hand detail is blank.

This is not a display-only nicety: "nothing captured" is the state that tells a user the bot failed
to record. A user who sees it on a meeting they know was transcribed learns to distrust the feed.

### Meeting platform

MS Teams (the affected meeting was a Teams bot run). The defect is platform-independent — it is a
row-projection issue, not a capture issue.

### Deployment

Self-hosted compose (`deploy/compose`, project `vexa-v012`), image tag `v012`. Deployed tag:
**v0.12.27-5-g59e2c413**. Observed there; every `file:line` below was re-verified against `main` at
`59e2c41` (2026-09-13), where the code is unchanged.

### Logs, screenshots, meeting shape

**Observed on our instance** — one past meeting, `status: completed`, 14 transcript segments,
`data.segments_captured: 14`, `data.recording_enabled: true`, and a `recordings[]` entry (one
741 KB webm, `status: completed`). Two calls, same row:

| Endpoint | `data.recordings` | What the UI computes |
|---|---|---|
| `GET /meetings` (the list the view renders) | **absent** | `has_recording = false` → "nothing captured" |
| `GET /meetings/{id}` (detail) | **present** | would be `true` |

**Where we are** — the client asks for a key the list is designed not to send:

- `clients/terminal/src/surfaces/liveMeetings.ts:192` derives the flag from the omitted key:
  `has_recording: !!(d.data?.recordings?.length),`
  The DTO it fills (`liveMeetings.ts:22`) and the file's own header comment (`liveMeetings.ts:4`,
  `data:{recordings:[...]}`) both assume the list carries `recordings`.
- `clients/terminal/src/surfaces/today.tsx:252-254` turns that flag into the phrase:
  `run.has_recording ? (unreviewed ? "recap ready" : null) : "nothing captured"`.
- `clients/terminal/src/surfaces/today.tsx:227` uses the same flag to bail out of the right-hand
  line entirely — `if (!m.has_recording) { setLine(""); return; }` — which is why the duration/notes
[...]

Issue #584, opened 2026-07-13
Meetings list (GET /bots, GET /meetings) embeds full meeting.data per row, unpaginated → 4.6 MB /meetings page


DmitriyG228:
### Why this matters

**This mechanism caused a ~1.5-hour hosted read outage on 2026-07-15 (07:30–09:04 UTC).** Under morning
load, heavy `/meetings` serialization wedged the meeting-api event loop → ~100% read failure on
`/meetings` and `/transcripts`. A restart plus a DB-pool bump did **not** hold (the wedge is
event-loop-bound, not pool-bound); it was mitigated only by scaling meeting-api to 2 replicas — which in
turn activates the multi-replica caveats now filed as #636 / #637. This issue is the durable cure: fix it
and prod can scale back to a single replica.

The meetings-list endpoint (`GET /bots`, and its twin `GET /meetings`) returns **every one of the
caller's meetings, each embedding its full `data` JSONB, unpaginated** — so the dashboard's `/meetings`
page downloads and parses megabytes of per-meeting transcript/log/recording content it never displays in
a list. Measured on a real account with 583 meetings: the list response is **4.6 MB**, and even
`?limit=20` is **2.8 MB**, because recent meetings carry multi-hundred-KB `data` blobs. The page is slow
not at the DB (630 ms) but at the browser parsing 4.6 MB of JSON — and, under concurrency, at the
meeting-api serializing it on the event loop.

The list view needs only lightweight metadata (id, platform, native id, status, times, a title). It does
**not** need `speaker_events`, `bot_logs`, `recordings`, chat, or error dumps — those belong to the
meeting-detail view. Today they ride along on every row.

### Value this issue delivers

> The meetings list loads fast and can't wedge the API: the list endpoint returns lightweight rows (a few
> KB total, bounded by a default page size), not every meeting's embedded transcript/logs/recordings.

### Where we are (honest)

Line references verified against tag **`v0.12.4`**.

- `core/meetings/services/meeting-api/src/meeting_api/collector/adapters.py:388` — `list_meetings` does
  `select(Meeting)` (the whole row, line 403) and returns `"data": m.data` verbatim per row (line 426).
- `core/meetings/services/meeting-api/src/meeting_api/collector/app.py:207` — `GET /bots` (the dashboard's
  primary list source) and `app.py:180` — `GET /meetings`. Both accept `limit`/`offset`
  (`app.py:185-186`, `211-212`: `Query(default=None, ge=1, le=100)`).
[...]

PR #1677, opened 2026-09-17
fix(meetings): a captured meeting no longer reads "nothing captured" — the list row carries the answer (#1670)

DmitriyG228:
**Refs #1670**

## Contribution rights

## The defect, restated

**What the user saw:** in the terminal's Meetings view, every past meeting row read **"nothing captured"**, and the duration/notes line to the right of the row was blank — including a completed MS Teams meeting with 14 transcript segments and a 741 KB recording.

**What they expected:** a meeting whose audio was captured reads as captured, and the row shows its duration and notes.

**Which surface:** the **meetings LIST payload** (`GET /meetings` / `GET /bots`), rendered by the terminal. Not the dashboard, not MCP, and not the detail endpoint — `GET /meetings/{id}` was always correct, which is precisely what made the two reads disagree.

**Why it matters beyond cosmetics:** "nothing captured" is a **failure report**. It is the string that tells a user the bot did not record. Showing it on a meeting the user knows was transcribed teaches them to distrust the feed.

**Where it is born.** `LIST_OMIT_KEYS` drops `data.recordings` from every list row — correct, and it stays: it is the #584 fix for a 4.6 MB list response that wedged the event loop. But `recordings` was also the only capture **evidence** a list row carried. The client derived the flag from the omitted key (`liveMeetings.ts:192`, `!!(d.data?.recordings?.length)`), so it read `undefined` on every row, and `today.tsx:252` / `:227` turned that one `false` into both symptoms — the phrase and the blank line.

The reporter's diagnosis was accurate end to end, including their reading of the `LIST_OMIT_KEYS` comment (it promises the flags "ride along", and `recording_enabled` does — but that is the **setting**, written at spawn time, not evidence that audio arrived). Their suggested direction is what this PR implements.

## The fix

At the producer, per "fix at the point of introduction": a key dropped for its **size** still owes the row the **question** it answered.

- `projection.has_capture(data)` — one documented pure function. True when the meeting has at least one stored recording **or** a non-zero `segments_captured`.
- Both row builders (`adapters._row`, `fakes._row`) hoist it as a top-level `has_capture` boolean, beside `constructed_meeting_url`, `completion_reason` and `failure_stage`. Because both list and detail rows come through `_row`, **the two reads now agree by construction.**
- The segment half is deliberately the *same test* `lifecycle/provenance.py:77` uses to rate `transcription_outcome == "served"`. The API's own rating already called the reported meeting served while the list called it empty; those two can no longer diverge.
- `recording_enabled` is explicitly **not** evidence, and a negative-control test holds it `true` on an empty row to keep it that way.

At the client, adaptation only: read the scalar, fall back to the evidence the list still carries against an older core. The terminal's flag is renamed `has_recording` → `has_capture` — a flag named for recordings that is true on a transcript-only meeting is the same trap one layer down, and the reporter named that trap explicitly.

## Observation bundle

- **C1 — reproduced from code, no live meeting.** Wrote a test that drives the real terminal store over its real seam (`GET /api/meetings`) with the reported row shape: completed Teams run, `recording_enabled: true`, `segments_captured: 14`, `data.recordings` absent. Ran it in a throwaway worktree at `origin/main` (`dba990b4`), unfixed. Saw:
  ```
  has_recording = false → phrase = nothing captured
  AssertionError: expected false to be true
  ```
  Concluded: the report reproduces exactly, at the client's mapping seam, with no stack and no meeting.

- **C2 — red at the core, on the same base.** Copied the seven new `test_slim_meetings_list.py` cases into that same `origin/main` worktree. Saw `7 failed, 19 passed`, the list-row case failing on `KeyError: 'has_capture'`. Concluded: the row genuinely carries no answer today; the tests are not tautological.

- **C3 — green at head.** `uv run pytest -q` in `core/meetings/services/meeting-api` → **1429 passed, 5 skipped**. The same payload through the terminal store now yields `has_capture = true`.

- **C4 — green at the client, whole suite.** `pnpm exec vitest run` in `clients/terminal` → **59 files, 467 tests passed**; `tsc --noEmit` clean. The 465→467 delta is the two new phrase tests; the five renamed fixtures are covered by the existing 465.

- **C5 — the user-visible string is now pinned.** The phrase was a ternary buried inside `PastRow`, untestable. Extracted it to `pastPhrase(run, unreviewed)` with no behaviour change, so the claim in this issue's title is asserted directly: a captured run reads "recap ready", an empty one still reads "nothing captured". The failure report keeps working as a failure report.

- **C6 — static gates.** All 14 pre-push gates green, including `contract-conformance` (2 services conform to the sealed api.v1). `has_capture` is a hoisted row field like `shared`, not a route change, so the sealed surface is untouched.

## Acceptance floor

#1670 is an incoming report with no acceptance table, so the rows below are the reporter's own claims.

| Row | Evidence |
|---|---|
| A meeting with segments + a recording reads as captured in the list | C1 (red on `dba990b4`) → C3/C5 (green) |
| The right-hand duration/notes line is no longer suppressed | Same flag, same fix — `today.tsx:227` now gates on `has_capture`. **Not separately witnessed live** (see below) |
| The `recordings` omission is preserved | `test_list_row_carries_capture_evidence_the_projection_drops` asserts `"recordings" not in row["data"]` alongside the new scalar |
| Fixed at the projection, not in the terminal | `projection.py` + both row builders; the client change is a read plus a compat fallback |
| List and detail agree | `test_detail_and_list_agree_on_capture` |
| A genuinely empty meeting still reads "nothing captured" | `test_a_meeting_that_captured_nothing_reads_false_though_recording_was_enabled`, and the `pastPhrase` negative case |

## Docs diff (D6c)

- `docs/docs/api/meetings.mdx` § *List rows are slim; detail is full* — documents `has_capture`, and states the two wrong ways to derive capture (`data.recordings`, absent on list rows; `data.recording_enabled`, the setting) so the next integrator does not walk into the same trap.
- `docs/changelog.d/1670-meeting-capture-status.md` — the per-PR fragment.

## What is NOT done

- **No live witness on a running deployment.** Everything above is offline: unit and store-level red→green against `origin/main` and head. The change is a row projection and its client read, both proven at their own seams, but **nobody has yet watched a real bot record a real meeting and seen the row read "recap ready"** — the reporter's own environment (self-hosted compose, `v012`) is the natural place for that.
- **Lite / helm / hosted unexercised**, which the reporter also flagged as unverified. Same code path; the row shapes are the same `_row`.
- **Contribution rights: unticked, and no `Signed-off-by` on the commit.** An agent must not make that certification on a human's behalf (`CONTRIBUTOR_RIGHTS.md`, and the template says so in its own comment). A maintainer needs to tick one box and add the sign-off — `git commit --amend --signoff` on this branch, or the DCO app's remediation commit.

## Validation request

**Preferred validator: @timjuedemann**, the reporter — they have the affected instance and already ran the client-side patch locally. What to watch: on `v012` + this branch, a completed meeting with a transcript reads **"recap ready"** (or goes quiet once opened) and shows its duration/notes line; and a bot that never joined still reads **"nothing captured"**. They are also the right person to say whether the `has_capture` scalar is what they would have wanted to read from their own integration.

## Authorship

---

## Drafted reply for a maintainer to post on #1670

*Not posted. The founder acked on 2026-09-16 and promised an update by 2026-09-19; this is that update, for a human to send.*

**COMMITMENTS IN THIS DRAFT**
1. **Work we will do** — none beyond the PR already open; the reply points at it and does not promise further work.
2. **A date** — none. The reply does not promise a release or a merge date.
3. **A right / precedent** — names @timjuedemann as preferred validator and credits the diagnosis. Standard practice per the delivery book (reporter is the preferred signer of their own fix), not a new precedent.
4. Money in, money out, published terms — none.

> Your diagnosis was right on every hop, including the `LIST_OMIT_KEYS` comment: `recording_enabled` rides along but it is the setting, not the evidence.
>
> Fixed at the projection, as you suggested. The row now carries a top-level `has_capture` — true on a stored recording or a non-zero `segments_captured`, the same test `provenance.py` already uses to rate a meeting `served`. List and detail both come through the same row builder, so they can't disagree again. The terminal reads the scalar and falls back to `segments_captured` against an older core, so your local patch keeps working.
>
> PR: https://github.com/Vexa-ai/vexa/pull/1677. Reproduced your row shape against main before fixing — it failed exactly as you described.
>
> Would you validate it on your instance? A meeting with a transcript should read "recap ready"; a bot that never joined should still read "nothing captured".


DmitriyG228:
**`value-fsm` and `lite-smoke` are red on infrastructure, not on this diff.** MinIO's Docker Hub images stopped pulling anonymously today, so neither live leg can start a stack:

```
value-fsm  · minio-init Error pull access denied for minio/mc, repository does not exist …
lite-smoke · Unable to find image 'minio/minio:latest' locally
             pull access denied for minio/minio … make: *** [Makefile:114: up] Error 125
```

In `value-fsm` the scenario suite itself is intact — `10 passed, 5 skipped, 1 error`, and the one error is the `stack` fixture, not a scenario. The same break hits every PR that triggers the workflow: [#1676](https://github.com/Vexa-ai/vexa/actions/runs/35207554326) and [#1675](https://github.com/Vexa-ai/vexa/actions/runs/35207717877) fail identically on `minio/minio`. It needs a pinned or mirrored MinIO reference in `deploy/compose/docker-compose.yml` and `deploy/lite`, which is its own change.

Flagging it so the red is not read as this diff failing the live leg. The live leg is genuinely unproven here either way — stated in the PR description.

Everything that does exercise this diff is green: `gates`, `static`, `node`, `python`, `docs-current`, CodeQL, GitGuardian. The remaining reds (`DCO`, `contribution-rights`, `dco-no-override`, `merge-card`) are the human gates — one rights box ticked with a sign-off, and a non-author value sign-off.

Issue #1671, opened 2026-09-14
[DEPLOYMENT] Four blockers hit in sequence on a first `make lite` install (clean machine)


infocillasas:
Hi, and great thanks for this project !!

While evaluating Vexa for internal use (meeting bot + transcription through a
third-party API), I hit four separate problems on a first install, on a machine with
no prior Vexa state. They are independent, but I'm filing them together because you
meet them one after another on the same path. Happy to split into separate issues if
you prefer.

Once past all four, the install works and the API is a pleasure to use — the
per-user `PUT /user/transcription` contract is exactly what I needed.

**Environment**

| | |
|---|---|
| Version | commit `59e2c413a53479125b70b712ade12ab470d55512` (2026-09-07) |
| OS | Linux (Ubuntu 24.04 / Zorin) |
| Docker | 29.8.0 |
| Docker Compose | v5.5.1 |
| Command | `make lite` (= `make -C deploy/lite all`) |

---

## 1. `minio/minio` is gone from Docker Hub — the install fails

```
Starting MinIO (vexa-lite-minio)...
Unable to find image 'minio/minio:latest' locally
docker: Error response from daemon: pull access denied for minio/minio,
repository does not exist or may require 'docker login'
make: *** [Makefile:114: up] Error 125
```

MinIO no longer publishes to Docker Hub. Checked from a machine with no Docker Hub
authentication:

```
docker manifest inspect minio/minio:latest                       → unavailable
docker manifest inspect minio/minio:RELEASE.2024-09-13T20-26-02Z → unavailable
docker manifest inspect quay.io/minio/minio:latest               → OK
```

**References:** `deploy/lite/Makefile:145`, `deploy/compose/docker-compose.yml:53`

**Impact:** both documented install paths (`make lite` and `make all`) fail on a
clean machine. MinIO starts before the app container, so nothing of Vexa comes up.

**Workaround:** switch to `quay.io/minio/minio:latest`. You may also want to pin a
dated tag rather than `:latest`, so an upstream change can't break fresh installs
again.

---

## 2. `make lite` writes an `.env` with an empty `ADMIN_TOKEN` — the install is unusable

The `env` target seeds a minimal template when `deploy/compose/.env` is absent, which
is the case on a fresh clone:

```make
[...]

Issue #462, opened 2026-06-24
[Bug] Duplicate MS Teams chat messages are received


terrence-yu-eb:
## What happened?

When a user types a chat message into an MS Teams meeting chat Vexa will return duplicates of the messages. Sometimes more than 2 times. When new chat messages are sent the previous chat messages seem to duplicate more.

## What did you expect?

Vexa to only receive one chat message for every chat message sent in the chat.

## How to reproduce?

1. Add Vexa bot to an MS Teams meeting
2. Type something into the meeting chat.
3. Duplicates are 

## Logs / screenshots?

Here is a log of using the GET /chat endpoint when 2 chat messages have been sent into the chat window.

"chat": {
  "messages": [
    {
      "sender": "Unknown",
      "text": "this is a chat message 1",
      "timestamp": 1782326342080,
      "is_from_bot": false
    },
    {
      "sender": "Unknown",
      "text": "this is a chat message 1",
      "timestamp": 1782326342173,
      "is_from_bot": false
    },
    {
      "sender": "Unknown",
      "text": "this is a second chat message",
      "timestamp": 1782326374374,
      "is_from_bot": false
    },
    {
      "sender": "Unknown",
      "text": "this is a chat message 1",
      "timestamp": 1782326374432,
      "is_from_bot": false
    },
    {
      "sender": "Unknown",
      "text": "this is a second chat message",
      "timestamp": 1782326374432,
      "is_from_bot": false
    }
  ],
  "meeting_id": 13
},

## Version / env?

0.10.6.3.14

DmitriyG228:
Closing into #500 (**MS Teams meeting chat: read and write**) together with your #449 — one code change closes both, so they ride one issue. Your log excerpt is the discriminating evidence: duplicates that multiply as the chat grows is a re-scan re-emission with no per-message identity, and #500 attacks it at both ends — C2 gives each message a stable identity with a seen-set (the triple-scan test: same panel scanned three times, each message once), C3 makes the store upsert on that identity so the record can't duplicate downstream even if an upstream re-emit slips through.

[...]

PR #1658, opened 2026-09-07
gate:dataflow: gitignored build residue under core/<dom>/src is not an unregistered module — gates.mjs all runs twice in one worktree (#1657)

DmitriyG228:
**Delivers issue:** #1657

## Contribution rights

- [ ] **Independent:** I created this contribution, or otherwise have the right to submit it
  under Apache-2.0, and it is not owned or controlled by an employer, client, or other entity.
  
- [ ] **Employer/client authorization required:** an employer, client, or other entity owns or
  may control this contribution. I am requesting Vexa's private corporate-authorization process.
  
- [ ] **Unsure:** I need a private rights review before merge.
  

## Observation bundle

- **C1 · locate the walker —** ran: `grep -n "skippable\|lsdirs" scripts/gates.mjs` at base `c02e88985` · saw: `skippable` at `:23`, honoured by `walkDirs` (`:57`), `findFile` (`:75`) and `scanEnvReads` (`:1129`); `gateDataflow`'s `lsdirs` (`:842`) a raw `readdirSync` + `isDirectory` · concluded: one population, two definitions — the completeness walker is the only reader that never filters.
- **C1 · reproduce —** ran: `python3 -m compileall -q core/flows/src && node scripts/gates.mjs dataflow` at base · saw: `completeness: 'core/flows/src/__pycache__' exists on disk but is not registered in architecture.calm.json`; `mkdir core/flows/src/.venv core/flows/src/zz.egg-info` → three rows; `git status` empty throughout · concluded: gitignored residue is read as modules. Checked `gate:db-schema` (`python3 scripts/schema_digest.py`): `ast` parse, no imports, no bytecode — the residue comes from the pytest gates (`gate:python` and `gate:health` run `uv run pytest` in `core/flows`, whose `tests/conftest.py` puts `src` on `sys.path`).
- **C1 · fix —** `lsdirs` filters with `!skippable(n)` before the `isDirectory` test; `skippable` also recognises `*.egg-info` (`.gitignore:6`), so `gate:readme` and every other `walkDirs()` user close the same hole at once. Two lines of logic; the comments state the population, not the history.
[...]

PR #1678, opened 2026-09-17
lite: require gateway identity for agent-api; bind loopback

DmitriyG228:
## What this changes

Lite ran agent-api on `0.0.0.0:8100` and published it as a host port, with no gateway in front of it.

agent-api is the unit control plane — dispatch, chat, workspace attach/publish, routines. It derives the acting subject from the `X-User-Id` that only the **gateway** can resolve from a verified api-key, and falls back to `VEXA_AGENT_DEFAULT_SUBJECT` when no such header is present. Inside the container that fallback is the single-user convenience it was written to be: the only callers are the gateway, the terminal's server side, and the runtime's scheduler POSTing `/invocations`. On a published host port it is the whole control plane answering to whoever asks.

Three parts:

**1. Lite binds agent-api to loopback and requires gateway identity.**
- `deploy/lite/supervisord.conf` — `uvicorn --host %(ENV_VEXA_AGENT_API_BIND)s` and `VEXA_REQUIRE_GATEWAY_IDENTITY` in the program env.
- `deploy/lite/entrypoint.sh` — defaults `VEXA_AGENT_API_BIND=127.0.0.1` and `VEXA_REQUIRE_GATEWAY_IDENTITY=1`.
- `deploy/lite/Makefile` — `HOST_AGENT_PORT` is now empty by default, so no `-p …:8100`. The way in is the gateway's `/agent/*` prefix on `HOST_GATEWAY_PORT`. Setting `HOST_AGENT_PORT=<port>` republishes it for debugging on a trusted host and opens the bind with it (publishing a loopback would otherwise reach nothing); identity enforcement stays on, so a direct caller gets 401 instead of the single-user fallback.
- `make test` and the startup banner no longer point at `:8100`.

**2. The gateway now stamps the marker it was always supposed to stamp.**

`VEXA_REQUIRE_GATEWAY_IDENTITY` looks for `x-gateway-verified`. The gateway already **stripped** that header from inbound requests (it is in the F95 authority-header vocabulary) but never **wrote** it, so nothing in the system could set it — the flag existed but could not be switched on in any deployment without the front door 401ing itself. `_authorize` now writes it after the strip and past every authorization branch, so its presence downstream means "the gateway resolved these identity headers" and nothing else.

Note what makes this trustworthy: it carries no secret and is not meant to. It works because agent-api is reachable only through the gateway — which is what part 1 establishes for Lite, and what compose (`127.0.0.1:18100` on the host) and the chart (internal network) already did.

**3. `git checkout` ref validation in `workspace_attach`.**

`body.ref` reached `git checkout` unvalidated (`core/agent/control_plane/workspace_attach.py:172`). `git` reads a leading `-` as an option whatever parameter position the string arrived in, so the revision slot was an argument-injection surface next to a URL that had already been gated by `repo_ref` (#1539).

`repo_ref.assert_valid_ref` allow-lists the shape a branch, tag or commit id actually has — a deny-list of option spellings would have to anticipate git's whole option table. It runs at the route (400, not a 502) and again inside `_git_clone` before any subprocess exists, the same double-siting `assert_public_host` uses, so a caller arriving through the MCP or a future route cannot reach git either. The checkout also gets a trailing `--`, checkout's revision/pathspec disambiguator.

## Tests

- `core/agent/tests/test_repo_ref_host.py` — new REVISION section: option-shaped, whitespace-bearing, `..`-bearing, over-long and `@{…}` refs refused with `kind="ref"`; ordinary branches/tags/shas accepted; **no subprocess started** on refusal (the same load-bearing assertion the host half uses); and a real clone of a repo with a branch named like a path in the tree, proving the `--` disambiguation.
- `core/gateway/services/gateway/tests/test_proxy.py` — the marker is stamped on an authorized hop; an unauthorized request produces none (the downstream is never called); and the existing F95 strip test now sends a *distinctive* client value and asserts the gateway's own `1` is what arrives, which is strictly stronger than the previous "absent" assertion.

Run locally: `core/agent` 132 passed · `core/gateway/services/gateway` 353 passed, 1 xfailed (`test_edge_guard.py` skipped — needs the `fastapi-guard` dep, unrelated). `gate:config-contract`, `gate:lite-makefile`, `gate:dataflow`, `gate:contract-conformance`, `gate:docs-version` green.

`VEXA_AGENT_API_BIND` is read by the process supervisor and by no service, so it is declared in `CONFIG_LITE_UNADOPTED` naming the program it serves, per `gate:config-contract`'s own instruction.

## Overlap with #1671 — stated, not resolved

[#1671](https://github.com/Vexa-ai/vexa/issues/1671) reports four first-install blockers on a clean machine and another dispatch is fixing them. **It touches `deploy/lite/Makefile`, which this PR also touches** — specifically `deploy/lite/Makefile:145`'s `minio/minio:latest` pull, which no longer resolves on Docker Hub. This PR's Makefile edits are in a different region (the `HOST_*_PORT` block at the top, the `docker run` port flags, and the `test` target), so a textual conflict is unlikely but the two should land aware of each other. I have deliberately **not** touched the MinIO line or anything else in #1671's scope.

One consequence worth flagging for whoever merges second: on a machine with no cached `minio/minio` image, `make lite` cannot come up at all until #1671 lands, which is why the run-up verification below was done on a host that already has it.

## Undone / not verified

- `make lite` build-and-up verification on `bbb` — see the comment below for the result.
- Compose and the Helm chart are unchanged: compose already binds agent-api to `127.0.0.1:18100` on the host and the chart keeps it on an internal network with `agentApi.enabled: false` in prod, so Lite was the outlier. Turning the flag on for compose is a separate decision, because compose's terminal-to-agent-api path would want checking first.

Refs #462


DmitriyG228:
## Verified on `bbb` — built and run up

`make -C deploy/lite build` off this branch, then the app container run exactly as the new Makefile emits it (`-p 8456:8056 -p 3401:3001`, **no** `-p …:8100`).

One deviation, stated because it matters: `bbb` already had a live `vexa-lite` on the default container names, network and volumes, so I ran the `up` recipe's `docker run` by hand against isolated sidecars (`vexa-lite-462-*`, own network, own volumes) rather than letting `make up` reuse the default names. The app container's own arguments are the recipe's, unchanged. All thirteen supervised programs reached `RUNNING`.

| Check | Result |
|---|---|
| agent-api listen address inside the container | `tcp 127.0.0.1:8100 LISTEN` (was `0.0.0.0:8100`) |
| host port map | `3001→3401`, `8056→8456` — **8100 absent** |
| direct `GET :8100/api/models` from inside the container | **401** `gateway-signed identity required (VEXA_REQUIRE_GATEWAY_IDENTITY)` |
| same, with a forged `X-User-Id: attacker` | **401** — the single-user fallback is no longer reachable |
| `GET :8100/health` | 200 — liveness stays open, no identity needed |
| `GET :8456/agent/models` with a minted api-key | **200**, real body `{"chat_model":"default",…}` |
| `GET :8456/agent/models` with a forged `X-Gateway-Verified: 1` + `X-User-Id`, no key | **401** at the gateway |
| gateway `/health` · terminal `/` | 200 · 200 |

The last two rows are the pair that matters: the marker is worth nothing on its own — a client can spell it — and it is worth something because the gateway strips it inbound, stamps it only past authorization, and agent-api is reachable from nowhere else.

First run of this stack failed on something unrelated and worth recording: `admin-api` and `meeting-api` refuse to boot when `ADMIN_TOKEN`/`ADMIN_API_TOKEN` hold the documented placeholder (`changeme`), which is `config_preflight` doing its job. Re-run with generated tokens, all green.

Test containers, network and volumes torn down after the run.

## One thing I could not account for, flagged rather than buried

When I started, `bbb` had a live `vexa-lite` published on `0.0.0.0:8056`, `:3001` and **`:8100`** — the exposure this PR is about, running. By the time I tore my own stack down ~6 minutes later, that container, its `vexa-lite-postgres` and `vexa-lite-minio` sidecars, the `vexa-lite-net` network **and** the `vexa-lite-pgdata` / `vexa-lite-miniodata` volumes were all gone.

Every container and volume I created or removed was named `vexa-lite-462-*`, and `docker rm`/`docker volume rm` take exact names. Volume removal is also not part of `make down`, which keeps them. The daemon returns no event history for past windows, so I cannot show who did it. Most likely another session ran `make lite-down` plus the documented wipe. Recording it because a running Lite instance on that host is a live instance of what this PR fixes, and because I touched that host in the same window.

DmitriyG228:
## CI reconciliation — this branch adds no new red

| check | here | cause |
|---|---|---|
| `gates` · `python` · `node` · `static` · `docs-current` · CodeQL (×3) · GitGuardian | **pass** | — |
| `DCO` | **pass** after sign-off (amended, force-pushed with `--force-with-lease`) | mine, fixed |
| `lite-smoke` · `value-fsm` | fail | `pull access denied for minio/minio` and `minio/mc`. **This is [#1671](https://github.com/Vexa-ai/vexa/issues/1671) item 1** — MinIO no longer publishes to Docker Hub. Both checks passed on [#1653](https://github.com/Vexa-ai/vexa/pull/1653) before the upstream removal and now fail on every PR that reaches them. Nothing to do with this branch, and the reason I verified on a host that already had the image cached |
| `contribution-rights` · `evaluate` · `merge-card` | fail | fail identically on **every** open PR checked (#1662, #1658, #1653) — repo-wide, pre-existing |

So the one substantive suite that would exercise these changes end-to-end, `lite-smoke`, **cannot run until #1671 lands** — which is the sharpest form of the overlap noted in the description. The [manual run-up on `bbb`](https://github.com/Vexa-ai/vexa/pull/1678#issuecomment-5712519575) stands in for it in the meantime; re-run `lite-smoke` once #1671's MinIO fix is on `main` and this is rebased.

PR #1329, opened 2026-08-25
helm: support immutable image digests

lifeofgurpreet:
Delivers #1001

## Contribution rights

- [x] **Employer/client authorization required:** Biji-Biji Initiative controls this contribution; requesting Vexa private corporate authorization.

## Observation bundle

- **C1 — structured chart-owned images:** ran `make -C deploy/helm test` at `9772dabc7e7177a6956c59bf4b88b29e5af35a8f`; lint, existing template assertions, and digest regression tests passed.
- **C2 — immutable runtime inputs:** the new render fixture proves every chart-owned workload image and every runtime-spawned bot/agent reference resolves to `repository@sha256:<64 lowercase hex>`.
- **C3 — false controls:** the digest suite proves malformed, truncated, uppercase, non-sha256, tag-plus-digest, and incomplete repository/digest inputs fail before Kubernetes.

## Acceptance floor

| Row | Evidence |
|---|---|
| Exact digest contract | `make -C deploy/helm test` passes on `9772dabc7e7177a6956c59bf4b88b29e5af35a8f`; all chart-owned images and spawned references use the structured helper. |
| Legacy tags | `test_image_digests.sh` confirms the existing tag render before digest mode. |
| Fail closed | The same test exercises malformed and ambiguous selectors as negative controls. |
| Runtime image inventory | The fixture covers all chart-owned Pod/Job images plus the three runtime spawn references. |

No registry, cluster, database, provider, customer, or production action was performed.

## Docs diff

- Helm values reference and Kubernetes deployment guide document the immutable-reference contract and required running-image verification.
- The changelog fragment records the behavior without touching the hot changelog file.

## Security checks

- `ubs` over all 23 changed files: no critical or warning findings.
- `node scripts/gates.mjs image-licenses`: green.
- `node scripts/gates.test.mjs`: 19/19 pass.
- Focused Helm assertions: green.
- The full repository gate begins with existing schema/config-contract failures on the unchanged main base; this patch changes neither subsystem.

## Validation request

A non-author should run `make -C deploy/helm test` at the exact head, review the rendered digest and negative-control output, and attest the Helm/docs behavior. No live deployment is claimed; cluster imageID readback remains a later deployment gate.

## Authorship

Sole author: Kali.
Tooling disclosure: automated coding tools were used as instruments; no co-author trailers are included.

Issue #1001, opened 2026-08-05
[DEPLOYMENT] Helm components accept immutable repository@sha256 images


DmitriyG228:
## Value

A release operator can render and deploy the exact reviewed OSS image bytes, not a mutable tag, and prove the same digest from Helm values through the running Pod `imageID`.

## Observation

Platform billing witness issue `Vexa-ai/vexa-platform#209` requires one sealed OSS composition. The v0.12 Helm chart currently renders every long-running component as `repository:tag`:

```text
image: "{{ .Values.<component>.image.repository }}:{{ tag }}"
```

There is no per-component digest field or shared helper. A platform overlay therefore cannot render `repository@sha256:<64 hex>` without forking/repacking product chart templates.

## Expected / Actual / Verdict

**Expected:** each OSS component accepts either an immutable digest or an explicit tag; digest wins, malformed/combined identities fail rendering, and the existing tag path remains backward compatible.

**Actual:** the chart accepts tags only. Registry readback may later reveal bytes, but the desired bytes are not bound before deployment.

**Verdict:** exact-image billing staging remains RED at the OSS Helm image-reference contract. Fix this in OSS, then vendor the reviewed chart into `vexa-platform`; do not maintain a hosted-only template fork.

## Acceptance

- One helper renders exactly `repository@sha256:<64 lowercase hex>` when `image.digest` is set.
- Empty digest preserves the existing `repository:tag` behavior.
- Digest plus mutable/global tag ambiguity fails closed rather than silently ranking.
- Malformed, truncated, uppercase or non-`sha256` digest fails `helm template`.
- Gateway, admin-api, meeting-api, runtime, agent-api and terminal consume the helper; any other chart-owned runtime image is inventoried and dispositioned.
- Spawned bot/agent images retain their existing explicit full-reference contract and are covered separately.
- Render tests prove positive digest, legacy tag and all false controls.
- Documentation states registry digest readback and running Pod `imageID` equality are still required delivery gates.

## Custody

Point of introduction: `Vexa-ai/vexa` Helm chart. Platform #209 consumes the reviewed result only. No provider, stage, production, customer, database or image-registry mutation is authorized by this issue.

DmitriyG228:
[...]

PR #1679, opened 2026-09-17
fix(lite): the four blockers a first `make lite` hits on a clean machine (#1671)

DmitriyG228:
Refs #1671.

All four blockers reproduced on a clean Ubuntu host (no prior Vexa state), fixed, and re-run end to end from a fresh clone of `origin/main` + this patch. Reproduction and proof ran on a 128-core Linux box; every command below is from those logs.

---

### 1. `minio/minio` is gone from Docker Hub

**Fix** — point lite, compose and the helm chart at **quay.io**, MinIO's own registry, pinned to a dated release rather than `:latest`. `MINIO_IMAGE` / `MINIO_MC_IMAGE` are overridable (make variables in lite, `${VAR:-default}` in compose, `.env.example` entries, `minio.image` / `minio.mcImage` in helm values). `image-licenses.json` follows the rename.

The pin is `RELEASE.2025-04-22T22-12-26Z` deliberately: an upstream retag is what broke this in the first place, and MinIO releases after 2025-05 drop the web console that the `--console-address ":9001"` flag configures.

**Proof** — on the clean host, with no `quay.io/minio` layer cached:

```
Status: Downloaded newer image for quay.io/minio/minio:RELEASE.2025-04-22T22-12-26Z
  ✓ MinIO ready
```

and against Docker Hub from the same host, for the record:

```
minio/minio:latest                UNAVAILABLE
quay.io/minio/minio:latest        OK
```

---

### 2. `make lite` wrote an `.env` with an empty `ADMIN_TOKEN`

The mechanism is slightly worse than the report suggests. The empty value is not the whole story: `deploy/lite/entrypoint.sh` mints a random admin token *per boot* when none is supplied, and `docker exec … printenv` cannot see it (exec gets the image's ENV, not the entrypoint's exports). So the stack held a working admin token that **nothing outside the container could ever learn** — hence 403 on every plausible guess.

**Fix** — the `env` target mints the token host-side (`openssl rand -hex 24`, falling back to `python3` then `/dev/urandom`) and writes it into the repo-root `.env`; an existing `.env` whose `ADMIN_TOKEN` is empty gets one too, which is the shape a pre-fix install leaves behind. `make lite` then **prints it** on completion, together with the ready-to-use API key `provision-key.sh` already mints for the built-in `self-host@vexa.ai` user — which existed all along and was equally invisible.

**Proof** — the tail of a clean `make lite`:

```
  ADMIN_TOKEN (in /…/clean/.env) — mints users and API keys:
    e3b3467d…
  API key for the built-in self-host@vexa.ai user — use it right now:
    vxa_bot_1zbBu9…
    curl -s -H "X-API-Key: vxa_bot_1zbBu9…" http://192.168.1.4:8056/bots/status
```

and those two credentials actually working:

```
[2a] printed ADMIN_TOKEN → POST /admin/users        HTTP 201  {"id":2,"email":"blocker2-proof@example.com",…}
[2b] a wrong admin token → POST /admin/users        HTTP 403
[2c] printed API key → GET :8056/bots/status        HTTP 200
```

The `env` target is idempotent — a second run does not re-mint.

---

### 3. `up` printed `✓ Vexa Lite running` even when the container never started

**Fix** — the wait loop now records whether the gateway *answered*, bails out early if the app container is not running, and on failure exits non-zero after dumping the container's last 40 log lines. The diagnosis, not just the verdict.

**Proof** — the same broken stack, before and after. Baseline (`origin/main`), a boot where the gateway process died:

```
  Waiting for the gateway (up to 3 min)...
  ✓ Vexa Lite running          ← false
  ✓ Schema converges in-process on service startup
  ✗ gateway not responding (:8056)
```

With this patch, the identical boot:

```
  Waiting for the gateway (up to 3 min)...
  ✗ Vexa Lite did NOT come up: nothing answered http://localhost:8056/health.
    Last 40 lines of `docker logs vexa-lite`:
      …
      gateway.routes_manifest.ManifestError: /app/core/agent/routes.v1.json is missing — a deployed domain must declare its routes
      INFO gave up: gateway entered FATAL state, too many start retries too quickly
make: *** [Makefile:181: up] Error 1
```

A `docker run` that fails outright also exits non-zero (`Error 125`, verified with an unresolvable image tag).

---

### 4. `deploy/lite` mounted `~/.claude/.credentials.json` with no opt-in

**Fix** — lite now mounts **only** when `HOST_CLAUDE_CREDENTIALS` is explicitly set, as a make variable, an environment variable, or a repo-root `.env` key. Unset means nothing is mounted, whatever exists on the host. A set-but-missing path warns instead of silently doing nothing. Same contract compose has always had. `deploy/lite/README.md` documents the variable in the configuration table.

**Proof** — on the host where `~/.claude/.credentials.json` exists, a default `make lite`:

```
=== mounts on vexa-lite ===
(none)
```

and the same host with the variable set:

```
  Mounting Claude subscription credentials from /home/dima/.claude/.credentials.json (HOST_CLAUDE_CREDENTIALS)
…
/home/dima/.claude/.credentials.json -> /var/lib/vexa/host-claude-credentials (read-only)
```

---

### Minor note

`up` now depends on `env`, so `make -C deploy/lite up LOCAL_STT=1` provisions its own `.env` instead of assuming a prior full run. The README documents it as `make lite LOCAL_STT=1`, as suggested.

---

### End-to-end

Fresh clone of `origin/main` + this patch, published image (`vexaai/vexa-lite:v012`), empty volumes:

```
  ✓ Vexa Lite running
  ✓ Schema converges in-process on service startup
  ✓ gateway (:8056)
  ✓ agent-api (:8100)
  ✓ terminal (:3001)
…
EXIT=0
```

`node scripts/gates.mjs image-licenses` green; `node --test --test-name-pattern="image-licenses" scripts/gates.test.mjs` 7/7 pass.

### Overlap to resolve elsewhere, not here

A separate security change is in flight against `deploy/lite/supervisord.conf` and `deploy/lite/Makefile` — requiring gateway identity, and binding **agent-api to loopback**. This PR deliberately leaves both concerns alone: it does not touch the `-p` publish lines, the `test` target's agent-api probe, or `supervisord.conf` at all. Two lines here will need reconciling once that lands, both cosmetic: the `Agent API: http://$HOST:8100` line `all` prints, and the `agent-api (:8100)` front-door probe in `test`, neither of which will be reachable off-box after the bind changes.

### Not covered

Blocker 1's *failure* could not be reproduced live on the test host — it has `minio/minio:latest` in its local image cache, so the pull never happened. The failure is established by `docker manifest inspect` against Docker Hub from that host (UNAVAILABLE) and by the reporter's own log; the *fix* is proven by a real, uncached pull from quay.

## Contribution rights


DmitriyG228:
Collision check from a session dispatched to fix the same failure — no second PR opened, this branch already covers it.

**What this PR fixes repo-wide.** `pr-value` (value-fsm + lite-smoke) has failed on every PR since 2026-09-13 on `pull access denied for minio/minio` — Docker Hub returns `object not found` for both `minio/minio` and `minio/mc`, and MinIO's README now says the community edition is source-only. Run [35210065589](https://github.com/Vexa-ai/vexa/actions/runs/35210065589) on this branch is the only green `pr-value` since 2026-09-11. Every other open PR stays red until this merges.

**Version note, not a blocker.** Until 09-13 CI was pulling Hub `minio/minio:latest` = `RELEASE.2025-09-07T16-13-09Z` (`sha256:14cea493…`, also what `quay.io/minio/minio:latest` resolves to) and `minio/mc:latest` = `RELEASE.2025-08-13T08-35-41Z` (`sha256:a7fe349e…`). The pins here (`RELEASE.2025-04-22T22-12-26Z` / `RELEASE.2025-04-16T18-13-26Z`) step back about five months. Verified on bbb that the digest form of the previous images pulls and is byte-identical to what CI ran:

```
quay.io/minio/minio@sha256:14cea493d9a34af32f524e538b8346cf79f3321eff8e708c1e2960462bd8936e
quay.io/minio/mc@sha256:a7fe349ef4bd8521fb8497f55c6042871b2ae640607cf99d9bede5e9bdf11727
```

Those work as the `MINIO_IMAGE` / `MINIO_MC_IMAGE` defaults in compose and the lite Makefile (`gate:image-licenses` strips `@sha256:` before matching). The helm `repository:tag` template cannot carry a digest until #1001 / #1329.

**Two mentions left behind, cosmetic:** `deploy/README.md:16` (`minio` + `minio/mc` in the infra row) and the `repository: minio/minio` example comment at `scripts/gates.mjs:512`. No `docs/changelog.d/` fragment in the PR.

DmitriyG228:
Heads-up: the MinIO registry slice of this PR is carved out as [#1685](https://github.com/Vexa-ai/vexa/pull/1685), same hunks byte-for-byte on `docker-compose.yml`, `values.yaml`, `.env.example` and `image-licenses.json`, MinIO-only lines in the Lite Makefile and docs. Reason: `minio/*` being gone from Docker Hub fails `pr-value` on every open PR since 2026-09-17, and this PR's other three blockers deserve their own read. Either can land first; git merges identical changes cleanly. If this one is signed first, #1685 gets closed.

Issue #1667, opened 2026-09-12
Request: No way to send a provider-required HTTP header to a custom model endpoint


timjuedemann:
### The job you're trying to get done

Point Vexa's agent at a **custom model endpoint that requires a provider-specific request header** to
accept traffic. Concretely: the gateway I self-host fronts an upstream that rejects requests missing
its own session header — it answers `MissingSessionID` until the header is present. The same shape
recurs with any gateway that wants a routing header, an organisation/project header, an entitlement
header, or a client identifier.

Today `mode: custom` gives me a **`base_url` and an `api_key`** and nothing else
(`docs/docs/api/settings.mdx:99-115`). There is no field for a header. A gateway whose auth is "key +
one extra header" therefore cannot be configured at all, even though its endpoint is genuinely
OpenAI-compatible.

**Expected:** I can declare the extra header(s) alongside the endpoint and the credential, and agent
turns reach my gateway.
**Actual:** I run a small translation proxy on the host, point `base_url` at *that*, and inject the
header there.

### How you work around it today

A ~200-line Node proxy on the host (`http`, systemd, bound to the Docker bridge gateway) that:

1. injects the required header (plus a stable client identifier),
2. rewrites the auth header per dialect (see the sibling report — same proxy),
3. strips hop-by-hop headers that the upstream CDN rejects.

Every Vexa deployment that needs a custom header needs its own copy of this proxy. That is real
operational surface for a *configuration* gap: the proxy is now the critical path for agent chat, so
it has to be versioned, monitored and restarted like production code — for a missing header field.

**Suggested direction:** an optional header map on the account's model settings, e.g.

```json
{
  "mode": "custom",
  "base_url": "https://gw.example.com/v1",
  "api_key": "sk-…",
  "headers": { "x-provider-session": "…" }
}
```

applied to **both** call shapes (the harness and the completion adapters) from the one place where
`base_url` is already stamped into the dispatch environment
(`core/agent/control_plane/dispatch.py:185-190`). The `api_key` masking rule
(`settings.mdx:75-80`) would want the same treatment for any value marked secret. If a general header
map is judged too broad, the narrower version is fine too — the job is "a gateway that needs one more
header must be configurable without a proxy", not "arbitrary header injection".

seven7763:
Read the code to check whether there's any existing slot for this — there isn't, and two of the plumbing layers would silently swallow a `headers` field today, which is worth knowing before anyone wires the obvious thing.

[...]

Issue #1666, opened 2026-09-12
Bug: Custom gateways fail with "empty or malformed response (HTTP 200)" and 401 retry loops


timjuedemann:
### What happened

I self-host the compose stack and pointed **Settings → Models** at `mode: custom` with a gateway that
fronts an open-source model (my gateway exposes both dialects, but with different auth per dialect).
Agent chat turns ("Brief Chat") could not run at all. Three symptoms, one cause:

1. `API Error: API returned an empty or malformed response (HTTP 200) — check for a proxy or gateway
   intercepting the request`
2. `The provider returned an HTML error page instead of an API response … CDN or gateway (e.g.
   Cloudflare) blocked the request`
3. A 401 retry loop: the UI sat on "Working" for ~70 s, retrying every ~5 s.

None of these was the gateway misbehaving. The harness is the **Claude Code CLI**, which speaks the
**Anthropic Messages** dialect at `{ANTHROPIC_BASE_URL}/v1/messages` and authenticates with
**`x-api-key`** — but `mode: custom` sets `ANTHROPIC_BASE_URL` to the user's `base_url`, which the
docs describe as an **OpenAI-compatible endpoint** (`docs/docs/api/settings.mdx:102`). A gateway that
serves only `/v1/chat/completions`, or serves both dialects but wants `Authorization: Bearer` on one
and `x-api-key` on the other, cannot serve the harness.

**Expected:** setting a custom `base_url` runs agent turns against my gateway.
**Actual:** the harness errors before a turn completes. I only got it working by running a
dialect- and auth-translating proxy on the host and pointing `base_url` at that proxy.

### Deployment

Self-hosted compose (`deploy/compose`, project `vexa-v012`), image tag `v012`. Deployed tag:
**v0.12.27-5-g59e2c413** — symptoms observed there; the code below was re-verified against `main` at `59e2c41`
(2026-09-11), where the same code is present.

### Logs, screenshots, meeting shape

**Where we are** — code as it is on `main`:

- `core/agent/control_plane/dispatch.py:185-190` — `mode: custom` sets `ANTHROPIC_BASE_URL = base_url`
  (the harness) **and** `VEXA_LLM_PROVIDER=openai-compat` + `VEXA_LLM_BASE_URL = base_url` (the
  completion beats) from the **same** value. Its own comment: *"`mode: custom` points BOTH call
  shapes at the supplied gateway (an Anthropic-/OpenAI-compatible endpoint…)"*.
[...]

PR #1682, opened 2026-09-17
fix(agent): a custom gateway's two dialects are two settings, and a wrong answer says so (#1666, #1667)

DmitriyG228:
**Delivers issue:** #1666 (and #1667 — same code path)

## Contribution rights

> Left blank deliberately: this PR was prepared by an agent, and the template says an agent must
> not select a rights path. A human owner ticks exactly one. The commit carries the repository's
> standard `Signed-off-by` trailer, as every commit on `main` does.

---

`mode: custom` configures **two different wire contracts** from one `base_url` and one `api_key`:

| what runs | request | auth header | success body |
|---|---|---|---|
| agent chat / docs / routines (the `claude` CLI harness) | `POST {base}/v1/messages` | `x-api-key` | `content[]` blocks |
| meeting-time summaries (the completion adapters) | `POST {base}/chat/completions` | `Authorization: Bearer` | `choices[]` |

A gateway that serves these on different paths, or wants a different credential on each, could not
be configured at all — which is [#1666](https://github.com/Vexa-ai/vexa/issues/1666). The paths do
not even agree on where `/v1` belongs: the completion call appends `/chat/completions` to
`base_url` (so it usually ends in `/v1`), the harness appends `/v1/messages` (so it usually must
not). One value cannot satisfy both.

## Reproduced first

A stub gateway returning the shapes the reporter describes, run against the real adapters and the
real Settings → Models **Test** probe. Before this branch:

```
openai-compat, gateway answers the Anthropic shape at HTTP 200   -> CompletionResult(text='', model='m')
anthropic,     gateway answers the OpenAI shape at HTTP 200      -> CompletionResult(text='', model='m')

Test button — mixed-dialect gateway:    ok=True  "Live completion OK"
Test button — CDN HTML gateway:         ok=True  "Live completion OK"
Test button — per-dialect-auth gateway: ok=True  "Live completion OK"
```

Two findings beyond the report:

1. **The adapters returned an empty completion with no error.** A meeting-note beat against such a
   gateway wrote nothing and raised nothing.
2. **The Test button greened every broken configuration.** It sent `x-api-key` *and*
   `Authorization: Bearer` together — so a gateway with per-dialect auth can never fail it — and it
   graded the HTTP status, never the body. The operator's only self-service diagnostic said the
   gateway was fine while no turn could run.

After:

```
openai-compat  -> LLMError: DIALECT MISMATCH at …: POST /chat/completions answered HTTP 200 with an
                  Anthropic Messages body. … Point this call shape at an endpoint serving
                  /chat/completions (auth: Authorization: Bearer), or use the gateway's /v1/messages path.
CDN HTML 200   -> LLMError: … answered POST /chat/completions with an HTML page, not an API response
                  (content-type text/html) — a CDN or gateway (e.g. Cloudflare) … intercepted the request.
401            -> LLMAuthError: … rejected the credential sent as 'x-api-key' (anthropic dialect).
                  Not retried: reconcile the key with the endpoint …

Test button — mixed-dialect gateway: ok=False  agent chat (Anthropic Messages at …/v1/messages):
  HTTP 200 answered an OpenAI chat.completion body … the claude-code harness rejects that as
  'empty or malformed'. · meeting beats (OpenAI chat-completions at …/chat/completions): …
```

## What changed

- **`harness_base_url` / `harness_api_key`** on the model config address the Messages side
  independently. Each falls back to its `base_url` / `api_key` twin, so a gateway serving both
  dialects on one origin with one credential is configured exactly as before — no migration.
- **`headers`** (`Name: Value` lines, or a JSON object) rides **both** call shapes:
  `VEXA_LLM_EXTRA_HEADERS` for the adapters, `ANTHROPIC_CUSTOM_HEADERS` for the CLI (verified
  present in the shipped CLI binary, newline-split). That closes
  [#1667](https://github.com/Vexa-ai/vexa/issues/1667) — same code path, same dispatch overlay —
  and removes the need for the ~200-line proxy described there. Values mask on read-back like
  `api_key`, names stay in the clear, and extras never override the auth header.
- **`core/agent/llm/dialects.py`** judges a 2xx body by shape rather than status: the other
  dialect's body, an HTML page, a non-JSON body, a refusal and an empty completion each fail by
  name, naming which dialect answered and which setting moves.
- **No retry on 401/403** in either adapter and in the probe — terminal, named, one request. The
  message says what header the key was sent as.
- **The Test probe** asks each call shape its own question at its own path with only its own auth
  header, and grades the body. It reports the harness shape first: a deployment whose Messages
  side fails has no working chat whatever the completion side says.
- Admin-api normalizes and validates `headers` at the edge (422 rather than storing a header that
  silently never arrives); agent-api passes the stored string through verbatim, so no consumer
  re-parses what the user typed.
- Docs (`/api/settings`) now state the two-endpoint contract, the `/v1` path asymmetry, and that an
  OpenAI-only gateway cannot run agent turns today.

## Tests

`core/agent`: 589 → 627 passing (38 new cases). New `tests/test_llm_dialects.py` (18) plus dialect-mismatch,
HTML-interception, empty-completion, no-retry-on-401 and extra-header cases on both adapters;
`tests/test_config_test.py` rewritten around the corrected probe (including
`test_a_401_is_terminal_for_that_shape_not_retried`, which asserts exactly one request per shape);
`tests/test_unit_foundation.py` covers the dispatch overlay's per-dialect split and header
passthrough. `core/identity/services/admin-api`: new `tests/test_model_headers.py` (13, no docker)
plus the DB-backed round trip in `test_model_settings.py`.

Gates run locally, all green: `python` (14 packages), `config-contract`, `isolation-py`,
`graph-py`, `test-isolation`, `exports`, `schema`, `contract-version`, `dataflow`, `readme`,
`docs-version`.

## Not done

- **No OpenAI-dialect harness** (option 3 in #1666). `HARNESS_RUNNERS` still holds only
  `claude-code`, so a gateway serving *only* `/chat/completions` still cannot run agent turns —
  it now fails loudly and says why, at the Test button, instead of hanging for ~70s. Making the
  docs' original promise literally true needs a harness that speaks Chat Completions with a tool
  loop; that is its own issue, not this one.
- **The CLI's own retry loop is not ours to stop.** The ~5s retries live inside the bundled
  `claude` binary. What this branch removes is the misconfiguration that fed it, and the silence
  around it. (`CLAUDE_CODE_MAX_RETRIES` exists in the binary but is not stamped here — capping it
  globally would also cap legitimate 429/529 backoff.)
- **Not verified on a real deployment.** Everything above was proven against a local stub gateway
  and unit tests; no compose stack was brought up.
- `clients/terminal` could not be typechecked (its `node_modules` is not installed in this
  worktree); the change there is three added form fields and a type widening.

- **C1 — the stub gateway.** Ran: a local HTTP stub answering the four shapes the reporter
  describes (OpenAI body at `/v1/messages`, Anthropic body at `/chat/completions`, a CDN HTML 200,
  a 401 unless the dialect's own auth header is present, a 400 `MissingSessionID` unless
  `x-provider-session` is present), driven through the **real** `OpenAICompatCompletion`,
  `AnthropicCompletion` and `config_test.run_models_test`. Saw: `CompletionResult(text='')` twice
  with no exception, and `ok=True "Live completion OK"` from the Test button on three
  configurations that cannot run a turn. Concluded: the report's cause is right, and the
  diagnostic surface is part of the bug — fix the parsing, the probe and the config together.
- **C2 — the CLI's header knob.** Ran: `strings` over the shipped `claude` binary. Saw:
  `ANTHROPIC_CUSTOM_HEADERS`, `Invalid ANTHROPIC_CUSTOM_HEADERS`, and a `\n|\r\n` split adjacent
  to it. Concluded: the harness half of #1667 needs no proxy — stamp the same `Name: Value` string
  the adapters read, and make that the stored form so nothing re-parses it.
- **C3 — the same stub after the change.** Ran: the identical script. Saw: every shape failing by
  name (quoted above), and — with `harness_base_url` / `harness_api_key` / `headers` configured
  through `overlay_model_config` — both call shapes reaching the gateway and the Test button green,
  naming `x-provider-session` without its value. Concluded: both reports are closed on the
  configuration axis; the OpenAI-harness axis is not (see **Not done**).

## Acceptance floor

| Row | Evidence |
|---|---|
| A gateway with per-dialect auth is configurable | `test_per_dialect_auth_and_endpoints_1666`, `test_dispatcher_custom_mode_splits_the_two_dialect_contracts_1666`, C3 |
| A 200 carrying the other dialect fails by name | `test_llm_dialects.py` (both directions), `test_openai_body_at_200_fails_by_name_not_empty_text` on both adapters |
| A CDN HTML 200 names the interception | `test_html_from_a_cdn_names_the_interception`, `test_html_error_page_at_200_names_the_gateway` |
| 401 is terminal, not retried | `test_401_is_not_retried` (both adapters, asserts one request), `test_a_401_is_terminal_for_that_shape_not_retried` (probe) |
| The Test button cannot green a config that cannot run | `test_openai_only_gateway_is_not_green_1666`, `test_openai_body_at_the_messages_path_is_not_green_1666`, `test_html_from_a_cdn_is_not_green_1666` |
| An extra header reaches both call shapes (#1667) | `test_extra_headers_ride_both_call_shapes_1667`, `test_dispatcher_custom_mode_extra_headers_ride_both_call_shapes_1667`, C3 |
| Header values never echo | `test_masking_keeps_names_and_hides_values`, `test_per_dialect_fields_round_trip_and_mask`, `test_run_models_test_routes_custom_vs_subscription` |

## Docs diff (D6c)

`docs/docs/api/settings.mdx` — the reference page the reporter cited (`settings.mdx:102`, "an
OpenAI-compatible endpoint") is the sentence that was wrong. It now carries the two-contract table,
the `/v1` path asymmetry, the new fields, and the statement that an OpenAI-only gateway cannot run
agent turns today. `docs/changelog.d/1666-custom-gateway-dialects.md` carries both entries.
`core/agent/llm/README.md` documents `dialects.py` and `VEXA_LLM_EXTRA_HEADERS`.

## Security checks (on the diff)

No new dependencies. The new surface is a config field carried to an HTTP header — masked on
read-back (names clear, values hidden), validated at the admin-api edge, ASCII-only header names
and values, and the extras are applied **before** the auth header so a configured header can never
replace a credential (`test_extra_headers_are_sent_and_never_override_auth`, both adapters).
Secrets scan: the credential-shaped strings in the new tests are `fake-…` fixtures.

## Validation request

Best witness is the reporter (@timjuedemann), who has the gateway this is about: point
Settings → Models at it **without** the translation proxy — `base_url` + `api_key` for the
OpenAI side, `harness_base_url` + `harness_api_key` for the Messages side, the required header in
`headers` — press **Test**, then run a Brief Chat turn. What to watch: the Test verdict should name
each call shape separately, and a wrong field should produce a message that says which dialect
answered rather than "empty or malformed". Deployment: self-hosted compose (`deploy/compose`).

Refs #1666
Refs #1667


DmitriyG228:
CI state, so nobody reads the reds as this diff:

- `gates` **pass** — python, node, static. CodeQL, GitGuardian, DCO, docs-current pass.
- `lite-smoke` and `value-fsm` **fail on `docker pull minio/minio`** — *"pull access denied … repository does not exist"*. MinIO withdrew its Docker Hub repository; it breaks every PR and `main`, and it is already fixed by [#1679](https://github.com/Vexa-ai/vexa/pull/1679) (moves to `quay.io/minio/*`). Nothing to do here — this branch will go green once that lands or this rebases onto it.
- `contribution-rights` / `merge-card` **fail by design**: the rights path and `state: value-signed` are human acts. An agent prepared this PR and the template forbids it selecting a rights path.

